PR #95, opened 2024-03-19
Fix CustomerInfo parsing and add support to non subscription parsing

tonidero:
## Motivation / Description
While I was testing the SDK I noticed we were not correctly creating the CustomerInfo object. There was a problem mapping from the subscriptions in the response to the final CustomerInfo. This fixes that and adds support for parsing non subscriptions. We will not be supporting buying them initially but I thought it would be good to still be able to access the relevant data in the CustomerInfo object


api-report/purchases-js.api.md
```diff
@@ -11,7 +11,7 @@ export interface CustomerInfo {
         [productIdentifier: string]: Date | null;
     };
     readonly allPurchaseDatesByProduct: {
-        [productIdentifier: string]: Date;
+        [productIdentifier: string]: Date | null;
```
tonidero, line 14: This is actually nullable in the mobile APIs

api-report/purchases-js.api.md
```diff
@@ -28,6 +28,7 @@ export interface EntitlementInfo {
     readonly identifier: string;
     readonly isActive: boolean;
     readonly isSandbox: boolean;
+    readonly latestPurchaseDate: Date;
```
tonidero, line 31: Added this field to follow the mobile SDKs API

src/entities/customer-info.ts
```diff
@@ -168,13 +174,42 @@ function isActive(
...
+      subscriptions[productIdentifier],
+    );
+  } else if (productIdentifier in latestNonSubscriptionPurchaseByProduct) {
+    return toNonSubscriptionEntitlementInfo(
+      entitlementIdentifier,
+      entitlementInfoResponse,
+      latestNonSubscriptionPurchaseByProduct[productIdentifier],
+    );
+  }
+  throw new PurchasesError(
```
tonidero, line 196: This should normally not happen on backend responses.

antoniobg (approved):

PR #98, opened 2024-03-20
Handle network errors

tonidero:
## Motivation / Description
To make a more flexible API, we shouldn't be returning the network error directly. Instead, this will wrap the error into our own `PurchasesError` type, with a `NetworkError` code.

### DIFF
--- a/src/networking/http-client.ts
+++ b/src/networking/http-client.ts
@@ -15,15 +15,27 @@ export async function performRequest<RequestBody, ResponseType>(
   body?: RequestBody,
   headers?: { [key: string]: string },
 ): Promise<ResponseType> {
-  const response = await fetch(endpoint.url(), {
-    method: endpoint.method,
-    headers: getHeaders(apiKey, headers),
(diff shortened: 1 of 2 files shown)


src/networking/http-client.ts
```diff
@@ -15,15 +15,27 @@ export async function performRequest<RequestBody, ResponseType>(
...
+      body: getBody(body),
+    });
 
-  await handleErrors(response, endpoint);
+    await handleErrors(response, endpoint);
 
-  return (await response.json()) as ResponseType; // TODO: Validate response is correct.
+    return (await response.json()) as ResponseType; // TODO: Validate response is correct.
+  } catch (error) {
+    if (error instanceof TypeError) {
```
tonidero, line 29: According to docs: https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#checking_that_the_fetch_was_successful. TypeError will be a network error.

antoniobg (approved):

PR #102, opened 2024-04-09
[BIL-292] Add `formattedPrice` to Price

tonidero:
## Motivation / Description
We want to have a precomputed label that developers can use to display the price. This will use the same formatter we use in our purchase form. This is currently using the hardcoded locale `en-US`, but we can change that once we support other locales.

## Additional comments

image

### DIFF
--- a/api-report/purchases-js.api.md
+++ b/api-report/purchases-js.api.md
@@ -170,6 +170,7 @@ export type PeriodType = "normal" | "intro" | "trial";
 export interface Price {
     readonly amount: number;
     readonly currency: string;
+    readonly formattedPrice: string;
 }
 
 // @public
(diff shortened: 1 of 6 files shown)


api-report/purchases-js.api.md
```diff
@@ -170,6 +170,7 @@ export type PeriodType = "normal" | "intro" | "trial";
 export interface Price {
     readonly amount: number;
     readonly currency: string;
+    readonly formattedPrice: string;
```
tonidero, line 173: Also, note that since this will be a precomputed-property, it won't respond to locale changes... But I think it's ok. Lmk if you think otherwise.

nicfix (approved):

PR #104, opened 2024-04-15
Improve API key validation

tonidero:
## Motivation / Description
Makes sure API keys only have the characters we allow.

### DIFF
--- a/src/helpers/configuration-validators.ts
+++ b/src/helpers/configuration-validators.ts
@@ -1,7 +1,8 @@
 import { ErrorCode, PurchasesError } from "../entities/errors";
 
 export function validateApiKey(apiKey: string) {
-  if (!apiKey.startsWith("rcb_")) {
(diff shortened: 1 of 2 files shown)


src/helpers/configuration-validators.ts
```diff
@@ -1,7 +1,8 @@
 import { ErrorCode, PurchasesError } from "../entities/errors";
 
 export function validateApiKey(apiKey: string) {
-  if (!apiKey.startsWith("rcb_")) {
+  const api_key_regex = /^rcb_[a-zA-Z0-9_.-]+$/;
```
tonidero, line 4: Right now we don't allow `_`, `.` or `-`. But I thought to support it just in case... Happy to change that back though.

antoniobg (approved):

src/helpers/configuration-validators.ts
antoniobg: We do support `_` for sandbox keys, which are `rcb_sb_...`

PR #111, opened 2024-05-13
BIL-551-redesign-demo-app

jefago:
## Motivation / Description

Redesigned demo app for customer demos


examples/rcbilling-demo/src/pages/paywall/index.tsx
```diff
@@ -1,71 +1,66 @@
...
-  P1M: "monthly",
-  P1Y: "yearly",
-  P2W: "2 months",
-  P1D: "daily",
-  PT1H: "hourly",
-  P1W: "weekly",
+  P3M: "quarter",
+  P1M: "mo",
+  P1Y: "yr",
+  P2M: "2mo",
```
tonidero, line 18: 👍

examples/rcbilling-demo/src/pages/paywall/index.tsx
```diff
@@ -88,12 +83,10 @@ const PaywallPage: React.FC<IPaywallPageProps> = ({ purchases }) => {
 
       console.log(`CustomerInfo after purchase: ${customerInfo}`);
 
-      window.location.href = "/success";
+      navigate(`/success/${purchases.getAppUserId()}`);
     } catch (e) {
       if (e instanceof PurchasesError) {
         console.log(`Error performing purchase: ${e}`);
-        // TODO: We should display an error here.
-        window.location.href = "/";
```
tonidero, line 96: Wonder if we should show an alert here... But that can come in a different PR

tonidero (approved):
Love the new App!

I only have a small concern. In my experience it's been useful to have 2 different test apps. One dev facing, which should look like a common app and display how to integrate our SDK (similar to MagicWeather in the mobile SDKs), and another one more internal, intended for our use only, where we can have more debugging tools for our SDK and test faster. The downside of this is that we usually test on the internal app and the dev-facing app usually is less commonly used.

I was wondering whether it might be worth to keep the existing app as the "internal RC app" and make the new app the "external dev-facing" app...

Having said that, I think this should be ok right now, so I'm ok keeping this as our only sample app for now

tonidero:
Additionally, I would rename the PR title to something user facing since the release automations use the PR title to include the change in the changelog. Also, since this doesn't affect the SDK itself, I would tag it with the `test` label instead of `feat`, so it's properly tagged in the changelog.

examples/rcbilling-demo/src/pages/paywall/index.tsx
jefago: yeah – the main reason I removed this for now is since this also fires when the purchase is cancelled, and then redirecting back to the home page feels very jarring. so we'd probably need better error handling overall.

jefago:
> Additionally, I would rename the PR title to something user facing since the release automations use the PR title to include the change in the changelog. Also, since this doesn't affect the SDK itself, I would tag it with the `test` label instead of `feat`, so it's properly tagged in the changelog.

Thanks, I did that. I wasn't sure which label to pick since they all don't quite fit, but makes sense if that's the way they are being used in the release automation.

PR #112, opened 2024-05-14
Allow multiple configure calls to create new purchases

tonidero:
## Motivation / Description
Currently if the developer calls `configure` multiple times, we return the same instance of `Purchases` and ignore the call. This might be confusing if the developer changes the user id in the other call. 

Normally, the developer should only call configure once, and we indicate that with a log, but in case they do, a more sensible option is to recreate the purchases instance with the given parameters (api key and user id) if they have changed.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,6 +1,6 @@
 GIT
   remote: https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal
-  revision: 1e624200ff2ddf0d398ce072171db35d017c408b
+  revision: 8ec007234d81f7a17a09a7e9243983485fac84b2
   branch: main
   specs:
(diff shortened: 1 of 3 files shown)


Gemfile.lock
```diff
@@ -1,6 +1,6 @@
 GIT
   remote: https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal
-  revision: 1e624200ff2ddf0d398ce072171db35d017c408b
+  revision: 8ec007234d81f7a17a09a7e9243983485fac84b2
```
tonidero, line 3: I got this in, just updating the fastlane plugin to the latest. Shouldn't affect much here.

jefago (approved):

PR #113, opened 2024-05-16
Added support for SubscriptionPurchaseOptions in ProductResponses

nicfix:
## Motivation / Description
We updated the /products endpoint to return purchase options that include, as example, trial offers.
This PR adds the necessary interfaces to support the data returned by Khepri as of now.

## Changes introduced
* Added the new interfaces in product-responses.ts
* Updated the test response from the endpoint with a trial offer and a base option

## Linear ticket
BIL-502

### DIFF
--- a/src/networking/responses/products-response.ts
+++ b/src/networking/responses/products-response.ts
@@ -19,10 +19,7 @@ export interface SubscriptionPurchaseOption extends PurchaseOption {
 }
 
 export interface ProductResponse {
-  current_price: {
-    amount: number;
-    currency: string;
-  };
+  current_price: Price;
   identifier: string;
   normal_period_duration: string;
   product_type: string;
--- a/src/networking/responses/products-response.ts
+++ b/src/networking/responses/products-response.ts
@@ -1,3 +1,23 @@
+export interface Price {
+  amount: number;
+  currency: string;
+}
+
+export interface PurchaseOptionPrice {
+  period_duration: string | null;
+  price: Price | null;
+  cycle_count: number;
+}
+
+export interface PurchaseOption {
+  id: string;
+}
+
+export interface SubscriptionPurchaseOption extends PurchaseOption {
+  base_price: PurchaseOptionPrice;
+  trial: PurchaseOptionPrice | null;
+}
+
 export interface ProductResponse {
   current_price: {
     amount: number;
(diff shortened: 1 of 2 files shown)


src/networking/responses/products-response.ts
```diff
@@ -1,3 +1,23 @@
...
+  id: string;
+}
+
+export interface SubscriptionPurchaseOption extends PurchaseOption {
+  base_price: PurchaseOptionPrice;
+  trial: PurchaseOptionPrice | null;
+}
+
 export interface ProductResponse {
   current_price: {
```
tonidero, line 22: You could potentially substitute the type here to `Price` I think? Though we will be mostly ignoring `current_price` in future PRs I imagine and we could even remove it then.

src/networking/responses/products-response.ts
```diff
@@ -1,3 +1,23 @@
+export interface Price {
+  amount: number;
+  currency: string;
+}
+
+export interface PurchaseOptionPrice {
+  period_duration: string | null;
+  price: Price | null;
```
tonidero, line 8: Hmm in the end, I think the `price` should never be null? And I'm not sure about the `period_duration`....

tonidero (approved):
Just a couple of questions but looking good!

src/networking/responses/products-response.ts
nicfix: Absolutely, changing it!

tonidero:
I would suggest labeling this PR with `refactor` so it works with the automations, and considering this PR itself doesn't provide any new features yet.

src/networking/responses/products-response.ts
nicfix: The price is null for trials

tonidero:
Also, might be good to change the PR title to something more user-facing since it's used for the CHANGELOG auto generation.

src/networking/responses/products-response.ts
tonidero: Ahh right makes sense. Same question about the duration though

nicfix: In case of the period_duration that was supposed to be reused for one-offs.

tonidero: Makes sense, thanks!

PR #117, opened 2024-05-23
Trial support

nicfix:
## Motivation / Description
We just added the SubscriptionOptions to support trials in khepri, now it's time to bring these changes to the SDK.

## Changes introduced
* Added support for  deserialising the `SubscriptionOptions`
* Added a new `.purchase` method 
* Marked the `.purchasePackage` method as deprecated in favour of `.purchase`
* Updated the demo app to support the purchaseOptions

## Linear ticket (if any)
BIL-500

### DIFF
--- a/src/ui/rcb-ui.svelte
+++ b/src/ui/rcb-ui.svelte
@@ -70,8 +70,6 @@
         productDetails = rcPackage.rcBillingProduct;
         brandingInfo = await backend.getBrandingInfo();
 
-        console.log("PRODUCT: ", productDetails);
-
         if (state === "present-offer") {
             if (customerEmail) {
                 handleSubscribe();


src/entities/offerings.ts
```diff
@@ -96,6 +115,12 @@ export interface Product {
    * The offering ID used to obtain this product.
    */
   readonly presentedOfferingIdentifier: string;
+
```
nicfix, line 118: Add docstrings here

src/entities/offerings.ts
```diff
@@ -96,6 +115,12 @@ export interface Product {
...
+  readonly defaultSubscriptionPurchaseOptionId: string | null;
+
+  readonly subscriptionPurchaseOptions: {
```
nicfix, line 121: Add docstrings here

src/entities/offerings.ts
```diff
@@ -71,6 +75,21 @@ export interface Price {
   readonly formattedPrice: string;
 }
 
+export interface PurchaseOptionPrice {
```
nicfix, line 78: Add docstrings to these new classes

src/main.ts
```diff
@@ -229,11 +229,13 @@ export class Purchases {
    * @param rcPackage - The package you want to purchase. Obtained from {@link Purchases.getOfferings}.
    * @param customerEmail - The email of the user. If null, RevenueCat will ask the customer for their email.
    * @param htmlTarget - The HTML element where the billing view should be added. If null, a new div will be created at the root of the page and appended to the body.
+   @param purchaseOptionId - The ID for a specific purchase option. Used to purchase trials and other offers. Obtained from {@link Purchases.getOfferings}.
    * @returns The customer info after the purchase is completed successfuly.
    * @throws {@link PurchasesError} if there is an error while performing the purchase. If the {@link PurchasesError.errorCode} is {@link ErrorCode.UserCancelledError}, the user cancelled the purchase.
    */
   public purchasePackage(
     rcPackage: Package,
+    purchaseOptionId: string, // WARNING: This is a breaking change
```
nicfix, line 238: Can I find a way to make this breaking change less impactful?

Or, if I have to make a breaking change, can I make it so that it reduces chances of a breaking change in the future? (Maybe using an object for the parameters with optional fields)

src/main.ts
```diff
@@ -229,11 +230,13 @@ export class Purchases {
    * @param rcPackage - The package you want to purchase. Obtained from {@link Purchases.getOfferings}.
    * @param customerEmail - The email of the user. If null, RevenueCat will ask the customer for their email.
    * @param htmlTarget - The HTML element where the billing view should be added. If null, a new div will be created at the root of the page and appended to the body.
+   * @param purchaseOptionId - The ID for a specific purchase option. Used to purchase trials and other offers. Obtained from {@link Purchases.getOfferings}.
    * @returns The customer info after the purchase is completed successfuly.
    * @throws {@link PurchasesError} if there is an error while performing the purchase. If the {@link PurchasesError.errorCode} is {@link ErrorCode.UserCancelledError}, the user cancelled the purchase.
    */
   public purchasePackage(
     rcPackage: Package,
+    purchaseOptionId: string = "base_option",
```
nicfix, line 239: This is a breaking change.

@tonidero I think we might want to handle these parameters with an object to avoid future changes to cause a breaking change.

Something like:
```
public purchasePackage(
    rcPackage: Package,
    options: {
        purchaseOptionId?: string | undefined,
        customerEmail?: string,
        htmlTarget?: HTMLElement        
    }
```

So that clients will be able to use it the fields as named parameters instead of ordered ones

```
 purchases.purchasePackage(package, { purchaseOptionId: 'some_id', htmlTarget: element}
```

tonidero: Yes! I agree with that approach. One thing I would do for now though is to keep the old one and deprecate it and add this one as a separate method. Also, we can give a name to the options like `PurchaseOptions`  or `PurchaseParams`

nicfix: What about `purchaseSubscriptionOption` for the name of the method?

Regarding the type for the params: yes I'll create an interface for it and call it `PurchaseOptions`.

I'll fix another couple of bugs and add some tests and then I'll change this interface, thanks for answering to the comment!

tonidero: > What about purchaseSubscriptionOption for the name of the method?

Hmm I think I would keep it as `purchasePackage` if possible? Just with different parameters. That way we try to keep everything into a single purchase method. 

In the android SDK we had a similar situation and had a ton of purchase methods. In the end, we switched to the builder pattern and now have a single purchase method that receives an object constructed from `PurchaseParams.Builder`. I think that exact approach might not be as common in TS, but having an object for all optional parameters makes the most sense to me.

nicfix: Since in typescript overloads allow only one implementation:
https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads

I would prefer to define the `purchase` method and add only the `package` parameter as positional (and mandatory).

For all the rest we will use the params object.

Wdyt?

nicfix: I updated the code with the `purchase` method here: https://github.com/RevenueCat/purchases-js/pull/117/files#diff-4fab5baaca5c14d2de62d8d2fceef376ddddcc8e9509d86cfa5643f51b89ce3dR260

src/networking/backend.ts
```diff
@@ -63,17 +64,24 @@ export class Backend {
...
     };
 
+    const requestBody: SubscribeRequestBody = {
+      app_user_id: appUserId,
+      product_id: productId,
+      email: email,
+      presented_offering_identifier: offeringIdentifier,
+    };
+
+    if (purchaseOptionId && purchaseOptionId !== "base_option") {
```
nicfix, line 77: Maybe we should move this logic to the backend.

The 'base_option' is a fake one that we return whenever we do not have an offer.

The backend endpoint expects an existing `offer_id` if specified, otherwise it shouldn't be passed.

We can definitely improve this as a second step (since it's not publicly exposed) and make sure that the backend accepts 'base_option' as if the `option_id` was not set.

src/main.ts
```diff
@@ -226,17 +229,40 @@ export class Purchases {
    * package from {@link Purchases.getOfferings}. This method will present the purchase
    * form on your site, using the given HTML element as the mount point, if
    * provided, or as a modal if not.
+   * @deprecated - please use .purchase
```
nicfix, line 232: @tonidero is this enough to mark the endpoint as deprecated?

Should we update the documentation somewhere else?

src/main.ts
tonidero: Hmm ok, the idea of naming it `purchasePackage` originally is that this will purchase a package, and we might want to allow someone not to use offerings in the future and purchase a product directly.

WDYT about adding the package to the PurchaseOptions type as mandatory instead? That way in the future we can try to make that object accept either a package or a product as mandatory and process it internally.

src/main.ts
tonidero: I believe it is yes: https://tsdoc.org/pages/tags/deprecated/. You can also generate the docs locally and see how they look, to make sure it's marked as deprecated.

We should update the docs in docusaurus with the new way to purchase once this is merged

src/main.ts
nicfix: I am fine with that, I can migrate all the parameters to the object and that's it.

src/main.ts
nicfix: Creating a ticket

nicfix: Done: https://linear.app/revenuecat/issue/BIL-590/[trials]-update-docusaurus-docs-with-the-new-purchase-method

src/main.ts
nicfix: Done!

examples/rcbilling-demo/src/pages/paywall/index.tsx
```diff
@@ -77,9 +113,17 @@ const PaywallPage: React.FC = () => {
...
+
     // How do we complete the purchase?
     try {
-      const { customerInfo } = await purchases.purchasePackage(pkg);
+      const { customerInfo } = offerId
+        ? await purchases.purchase({
+            rcPackage: pkg,
+            subscriptionPurchaseOptionId: offerId,
+          })
+        : await purchases.purchasePackage(pkg); // Testing backwards compatibility.
```
tonidero, line 126: Hmm I would probably avoid calling deprecated call here, since some people might be looking at this as sample code. I think the `purchasePackage` method should just call the `purchase` method directly, and we can have some unit tests to make sure it's doing the correct calls so I think it's ok not to test it here.

examples/rcbilling-demo/src/tests/main.test.ts
```diff
@@ -47,6 +47,7 @@ async function setupTest(
   });
   const page = await browser.newPage();
   await navigateToUrl(page, userId);
+  await page.setViewport({ width: 1366, height: 768 });
```
tonidero, line 50: Oh curious, was this needed?

src/entities/offerings.ts
```diff
@@ -71,6 +75,57 @@ export interface Price {
...
 }
 
+/**
+ * Represents the price and duration information for a phase of the purchase option.
+ * @public
+ */
+export interface PurchaseOptionPrice {
+  /**
+   * The duration of the purchase option price.
+   * For applicable options (trials, initial/promotional prices), otherwise null
```
tonidero, line 85: We might want to mention the format `ISO 8601` in the docs here.

src/entities/offerings.ts
```diff
@@ -71,6 +75,57 @@ export interface Price {
   readonly formattedPrice: string;
 }
 
+/**
+ * Represents the price and duration information for a phase of the purchase option.
+ * @public
+ */
+export interface PurchaseOptionPrice {
```
tonidero, line 82: Hmm considering this has more than the price, I'm thinking about renaming this to `PurchaseOptionPhase`. Wdyt?

src/entities/offerings.ts
```diff
@@ -71,6 +75,57 @@ export interface Price {
...
+ */
+export interface SubscriptionPurchaseOption extends PurchaseOption {
+  /**
+   * The base price for a SubscriptionPurchaseOption, represents
+   * the price that the customer will be charged after all the discounts have
+   * been consumed.
+   */
+  readonly basePrice: PurchaseOptionPrice;
+  /**
+   * The trial PurchaseOptionPrice.
```
tonidero, line 123: We could add links to the other classes, so it's easier to navigate the docs.

src/entities/offerings.ts
```diff
@@ -96,6 +151,23 @@ export interface Product {
    * The offering ID used to obtain this product.
    */
   readonly presentedOfferingIdentifier: string;
+
+  /**
+   * The default option id to be used to subscribe to this product.
+   * Null if no subscription options are available.
```
tonidero, line 157: Hmm I think this should never happen except for in-app products once we add support for them right? I think we should mention that here.

src/entities/offerings.ts
```diff
@@ -198,16 +270,47 @@ const toPrice = (priceData: { amount: number; currency: string }): Price => {
...
+    ([key, value]) => {
+      options[key] = toSubscriptionPurchaseOption(value);
+    },
+  );
+
   return {
     identifier: productDetailsData.identifier,
     displayName: productDetailsData.title,
     currentPrice: toPrice(productDetailsData.current_price),
     normalPeriodDuration: productDetailsData.normal_period_duration,
```
tonidero, line 309: I think for the `currentPrice` and `normalPeriodDuration` maybe we should default to the value of the defaultSubscriptionOption and update the docs on these 2? Seems like the backend returned values should be deprecated (kept only for backwards compatibility). 

This might not work for in app products though, once we add them. So I'm thinking for this version to filter out responses where there is no `subscription_purchase_options` field in the backend response (and add a log). That way, once we add support for one time products, we don't break the parsing on these older SDKs. 

Separately, it might be good to provide an easy accessor for the trial info, even if it's only sintactic sugar. Otherwise it's more difficult to check whether there is a trial.

src/ui/rcb-ui.svelte
```diff
@@ -1,317 +1,323 @@
...
-  const productId = rcPackage.rcBillingProduct?.identifier ?? null;
+    let productDetails: any = null;
+    let brandingInfo: BrandingInfoResponse | null = null;
+    let paymentInfoCollectionMetadata: SubscribeResponse | null = null;
+    let lastError: PurchaseFlowError | null = null;
+    const productId = rcPackage?.rcBillingProduct?.identifier ?? null;
+    const defaultPurchaseOptionId = rcPackage?.rcBillingProduct?.defaultSubscriptionPurchaseOptionId;
+    const optionId = subscriptionPurchaseOptionId ? subscriptionPurchaseOptionId : defaultPurchaseOptionId;
+    const purchaseOption = optionId ?
+        rcPackage?.rcBillingProduct?.subscriptionPurchaseOptions[optionId] : undefined;
```
tonidero, line 50: We should probably log something and fail in the case there is no purchase option I think.

src/networking/responses/products-response.ts
```diff
@@ -1,31 +1,35 @@
-export interface Price {
+export interface PriceResponse {
```
tonidero, line 1: Like these renames! It definitely helps organize it better 👍

tonidero (commented):
Looking good! Might handle some of my comments before getting it merged.

src/networking/backend.ts
vicfergar: I created [this ticket](https://linear.app/revenuecat/issue/BIL-608/%5Btrials%5D-%5Bbackend%5D-accept-base-plan-in-the-purchase-endpoint-as-offer) to make this change.

src/networking/responses/products-response.ts
```diff
@@ -1,31 +1,35 @@
...
+export interface PurchaseOptionResponse {
   id: string;
 }
 
-export interface SubscriptionPurchaseOption extends PurchaseOption {
-  base_price: PurchaseOptionPrice;
-  trial: PurchaseOptionPrice | null;
+export interface SubscriptionPurchaseOptionResponse
+  extends PurchaseOptionResponse {
+  base_price: PurchaseOptionPriceResponse;
```
tonidero, line 18: I'm going to rename the type here to `PurchaseOptionPhaseResponse`, and the public interface to `PurchaseOptionPhase`.

As for these fields, I'm wondering if we should rename them to `base_phase` and `trial_phase`. Wdyt? @vicfergar

vicfergar: Shipped!

src/networking/backend.ts
vicfergar: Shipped!

api-report/purchases-js.api.md
```diff
@@ -176,10 +176,34 @@ export interface Price {
...
+    readonly subscriptionPurchaseOptions: {
+        [optionId: string]: SubscriptionPurchaseOption;
+    };
+}
+
+// @public
+export interface PurchaseFlowParams {
+    customerEmail?: string;
+    htmlTarget?: HTMLElement;
+    purchaseOption?: PurchaseOption | null;
```
tonidero, line 193: I made some extra changes to the API here. So instead of passing an ID, now we pass an entire `PurchaseOption` object. I believe this is preferable to passing strings around.

src/networking/responses/products-response.ts
```diff
@@ -1,31 +1,33 @@
...
+export interface SubscriptionPurchaseOptionResponse
+  extends PurchaseOptionResponse {
+  base_phase: PurchaseOptionPhaseResponse | null;
+  trial_phase: PurchaseOptionPhaseResponse | null;
 }
 
 export interface ProductResponse {
-  current_price: Price;
   identifier: string;
-  normal_period_duration: string;
```
tonidero, line 24: I removed these since I changed the product-level API to use the price and duration from the base phase of the default purchase option. We might need to bring them back once we add support for in-app products, but I think it's better to remove it for now.

api-report/purchases-js.api.md
vegaro: does it need to be nullable and optional? Or only optional works?

package.json
```diff
@@ -52,7 +52,7 @@
     "extract-api": "api-extractor run --local --verbose"
   },
   "devDependencies": {
-    "@microsoft/api-extractor": "^7.40.1",
+    "@microsoft/api-extractor": "7.43.0",
```
vegaro, line 55: should this be "^7.43.0"? or is the pinning intentional

src/entities/offerings.ts
```diff
@@ -85,17 +140,36 @@ export interface Product {
...
    * The offering ID used to obtain this product.
    */
   readonly presentedOfferingIdentifier: string;
+
+  /**
+   * The default subscription option for this product. Null if no subscription
+   * options are available like in the case of consumables and non-consumables.
+   */
+  readonly defaultSubscriptionOption: SubscriptionPurchaseOption | null;
+
```
vegaro, line 162: nit, extra line

src/entities/offerings.ts
```diff
@@ -85,17 +140,36 @@ export interface Product {
...
-   * The period duration for a subscription product.
+   * The period duration for a subscription product. This will match the default
+   * option's base phase period duration.
    */
   readonly normalPeriodDuration: string | null;
   /**
    * The offering ID used to obtain this product.
    */
   readonly presentedOfferingIdentifier: string;
+
```
vegaro, line 156: nit, extra line

src/entities/offerings.ts
```diff
@@ -198,16 +272,91 @@ const toPrice = (priceData: { amount: number; currency: string }): Price => {
   };
 };
 
+const toPurchaseOptionPrice = (
```
vegaro, line 275: I think this is meant to be `toPurchaseOptionPhase`

src/entities/offerings.ts
```diff
@@ -198,16 +272,91 @@ const toPrice = (priceData: { amount: number; currency: string }): Price => {
...
+  optionPrice: PurchaseOptionPhaseResponse,
```
vegaro, line 276: Is this a price as the name infers?

src/entities/offerings.ts
```diff
@@ -198,16 +272,91 @@ const toPrice = (priceData: { amount: number; currency: string }): Price => {
...
+    );
+    return null;
+  }
+
+  console.log(
+    "NORMAL PERIOD DURATION FOR PRODUCT ID: ",
+    productDetailsData.identifier,
+    " IS: ",
+    defaultOption.basePhase.periodDuration,
+  );
```
vegaro, line 351: Is this console log debugging leftovers?

src/entities/purchase-flow-params.ts
```diff
@@ -0,0 +1,24 @@
...
+ */
+export interface PurchaseFlowParams {
+  /**
+   * The package you want to purchase. Obtained from {@link Purchases.getOfferings}.
+   */
+  rcPackage: Package;
+  /**
+   * The option to be used for this purchase. If not specified the default one will be used.
+   */
+  purchaseOption?: PurchaseOption | null;
```
vegaro, line 15: not a JS expert, but wondering if this needs to be nullable as well, or just optional?

src/entities/purchase-flow-params.ts
```diff
@@ -0,0 +1,24 @@
...
+  /**
+   * The HTML element where the billing view should be added. If null, a new div will be created at the root of the page and appended to the body.
+   */
+  htmlTarget?: HTMLElement;
+  /**
+   * The email of the user. If null, RevenueCat will ask the customer for their email.
```
vegaro, line 21: Same here, take it with a grain of salt since I am no expert, but I think this should be "If undefined" instead of "If null"

src/main.ts
```diff
@@ -226,17 +231,38 @@ export class Purchases {
...
    * @param rcPackage - The package you want to purchase. Obtained from {@link Purchases.getOfferings}.
    * @param customerEmail - The email of the user. If null, RevenueCat will ask the customer for their email.
    * @param htmlTarget - The HTML element where the billing view should be added. If null, a new div will be created at the root of the page and appended to the body.
```
vegaro, line 237: can this be null? or only undefined? "If null..."

src/networking/backend.ts
```diff
@@ -55,6 +55,7 @@ export class Backend {
   async postSubscribe(
     appUserId: string,
     productId: string,
+    purchaseOptionId: string | undefined,
```
vegaro, line 58: I think this is the same and more consistent with the rest of the code

```suggestion
    purchaseOptionId?: string,
```

src/networking/backend.ts
```diff
@@ -63,17 +64,24 @@ export class Backend {
       product_id: string;
       email: string;
       presented_offering_identifier: string;
+      offer_id?: string | undefined;
```
vegaro, line 67: Same here, I don't think the `| undefined` is needed

src/entities/offerings.ts
```diff
@@ -71,6 +76,56 @@ export interface Price {
...
+  /**
+   * The base phase for a SubscriptionPurchaseOption, represents
+   * the price that the customer will be charged after all the discounts have
+   * been consumed and the period at which it will renew.
+   */
+  readonly basePhase: PurchaseOptionPhase;
+  /**
+   * The trial information for this subscription option if available.
+   */
+  readonly trialPhase: PurchaseOptionPhase | null;
```
vegaro, line 126: do these two need the "Phase" in the name? or can they just be "base" and "trial"

package.json
tonidero: Yeah it's related to #118

src/entities/offerings.ts
```diff
@@ -71,6 +76,56 @@ export interface Price {
   readonly formattedPrice: string;
 }
 
+/**
+ * Represents the price and duration information for a phase of the purchase option.
+ * @public
+ */
+export interface PurchaseOptionPhase {
```
JayShortway, line 83: In purchases-android, there's a `PricingPhase`, which seems to hold very similar info. If they are indeed describing the same concept, should this be renamed to `PricingPhase` for consistency?

src/entities/offerings.ts
```diff
@@ -71,6 +76,56 @@ export interface Price {
...
+   * The unique id for a purchase option
+   */
+  readonly id: string;
+}
+
+/**
+ * Represents a possible option to purchase a subscription product.
+ * @public
+ */
+export interface SubscriptionPurchaseOption extends PurchaseOption {
```
JayShortway, line 116: Should we name this `SubscriptionOption` for consistency with purchases-android?

src/entities/purchase-flow-params.ts
```diff
@@ -0,0 +1,24 @@
+import { type Package, type PurchaseOption } from "./offerings";
+
+/**
+ * Parameters used to customise the purchase flow when invoking the `.purchase` method.
+ * @public
+ */
+export interface PurchaseFlowParams {
```
JayShortway, line 7: Another consistency suggestion 😅: what about using the purchases-android name, `PurchaseParams`? I think it would make context-switching between the various SDKs easier if concepts are named similarly.

api-report/purchases-js.api.md
tonidero: I made it both for convenience... Basically if you make it only optional but not nullable, you can't make it receive a null value like:
`purchase({ purchaseOption: possiblyNullablePurchaseOption })`... Honestly, I'm not super familiar with TS conventions either, but I thought this was more flexible.

JayShortway (commented):
Looking good! I just had a few consistency/naming suggestions.

src/entities/purchase-flow-params.ts
tonidero: Mentioned my logic here: https://github.com/RevenueCat/purchases-js/pull/117#discussion_r1625794328

src/entities/offerings.ts
tonidero: Good point about all these! No reason to differentiate the name I think. Will rename those, thanks!

src/entities/offerings.ts
tonidero: Hmm I think it's fine to remove that yeah 👍

vegaro (approved):
lgtm!

api-report/purchases-js.api.md
```diff
@@ -173,13 +173,37 @@ export interface Price {
...
+
 // @public
 export interface Product {
     readonly currentPrice: Price;
+    readonly defaultSubscriptionOption: SubscriptionOption | null;
     readonly displayName: string;
     readonly identifier: string;
     readonly normalPeriodDuration: string | null;
     readonly presentedOfferingIdentifier: string;
+    readonly subscriptionOptions: {
```
tonidero, line 191: This is pretty much unneeded for now (the backend should be the one choosing the offer for the SDK), but adding this to make sure the system is flexible if needed to choose between different options from the SDK.

examples/rcbilling-demo/src/pages/paywall/index.tsx
```diff
@@ -21,6 +21,17 @@ const priceLabels: Record<string, string> = {
   P1W: "wk",
 };
 
+const trialLabels: Record<string, string> = {
+  P3D: "3 days",
+  P1W: "1 week",
+  P2W: "2 weeks",
+  P1M: "1 month",
+  P2M: "1 months",
+  P3M: "1 months",
```
jefago, line 30:
```suggestion
  P2M: "2 months",
  P3M: "3 months",
```

src/ui/states/state-present-offer.svelte
```diff
@@ -1,51 +1,81 @@
...
+            {#if !trial?.periodDuration && basePrice }
+                {basePrice.currency || ''} {formatPrice(
+                    basePrice.amount,
+                    basePrice.currency,
+                )}
+            {/if}
+
+        </span>
+        {#if (trial && basePrice)}
+			<span class="rcb-product-price-after-trial">
```
jefago, line 34: Super nit – I think there are some tab formatted lines in here while everything else uses spaces

jefago (commented):
Looks good overall, couple of nits

examples/rcbilling-demo/src/pages/paywall/index.tsx
tonidero: Oh nice catch!

PR #120, opened 2024-06-04
Update vite dependency in RCBilling demo app

tonidero:
## Motivation / Description
Update vite in rcbilling demo app to solve security issue: https://github.com/RevenueCat/purchases-js/security/dependabot/7


JayShortway (approved):
LGTM!

PR #123, opened 2024-06-07
Add processed `period` to `PricingPhase`

tonidero:
## Motivation / Description
Currently, we only expose the periodDuration as a ISO8601 string. While developers can parse that, we can also provide a parsed version of this duration in the `PricingPhase` which includes the duration of the trials and base phases for convenience. This is a followup of trial support in #117

### DIFF
--- a/api-report/purchases-js.api.md
+++ b/api-report/purchases-js.api.md
@@ -164,7 +164,25 @@ export enum PackageType {
 }
 
 // @public
+export interface Period {
+    number: number;
+    unit: PeriodUnit;
+}
+
+// @public
 export type PeriodType = "normal" | "intro" | "trial";
+
+// @public
+export enum PeriodUnit {
+    // (undocumented)
+    Day = "day",
+    // (undocumented)
+    Month = "month",
+    // (undocumented)
+    Week = "week",
+    // (undocumented)
+    Year = "year"
+}
 
 // @public
 export interface Price {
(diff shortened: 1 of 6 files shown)


src/entities/offerings.ts
```diff
@@ -82,10 +83,13 @@ export interface Price {
  */
 export interface PricingPhase {
   /**
-   * The duration of the purchase option price in ISO 8601 format.
-   * For applicable options (trials, initial/promotional prices), otherwise null
```
tonidero, line 86: This is no longer true. We changed the backend so the normal phases also included the period.

src/main.ts
```diff
@@ -59,6 +59,7 @@ export {
   PurchasesError,
   UninitializedPurchasesError,
 } from "./entities/errors";
+export type { Period, PeriodUnit } from "./helpers/duration-helper";
```
tonidero, line 62: We were not exporting these types.

api-report/purchases-js.api.md
```diff
@@ -176,6 +194,7 @@ export interface Price {
 // @public
 export interface PricingPhase {
     readonly cycleCount: number;
+    readonly period: Period | null;
```
tonidero, line 197: This is the convenience accessor added in this PR. We still keep the `periodDuration` as a ISO8601 string.

jefago (approved):
:shipit:

PR #133, opened 2024-06-24
Show better errors in purchase flow when user is subscribed

tonidero:
## Motivation / Description
When a user tries to purchase a product they are already subscribed to, the backend returns an error, but the SDK wasn't handling that well and displayed a generic error. This modifies that to return a more readable error.

image

### DIFF
--- a/src/ui/states/state-error.svelte
+++ b/src/ui/states/state-error.svelte
@@ -37,7 +37,7 @@
       case PurchaseFlowErrorCode.MissingEmailError:
         return "Email is required to complete the purchase.";
       case PurchaseFlowErrorCode.AlreadySubscribedError:
-        return "This user is already subscribed to this product."
+        return "You are already subscribed to this product."
     }
     return lastError.message;
   }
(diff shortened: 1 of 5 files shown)


tonidero:
I wasn't sure whether to also change the "Something went wrong"... But I think changing the error description might be enough... Lmk what you think!

src/ui/states/state-error.svelte
```diff
@@ -36,6 +36,8 @@
         return lastError.message;
       case PurchaseFlowErrorCode.MissingEmailError:
         return "Email is required to complete the purchase.";
+      case PurchaseFlowErrorCode.AlreadySubscribedError:
+        return "This user is already subscribed to this product."
```
jefago, line 40: I would change this to "You are already subscribed to this product", since this is what the customer sees

jefago (commented):
I think we could also change "Something went wrong" to "Already subscribed"

jefago (approved):

PR #136, opened 2024-06-26
Added the possibility to configure some http parameters when initialising the purchases class

nicfix:


src/main.ts
```diff
@@ -120,9 +125,14 @@ export class Purchases {
...
    * @param apiKey - RevenueCat API Key. Can be obtained from the RevenueCat dashboard.
    * @param appUserId - Your unique id for identifying the user.
+   * @param httpConfig - Advanced http configuration to customise the SDK usage {@link HttpConfig}.
    * @throws {@link PurchasesError} if the API key or user id are invalid.
    */
-  static configure(apiKey: string, appUserId: string): Purchases {
+  static configure(
+    apiKey: string,
+    appUserId: string,
+    httpConfig?: HttpConfig,
```
tonidero, line 134: In the mobile SDKs we have the concept of `DangerousSettings` for settings that are not meant to be used by developers unless specifically directed to like: https://github.com/RevenueCat/purchases-android/blob/main/purchases/src/main/kotlin/com/revenuecat/purchases/DangerousSettings.kt.

I'm wondering if we should add these now, and add the `HttpConfig` parameter like one of the fields there. This is mostly as a way to make it even clearer that devs shouldn't use this inadvertently. Wdyt?

src/main.ts
```diff
@@ -131,14 +141,19 @@ export class Purchases {
...
   /** @internal */
-  private constructor(apiKey: string, appUserId: string) {
+  private constructor(
+    apiKey: string,
+    appUserId: string,
+    httpConfig?: HttpConfig,
+  ) {
     this._API_KEY = apiKey;
     this._appUserId = appUserId;
+    this._httpConfig = httpConfig ? httpConfig : { includeCredentials: false };
```
tonidero, line 156: Nitpick but we could also leave the `_httpConfig` to null if it's not passed in? NABD though.

tonidero (commented):
Looks great! Just a couple comments. Lmk when it's final and I can take another look

src/main.ts
nicfix: I added a default value to avoid this ternary expression.

src/main.ts
nicfix: Do we consider the `proxy` field we have in other SDKs dangerous?

I think that these settings are as dangerous as the `proxy`.

src/main.ts
tonidero: Ok that's fine too 👍

tonidero (approved):

PR #147, opened 2024-07-09
Add support for adding additional headers on http requests

tonidero:
## Motivation / Description
This adds support for adding additional headers to all http requests. This can be useful in some situations like when specifying a proxy url (to handle the connections yourself)

### DIFF
--- a/api-report/purchases-js.api.md
+++ b/api-report/purchases-js.api.md
@@ -109,6 +109,7 @@ export enum ErrorCode {
 
 // @public
 export interface HttpConfig {
+    additionalHeaders?: Record<string, string>;
     includeCredentials?: boolean;
 }
 
--- a/api-report/purchases-js.api.md
+++ b/api-report/purchases-js.api.md
@@ -108,6 +108,11 @@ export enum ErrorCode {
 }
 
 // @public
+export interface HttpConfig {
+    includeCredentials?: boolean;
+}
+
+// @public
 export enum LogLevel {
     Debug = 4,
     Error = 1,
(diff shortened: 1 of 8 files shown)


src/networking/http-client.ts
```diff
@@ -103,8 +105,11 @@ function getHeaders(
     "X-Version": VERSION,
     "X-Is-Sandbox": `${isSandboxApiKey(apiKey)}`,
   };
-  if (headers != null) {
+  if (headers) {
     all_headers = { ...all_headers, ...headers };
   }
+  if (additionalHeaders) {
+    all_headers = { ...additionalHeaders, ...all_headers };
```
vicfergar, line 112: Could we change this for consistency?
```suggestion
    all_headers = { ...all_headers, ...additionalHeaders };
```

src/tests/networking/backend.test.ts
```diff
@@ -29,6 +28,89 @@ afterEach(() => {
...
+    let requestPerformed: Request | undefined;
+    server.events.on("request:start", (req) => {
+      requestPerformed = req.request;
+    });
+    await backend.getCustomerInfo("someAppUserId");
+    expect(requestPerformed).not.toBeNull();
+    expect(requestPerformed?.credentials).toEqual("omit");
+  });
+
+  test("credentials is include if passing httpConfig parameter", async () => {
```
vicfergar, line 54:
```suggestion
  test("credentials are included if passing httpConfig parameter", async () => {
```

vicfergar (approved):
I added some minor suggestions, LGTM!

src/tests/networking/backend.test.ts
tonidero: Actually this is correct but I should likely put it between quotes :P. Same for the other one

api-report/purchases-js.api.md
```diff
@@ -107,7 +107,7 @@ export enum ErrorCode {
 
 // @public
 export interface HttpConfig {
-    includeCredentials?: boolean;
```
tonidero, line 110: I removed the `includeCredentials` parameter since we're not sure we want to include it. For now, we will just add the `additionalHeaders` parameter here.

tonidero:
Talked with @vicfergar. We also need to add support to changing the base url. I will add that in this PR and ask for feedback once it's ready

PR #148, opened 2024-07-09
Add support for `proxyURL` to redirect requests

tonidero:
## Motivation / Description
This adds a new property that can be set as part of the `httpConfig` during the `configure` method. This will be used as the base url for all backend requests performed by the SDK.

### DIFF
--- a/src/entities/http-config.ts
+++ b/src/entities/http-config.ts
@@ -7,6 +7,12 @@ export interface HttpConfig {
    * Additional headers to include in all HTTP requests.
    */
   additionalHeaders?: Record<string, string>;
+  /**
+   * Set this property to your proxy URL *only* if you've received a proxy
+   * key value from your RevenueCat contact. This value should never end with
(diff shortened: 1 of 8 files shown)


src/entities/http-config.ts
```diff
@@ -7,6 +7,12 @@ export interface HttpConfig {
    * Additional headers to include in all HTTP requests.
    */
   additionalHeaders?: Record<string, string>;
+  /**
+   * Set this property to your proxy URL *only* if you've received a proxy
+   * key value from your RevenueCat contact. This value should never end with
+   * a trailing slash.
+   */
+  proxyURL?: string;
```
tonidero, line 15: Note that the API differs from how we do it in the mobile SDKs @RevenueCat/coresdk. In those, we set this up as a static variable before `configure`. I felt this made more sense but happy to hear your thoughts

vicfergar (approved):
LGTM!

PR #151, opened 2024-07-22
Change modal border radius to 16px

jefago:
## Motivation / Description
The border radius of modals was not consistent with the designs (8px instead of 16px); this changes that

## Linear ticket (if any)
https://revenuecat.slack.com/archives/C05TUASTSG4/p1721591753803999


tonidero (approved):
LGTM! (I added the `refactor` label so it passes the test since `enhancement` is not one of the allowed labels

PR #152, opened 2024-07-23
Add `presentedOfferingContext` type to include all context information

tonidero:
## Motivation / Description
This will allow us to track targeting information on purchases performed with purchases-js.

### DIFF
--- a/src/entities/offerings.ts
+++ b/src/entities/offerings.ts
@@ -1,6 +1,7 @@
 import {
   type OfferingResponse,
   type PackageResponse,
+  type TargetingResponse,
 } from "../networking/responses/offerings-response";
 import {
   type ProductResponse,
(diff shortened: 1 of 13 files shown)


src/entities/offerings.ts
```diff
@@ -381,12 +418,24 @@ const toPackage = (
...
 ): Offering | null => {
+  const presentedOfferingContext: PresentedOfferingContext = {
+    offeringIdentifier: offeringsData.identifier,
+    targetingContext:
+      isCurrentOffering && targetingResponse
+        ? {
+            ruleId: targetingResponse.rule_id,
+            revision: targetingResponse.revision,
+          }
+        : null,
```
antoniobg, line 434: This is probably a stupid question, but why do we only set `targetingContext` for the current offering?

antoniobg (approved):

src/entities/offerings.ts
tonidero: Targeting works by modifying the `current` offering depending on certain parameters (country, sdk version,....). So the targeting information is only relevant to that offering. The developer can always choose to purchase a package in a different offering, but in that case, the targeting information doesn't matter, since the rules applied didn't affect anything.

There is one thing to note that is different here vs the mobile SDKs. That is, when accessing the `current` offering, not through the `current` accessor but through the `all` dictionary, we are adding targeting information in `purchases-js`, but not in the mobile SDKs. This is needed to support the hosted checkout page, but is a distinction we should be aware of. cc @joshdholtz

PR #160, opened 2024-08-20
Improve formatted price calculation

tonidero:
## Motivation / Description
Some changes to the formatted price calculation in preparatio for multi-currency
- Removes hardcoded `en-US` locale. Now we will leave `Intl` to get the locale
- Cleanup to use `formattedPrice` instead of calculating it again.

### DIFF
--- a/examples/rcbilling-demo/src/tests/main.test.ts
+++ b/examples/rcbilling-demo/src/tests/main.test.ts
@@ -14,7 +14,7 @@ test.describe("Main", () => {
     // Gets all elements that match the selector
     const packageCards = await getAllElementsByLocator(page, CARD_SELECTOR);
 
-    const EXPECTED_VALUES = [/30\.00/, /19\.99/, /15\.00/];
+    const EXPECTED_VALUES = [/30[,.]00/, /19[,.]99/, /15[,.]00/];
 
     await Promise.all(
       packageCards.map(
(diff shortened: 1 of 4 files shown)


examples/rcbilling-demo/src/tests/main.test.ts
```diff
@@ -14,7 +14,7 @@ test.describe("Main", () => {
     // Gets all elements that match the selector
     const packageCards = await getAllElementsByLocator(page, CARD_SELECTOR);
 
-    const EXPECTED_VALUES = [/30\.00/, /19\.99/, /15\.00/];
+    const EXPECTED_VALUES = [/30[,.]00/, /19[,.]99/, /15[,.]00/];
```
tonidero, line 17: Did this to also account for other locales where commas are used as decimal delimiters.

src/helpers/price-labels.ts
```diff
@@ -10,9 +10,13 @@ export const priceLabels: Record<string, string> = {
...
 
-export const formatPrice = (priceInCents: number, currency: string): string => {
+export const formatPrice = (
+  priceInCents: number,
+  currency: string,
+  locale?: string,
+): string => {
   const price = priceInCents / 100;
-  const formatter = new Intl.NumberFormat("en-US", {
+  const formatter = new Intl.NumberFormat(locale, {
```
tonidero, line 19: Now this will default to `undefined`, leaving it to `Intl` to get the locale

guido732 (approved):

PR #161, opened 2024-08-20
[Beta] Support different currencies

tonidero:
## Motivation / Description
There is a new optional parameter in the `getOfferings` method. This will allow to add future params. Currently, only a `currency` parameter is optional and allows to select the currency.


vicfergar (approved):
LGTM!

nicfix (approved):

PR #162, opened 2024-08-20
Parse `price_id` and pass it back on subscribe

tonidero:
## Motivation / Description
This PR is based on #161. It has a few changes to allow performing purchases in different currencies.

- It adds a `defaultPurchaseOption` to the `StoreProduct`, this will be not null so it's also available for one-time products.
- Adds the `priceId` to the `PurchaseOption` interface.
- Parses the `price_id` from the `/products` response into the `PurchaseOption` interface.
- Passes the `price_id` to the `POST /subscribe` endpoint

### DIFF
--- a/src/networking/backend.ts
+++ b/src/networking/backend.ts
@@ -14,7 +14,10 @@ import { type ProductsResponse } from "./responses/products-response";
 import { type BrandingInfoResponse } from "./responses/branding-response";
 import { type CheckoutStatusResponse } from "./responses/checkout-status-response";
 import { defaultHttpConfig, type HttpConfig } from "../entities/http-config";
-import { type PresentedOfferingContext } from "../entities/offerings";
+import type {
+  PresentedOfferingContext,
+  PurchaseOption,
+} from "../entities/offerings";
 
 export class Backend {
   private readonly API_KEY: string;
@@ -74,7 +77,7 @@ export class Backend {
     productId: string,
     email: string,
     presentedOfferingContext: PresentedOfferingContext,
-    purchaseOptionId?: string,
+    purchaseOption: PurchaseOption,
   ): Promise<SubscribeResponse> {
     type SubscribeRequestBody = {
       app_user_id: string;
@@ -82,6 +85,7 @@ export class Backend {
       email: string;
       presented_offering_identifier: string;
       offer_id?: string;
+      price_id: string;
       applied_targeting_rule?: {
         rule_id: string;
         revision: number;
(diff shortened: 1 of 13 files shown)


src/networking/backend.ts
```diff
@@ -92,12 +96,13 @@ export class Backend {
...
       email: email,
+      price_id: purchaseOption.priceId,
       presented_offering_identifier:
         presentedOfferingContext.offeringIdentifier,
     };
 
-    if (purchaseOptionId && purchaseOptionId !== "base_option") {
-      requestBody.offer_id = purchaseOptionId;
+    if (purchaseOption.id !== "base_option") {
+      requestBody.offer_id = purchaseOption.id;
```
tonidero, line 105: Not sure if we should still filter out the `base_option`... But decided to keep the same logic for now.

src/ui/rcb-ui.svelte
```diff
@@ -43,8 +43,8 @@
     let brandingInfo: BrandingInfoResponse | null = null;
     let paymentInfoCollectionMetadata: SubscribeResponse | null = null;
     let lastError: PurchaseFlowError | null = null;
-    const productId = rcPackage?.rcBillingProduct?.identifier ?? null;
-    const defaultPurchaseOption = rcPackage?.rcBillingProduct?.defaultSubscriptionOption;
+    const productId = rcPackage.rcBillingProduct.identifier ?? null;
+    const defaultPurchaseOption = rcPackage.rcBillingProduct.defaultPurchaseOption;
```
tonidero, line 47: This should never be null now.

src/ui/states/state-present-offer.svelte
```diff
@@ -21,13 +21,13 @@
                 {getTrialsLabel(trial.periodDuration)} free trial
             {/if}
             {#if !trial?.periodDuration && basePrice }
-                {basePrice.currency || ''} {basePrice.formattedPrice}
```
tonidero, line 24: I added these changes as well. The `formattedPrice` already includes the currency so it was displaying duplicate in the UI. This removes it so we only display the currency as it comes from the `Intl.NumberFormat` utility.

src/networking/responses/products-response.ts
```diff
@@ -11,6 +11,7 @@ export interface PricingPhaseResponse {
 
 export interface PurchaseOptionResponse {
   id: string;
+  price_id: string;
```
nicfix, line 14: This is theoretically optional in the backend response, but it's never optional in practice, I'll double check if I can make it mandatory later on.

nicfix (approved):
LGTM!

Can we avoid shipping this new version until the entire multi currency is shipped?

tonidero:
> Can we avoid shipping this new version until the entire multi currency is shipped?

As mentioned, we will ship this, but mention it's on BETA in the CHANGELOG until it's ready

PR #164, opened 2024-08-26
Make main CTA not say "Pay" if the customer is starting a trial

jefago:
## Motivation / Description

https://discord.com/channels/1121319986699440218/1220385785438670958/1277027263929516176

The main purchase CTA says "Pay" when using a trial, this changes it to say "Start Trial" instead.

### Before

image

### After
image

### DIFF
--- a/examples/rcbilling-demo/src/tests/main.test.ts
+++ b/examples/rcbilling-demo/src/tests/main.test.ts
@@ -86,7 +86,7 @@ async function enterCreditCardDetailsAndContinue(page: Page): Promise<void> {
   await stripeFrame.getByPlaceholder("CVC").fill("123");
   await stripeFrame.getByLabel("Country").selectOption("US");
   await stripeFrame.getByPlaceholder("12345").fill("12345");
-  await page.getByRole("button", { id: "Pay" }).click();
+  await page.getByTestId("PayButton").click();
 }
 
 async function navigateToUrl(page: Page, userId: string): Promise<void> {
--- a/examples/rcbilling-demo/src/tests/main.test.ts
+++ b/examples/rcbilling-demo/src/tests/main.test.ts
@@ -86,7 +86,7 @@ async function enterCreditCardDetailsAndContinue(page: Page): Promise<void> {
   await stripeFrame.getByPlaceholder("CVC").fill("123");
   await stripeFrame.getByLabel("Country").selectOption("US");
   await stripeFrame.getByPlaceholder("12345").fill("12345");
-  await page.getByRole("button", { name: "Pay" }).click();
+  await page.getByRole("button", { id: "Pay" }).click();
 }
 
 async function navigateToUrl(page: Page, userId: string): Promise<void> {
--- a/src/ui/rcb-ui.svelte
+++ b/src/ui/rcb-ui.svelte
@@ -1,56 +1,56 @@
 <script lang="ts">
-    import { onMount } from "svelte";
-    import type { Package, Product, PurchaseOption, Purchases } from "../main";
-    import StatePresentOffer from "./states/state-present-offer.svelte";
-    import StateLoading from "./states/state-loading.svelte";
-    import StateError from "./states/state-error.svelte";
-    import StateSuccess from "./states/state-success.svelte";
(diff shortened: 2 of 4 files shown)


src/ui/rcb-ui.svelte
```diff
@@ -1,323 +1,334 @@
...
+          {/if}
+          {#if paymentInfoCollectionMetadata && (state === "needs-payment-info" || state === "polling-purchase-status") && productDetails && purchaseOptionToUse}
+            <StateNeedsPaymentInfo
+              {paymentInfoCollectionMetadata}
+              onContinue={handleContinue}
+              onClose={handleClose}
+              onError={handleError}
+              processing={state === "polling-purchase-status"}
+              {productDetails}
+              {purchaseOptionToUse}
```
jefago, line 240: These 2 lines are the only meaningful change in this file, the rest is just automatically fixed indentation

vicfergar (approved):

src/ui/states/state-needs-payment-info.svelte
```diff
@@ -125,10 +128,12 @@
             <Button disabled={processing}>
               {#if processing}
                 <ProcessingAnimation />
+              {:else if productDetails.subscriptionOptions[purchaseOptionToUse.id].trial !== null}
```
tonidero, line 131: Hmm instead of having to pass the product to this view, we could try to downcast the `purchaseOption` to a `subscriptionOption`, like:
```suggestion
              {:else if (purchaseOptionToUse as SubscriptionOption).trial !== undefined}
```
That way, we don't need to pass in the `productDetails`. Wdyt?

src/ui/states/state-needs-payment-info.svelte
```diff
@@ -17,12 +17,15 @@
   import IconLock from "../assets/icon-lock.svelte";
   import { Colors } from "../../assets/colors";
   import ProcessingAnimation from "../processing-animation.svelte";
+  import { Product, PurchaseOption } from "../../main";
```
tonidero, line 20: I think we should import these from the model classes. Less chances of some circular reference.
```suggestion
  import { Product, PurchaseOption } from "../../entities/offerings";
```

tonidero (commented):
Just a couple comments but looking good!

tonidero:
We will need to fix the e2e tests. Seems they relied on the `Pay` button. Maybe we can click it using the id? Or the new `Start trial` text since the tests would always be getting that right now.

jefago:
> We will need to fix the e2e tests. Seems they relied on the `Pay` button. Maybe we can click it using the id? Or the new `Start trial` text since the tests would always be getting that right now.

Yeah I pushed a commit to fix the test

src/ui/states/state-needs-payment-info.svelte
jefago: I was going back and forth on this. I think the disadvantage there is that we might have to change that back when we implement one-time purchases, so I think I would leave it as is. I am making the condition more robust by adding optional chaining though so this doesn't break in the future if any of the fields are null (eg. if you have a one-time product that doesn't have subscription options)

tonidero: Ok I think this is fine for now 👍

examples/rcbilling-demo/src/tests/main.test.ts
```diff
@@ -86,7 +86,7 @@ async function enterCreditCardDetailsAndContinue(page: Page): Promise<void> {
   await stripeFrame.getByPlaceholder("CVC").fill("123");
   await stripeFrame.getByLabel("Country").selectOption("US");
   await stripeFrame.getByPlaceholder("12345").fill("12345");
-  await page.getByRole("button", { name: "Pay" }).click();
+  await page.getByRole("button", { id: "Pay" }).click();
```
tonidero, line 89: From looking at Playwright's documentation, not sure if we can reference an id directly :/. I did see you can reference by `data-testid`: https://playwright.dev/docs/locators#locate-by-test-id.

Lmk if you would like some help on this!

jefago: ha I was just lazy (and had to jump on a call). I will make another attempt at fixing and also verify locally

PR #166, opened 2024-08-29
Load branding before displaying UI

tonidero:
## Motivation / Description
There is a small jump when loading the billing flow UI, this is because it has to download the branding info after already displaying some info. This loads the branding during SDK configuration. In case it's not downloaded by the time we present the purchase flow, we will try to download it again.

### Before

https://github.com/user-attachments/assets/54bea295-8e9d-46af-82a4-a50f0364ce87

### After

https://github.com/user-attachments/assets/e7ca33cd-a3ec-4400-afd5-06c15d2837ad

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -40,6 +40,7 @@ import { type PurchaseParams } from "./entities/purchase-params";
 import { defaultHttpConfig, type HttpConfig } from "./entities/http-config";
 import { type GetOfferingsParams } from "./entities/get-offerings-params";
 import { validateCurrency } from "./helpers/validators";
+import type { BrandingInfoResponse } from "./networking/responses/branding-response";
 
 export type {
   Offering,
@@ -310,13 +311,22 @@ export class Purchases {
       `Presenting purchase form for package ${rcPackage.identifier}`,
     );
 
-    return new Promise((resolve, reject) => {
+    return new Promise(async (resolve, reject) => {
+      let brandingInfo: BrandingInfoResponse | null = null;
+      try {
+        brandingInfo = await this.backend.getBrandingInfo();
+      } catch (e) {
+        Logger.errorLog(
+          "Error fetching branding info. This is not fatal, but the purchase flow may not look as expected.",
+        );
+      }
       new RCPurchasesUI({
         target: certainHTMLTarget,
         props: {
           appUserId,
           rcPackage,
           purchaseOption,
+          brandingInfo,
           customerEmail,
           onFinished: async () => {
             Logger.debugLog("Purchase finished");
@@ -336,7 +346,6 @@ export class Purchases {
             reject(PurchasesError.getForPurchasesFlowError(e));
           },
           purchases: this,
-          backend: this.backend,
           purchaseOperationHelper: this.purchaseOperationHelper,
           asModal,
         },
--- a/src/main.ts
+++ b/src/main.ts
@@ -87,6 +87,9 @@ export class Purchases {
   private _appUserId: string;
 
   /** @internal */
+  private _brandingInfo: BrandingInfoResponse | null = null;
+
+  /** @internal */
   private readonly backend: Backend;
 
   /** @internal */
@@ -171,6 +174,20 @@ export class Purchases {
     }
     this.backend = new Backend(this._API_KEY, httpConfig);
     this.purchaseOperationHelper = new PurchaseOperationHelper(this.backend);
+    this.initialize();
+  }
+
+  private initialize() {
+    this.backend
+      .getBrandingInfo()
+      .then((brandingInfo) => {
+        this._brandingInfo = brandingInfo;
+      })
+      .catch(() => {
+        Logger.errorLog(
+          "Error fetching branding info. This is not fatal, but the purchase flow may not look as expected.",
+        );
+      });
   }
 
   /** @internal */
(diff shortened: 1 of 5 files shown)


src/main.ts
```diff
@@ -310,13 +311,22 @@ export class Purchases {
       `Presenting purchase form for package ${rcPackage.identifier}`,
     );
 
-    return new Promise((resolve, reject) => {
+    return new Promise(async (resolve, reject) => {
+      let brandingInfo: BrandingInfoResponse | null = null;
+      try {
+        brandingInfo = await this.backend.getBrandingInfo();
```
tonidero, line 317: We might want to display some sort of loading UI for this... It should be pretty fast though so maybe not needed?

src/main.ts
```diff
@@ -170,6 +174,20 @@ export class Purchases {
     }
     this.backend = new Backend(this._API_KEY, httpConfig);
     this.purchaseOperationHelper = new PurchaseOperationHelper(this.backend);
+    this.initialize();
+  }
+
+  private initialize() {
+    this.backend
+      .getBrandingInfo()
```
tonidero, line 182: Note this will cause a big increase in requests to this endpoint

src/main.ts
```diff
@@ -310,13 +328,24 @@ export class Purchases {
       `Presenting purchase form for package ${rcPackage.identifier}`,
     );
 
-    return new Promise((resolve, reject) => {
+    return new Promise(async (resolve, reject) => {
+      if (this._brandingInfo === null) {
```
tonidero, line 332: I thought to add this as a fallback, in case there is an issue downloading initially

src/main.ts
```diff
@@ -85,6 +86,9 @@ export class Purchases {
   /** @internal */
   private _appUserId: string;
 
+  /** @internal */
+  private _brandingInfo: BrandingInfoResponse | null = null;
```
tonidero, line 90: This is now cached in memory, but I believe it should be ok, since this shouldn't change right now (except maybe with a different API key)

src/main.ts
nicfix: Ouch, the `initialize()` method is sync.

Using `.getBrandingInfo().then` inside it might cause some weird scenario in which the `purchase` method is called before `.getBrandingInfo()` is loaded. It's rare but it might happen especially in cases of low/slow connectivity.

Maybe we should define an explicit method to preload the branding async or make a breaking change in which we convert `initialize` to async.

Wdyt?

tonidero: Note that `initialize` is private and is currently called in the `constructor`. I guess we could make `configure` (which is the one that calls this) async... I just worry about delaying initialization for a secondary task like getting branding info, since I would assume devs might call the `configure` method very early in their app.

I was thinking that the worst thing that would happen would be that we fetch branding info multiple times right? I don't think that would be a big deal, specially since it shouldn't cause a race condition since the response should be the same... Alternatively, we could write some sort of cancellation or hook to the existing request if any so we only do it once but I feel it's overkill. Wdyt?

nicfix: I would personally prefer to convert `.configure` in async if possible.
Working with hooks and such will make it really dirty compared to the built in solution from the language. And we will then have to maintain that "custom" implementation going forward.

Regarding the fact that branding is a secondary task, I think that now we are having just the branding but later we might have some more stuff to download so moving to `async` now might make our life easier later.

It would be a breaking change, but we could introduce it together with the customisation release (when it will come) to sweeten the change.

For every developer that calls it very early in the app (and doesn't really care about the result) they would just not implement the `.then()` which will have the same behaviour you just proposed. It will eventually load and it might cause a race condition in rare scenarios.

But at least we keep it clean and we have a clearly documented change that states that devs should consider `.configure` async from now on.

So worst case it works as well as in the solution you proposed, but it gives us freedom to load whatever we want in the `.configure` going forward. We are still in beta, some breaking changes will happen.

tonidero: Hmm I'm not so worried about the breaking change but about causing the SDK to be harder to use... Right now the `configure` call will fill the static `Purchases.sharedInstance`, which would be delayed by this change... If someone doesn't want to wait for the SDK to initialize and then try to call the SDK using that static accessor, it might still not be configured. Even if they don't use the static accessor (and they don't want to wait for the `configure` to finish) and want to use the return value of the `configure` being now a `Promise<Purchases>`, they would need to store the value somewhere else that they can check on whether the SDK has already been configured leading to several possible issues IMO.

> later we might have some more stuff to download

That's a valid concern, but for what is worth, most of our mobile SDKs methods are async, but the `configure` method is synchronous. We do prefetch a bunch of information during the `configure` in the background. To solve the issue of devs asking for information before this is done, we currently have some deduping logic for requests, which we might be able to add to purchases-js.

nicfix: What about we make this a topic for the next grooming session?

We can have a chat tomorrow morning or Monday and prepare it to start working on eventual changes on the next cycle (starting Monday).

I created this ticket as a reference for the grooming:
https://linear.app/revenuecat/issue/ECO-1396/[rcbilling]-[sdk]-preloading-branding-and-configuration-at-configure

tonidero: Agreed! This PR is not urgent, so makes sense to talk more about it. Thanks!

tonidero:
Closing in favor of #169

PR #167, opened 2024-08-29
ECO-1398: Added the possibility to pass a currency in the demo app query string params

nicfix:
## Motivation / Description
testing the new multi-currency setup through the demo app
## Changes introduced
* Added the currency parameter in the loadPurchases loader function
## Linear ticket (if any)
ECO-1398

### DIFF
--- a/examples/rcbilling-demo/src/util/PurchasesLoader.ts
+++ b/examples/rcbilling-demo/src/util/PurchasesLoader.ts
@@ -16,8 +16,12 @@ type IPurchasesLoaderData = {
 
 const loadPurchases: LoaderFunction<IPurchasesLoaderData> = async ({
   params,
+  request,
 }) => {
   const appUserId = params["app_user_id"];
+  const searchParams = new URL(request.url).searchParams;
+  const currency = searchParams.get("currency");
+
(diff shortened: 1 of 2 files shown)


examples/rcbilling-demo/src/util/PurchasesLoader.ts
```diff
@@ -32,7 +36,7 @@ const loadPurchases: LoaderFunction<IPurchasesLoaderData> = async ({
 
     const [customerInfo, offerings] = await Promise.all([
       purchases.getCustomerInfo(),
-      purchases.getOfferings(),
+      purchases.getOfferings({ currency: currency || "USD" }),
```
tonidero, line 39: Should we default to `undefined`? That way we allow the SDK to use the default.

tonidero (approved):
LGTM! 

Note: I changed the labels since I didn't want this to show as a new feature in the changelog.

examples/rcbilling-demo/src/util/PurchasesLoader.ts
nicfix: Changing it!

PR #169, opened 2024-09-02
Create a `preload` method in `Purchases` to prefetch branding information.

tonidero:
## Motivation / Description
This PR adds a new `preload` method to `Purchases`. It will prefetch information that we don't want to do during the `configure` call due to its async nature. Currently only the branding information is prefetched.

Additionally, it adds a new decorator `requiresLoadedResources`. Any method marked with this decorator will load resources in the SDK first if it hasn't loaded already.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -15719,6 +15719,7 @@
       "resolved": "https://registry.npmjs.org/svelte/-/svelte-4.2.19.tgz",
       "integrity": "sha512-IY1rnGr6izd10B0A8LqsBfmlT5OILVuZ7XsI0vdGPEvuonFV7NYEUK4dAkm9Zg2q0Um92kYjTpS1CAP3Nh/KWw==",
       "dev": true,
+      "license": "MIT",
       "dependencies": {
         "@ampproject/remapping": "^2.2.1",
         "@jridgewell/sourcemap-codec": "^1.4.15",
@@ -15740,15 +15741,14 @@
       }
     },
     "node_modules/svelte-check": {
-      "version": "3.6.8",
-      "resolved": "https://registry.npmjs.org/svelte-check/-/svelte-check-3.6.8.tgz",
-      "integrity": "sha512-rhXU7YCDtL+lq2gCqfJDXKTxJfSsCgcd08d7VWBFxTw6IWIbMWSaASbAOD3N0VV9TYSSLUqEBiratLd8WxAJJA==",
+      "version": "3.8.6",
+      "resolved": "https://registry.npmjs.org/svelte-check/-/svelte-check-3.8.6.tgz",
+      "integrity": "sha512-ij0u4Lw/sOTREP13BdWZjiXD/BlHE6/e2e34XzmVmsp5IN4kVa3PWP65NM32JAgwjZlwBg/+JtiNV1MM8khu0Q==",
       "dev": true,
+      "license": "MIT",
       "dependencies": {
         "@jridgewell/trace-mapping": "^0.3.17",
         "chokidar": "^3.4.1",
-        "fast-glob": "^3.2.7",
-        "import-fresh": "^3.2.1",
         "picocolors": "^1.0.0",
         "sade": "^1.7.4",
         "svelte-preprocess": "^5.1.3",
--- a/package-lock.json
+++ b/package-lock.json
@@ -2177,13 +2177,14 @@
       }
     },
     "node_modules/@esbuild/aix-ppc64": {
-      "version": "0.20.2",
-      "resolved": "https://registry.npmjs.org/@esbuild/aix-ppc64/-/aix-ppc64-0.20.2.tgz",
-      "integrity": "sha512-D+EBOJHXdNZcLJRBkhENNG8Wji2kgc9AZ9KiPr1JuZjsNtyHzrsfLRrY0tk2H2aoFu6RANO1y1iPPUCDYWkb5g==",
+      "version": "0.21.5",
+      "resolved": "https://registry.npmjs.org/@esbuild/aix-ppc64/-/aix-ppc64-0.21.5.tgz",
+      "integrity": "sha512-1SDgH6ZSPTlggy1yI6+Dbkiz8xzpHJEVAlF/AM1tHPLsf5STom9rwtjE4hKAF20FfXXNTFqEYXyJNWh1GiZedQ==",
       "cpu": [
         "ppc64"
       ],
       "dev": true,
+      "license": "MIT",
       "optional": true,
       "os": [
         "aix"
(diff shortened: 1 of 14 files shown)


package-lock.json
tonidero, line 1: Had to update `vite` to be able to work with decorators. This is the result of running `npm update vite`

src/helpers/decorators.ts
```diff
@@ -0,0 +1,21 @@
+import { type Purchases } from "../main";
+
+export function requiresInitialization<
+  This extends Purchases,
+  Args extends never[],
+  Result,
+>(
+  target: (this: This, ...args: Args) => Promise<Result>,
```
tonidero, line 8: Making the result a `Promise<Result>` I make sure that this is only used for async functions

src/main.ts
```diff
@@ -151,6 +159,44 @@ export class Purchases {
...
+   * Currently only prefetching branding information. You can call this method
+   * after configuring the SDK to speed up the first call to
+   * {@link Purchases.purchase}.
+   */
+  public async initialize(): Promise<void> {
+    if (this.isInitialized()) {
+      Logger.verboseLog("Purchases is already initialized. Skipping.");
+      return;
+    }
+    if (this._initializationPromise !== null) {
```
tonidero, line 173: In a multithreading context, this would be to be much better handled, but JS is single threaded, so it shouldn't cause any issues I believe?

src/ui/states/state-needs-payment-info.svelte
```diff
@@ -17,7 +17,7 @@
   import IconLock from "../assets/icon-lock.svelte";
   import { Colors } from "../../assets/colors";
   import ProcessingAnimation from "../processing-animation.svelte";
-  import { Product, PurchaseOption } from "../../entities/offerings";
+  import type { Product, PurchaseOption } from "../../entities/offerings";
```
tonidero, line 20: This is to fix a small warning I saw while building the SDK.

src/tests/main.test.ts
```diff
@@ -458,3 +458,10 @@ describe("Purchases.close()", () => {
     expect(Purchases.isConfigured()).toBeFalsy();
   });
 });
+
+describe("Purchases.initialize()", () => {
+  test("can initialize without failing", async () => {
+    const purchases = configurePurchases();
+    await purchases.initialize();
```
tonidero, line 465: I would like to write better tests for this, but currently the side effects of this operation stay inside the `purchases` object, and are only noticeable in the `purchase` method, which is not very testable through unit tests since it mostly displays a UI...

vitest.config.js
```diff
@@ -3,6 +3,10 @@ import { svelte } from "@sveltejs/vite-plugin-svelte";
 
 export default defineConfig({
   plugins: [svelte()],
+
+  // This is needed to make test pass after adding decorators in Purchases.
+  esbuild: { target: "es2022" },
```
tonidero, line 8: This was needed for tests but not for the actual vite config for some reason.

src/tests/test-responses.ts
```diff
@@ -231,5 +239,11 @@ export function getRequestHandlers(): RequestHandler[] {
     );
   });
 
+  requestHandlers.push(
```
tonidero, line 242: I want to refactor the tests so they are better split instead of having this monster config... Originally I did it like this to avoid repetition, but we should be refactoring this soon

src/main.ts
```diff
@@ -281,6 +327,7 @@ export class Purchases {
    * @returns a Promise for the customer info after the purchase is completed successfully.
    * @throws {@link PurchasesError} if there is an error while performing the purchase. If the {@link PurchasesError.errorCode} is {@link ErrorCode.UserCancelledError}, the user cancelled the purchase.
    */
+  @requiresInitialization
```
tonidero, line 330: Note that currently only this method is decorated, since it's the only one that right nows requires "initialization".

src/main.ts
```diff
@@ -151,6 +159,44 @@ export class Purchases {
     return Purchases.getSharedInstance();
   }
 
+  /**
+   * Initializes the Purchases SDK by prefetching some optional data.
+   * Currently only prefetching branding information. You can call this method
+   * after configuring the SDK to speed up the first call to
+   * {@link Purchases.purchase}.
+   */
+  public async initialize(): Promise<void> {
```
tonidero, line 168: This conceptually introduces a new state to the SDK which I don't fully like... Honestly, I'm thinking about renaming this to something like `preloadResources` or something like that, so it's clearer what this does... Wdyt?

src/helpers/decorators.ts
```diff
@@ -0,0 +1,21 @@
+import { type Purchases } from "../main";
```
tonidero, line 1: Only importing the type to avoid a circular reference

src/main.ts
guido732: correct, shouldn't unless we're running different promises in parallel which I don't think we're doing anywhere

guido732 (approved):

src/main.ts
tonidero: I actually think I will do this rename... I feel this name could be confusing to what this does/we're planning to do in here (prefetching some data)

tonidero: Renamed in https://github.com/RevenueCat/purchases-js/pull/169/commits/02910315d76077b7c1c86c0752b7bfdf0c49be04

src/helpers/decorators.ts
nicfix: Nice one, TIL!

nicfix (commented):
LGTM, 

Can you add a test in which:
1. We call `.configure`
2. We **don't** call `.loadResources`
3. We call any other async method with the decorator (say `.getOfferings`)
4. We check that `.loadResources` was anyway called before the `.getOfferings`
5. We check that the internal fields are populated

nicfix:
Nit: I am not sure about the `.loadResources` method name. But I am never sure about names, not even the ones I choose.

My proposals would be:
* `.preloadAssets`
* `.preload`
* The infamous `.init`


But I really don't have a super strong opinion on this one 🙂

tonidero:
Hmm I think I could go with `preload`, since we might be loading other things other than assets... Will rename to that

src/tests/base.purchases_test.ts
```diff
@@ -0,0 +1,26 @@
...
+import { setupServer } from "msw/node";
+import { APIGetRequest, getRequestHandlers } from "./test-responses";
+import { afterAll, beforeAll, beforeEach } from "vitest";
+
+export const testApiKey = "rcb_test_api_key";
+export const testUserId = "someAppUserId";
+
+const server = setupServer(...getRequestHandlers());
+
+beforeAll(() => {
```
tonidero, line 11: I refactored these to another file, to avoid repetition and having all tests in the same file.

src/tests/purchase.test.ts
```diff
@@ -0,0 +1,36 @@
...
+    const purchases = configurePurchases();
+    const expectedRequest: GetRequest = {
+      url: "http://localhost:8000/rcbilling/v1/branding",
+    };
+    const offerings = await purchases.getOfferings();
+    const packageToBuy = offerings.current?.availablePackages[0];
+    expect(APIGetRequest).not.toHaveBeenCalledWith(expectedRequest);
+    // Currently we hold on the purchase UI, so we add a timeout to not hold the test forever.
+    // We're just checking that the request happened as expected.
+    await Promise.race([
```
tonidero, line 17: I had to do this, since the `purchase` method is stuck waiting for the email to be inputted... let me know if you have other suggestions.

nicfix: I think this is the only way to do it for now.
If we would eventually have another method that uses this decorator then we can change it to that one.

But for now it's fine

nicfix (approved):
Thanks for adding the test and renaming the method/decorator to `preload`

PR #170, opened 2024-09-03
Add `amountMicros` to `Price` interface

tonidero:
## Motivation / Description

This adds a new `amountMicros` field to `Price`. It keeps the previous `amount` field but deprecates it.

Need to hold until the backend part of this is deployed.


nicfix (approved):
LGTM.

Could you just double check if the demo app uses the formatted price or the actual amount value?

Looks like the e2e tests are failing, maybe is connected to that or maybe is just waiting for khepri changes to be deployed

tonidero:
Confirmed it works as expected and tests pass after deploying backend changes :)

PR #173, opened 2024-09-11
Improve purchase flow error handling for unknown errors

tonidero:
## Motivation / Description
We changed the handling of errors [here](https://github.com/RevenueCat/purchases-js/pull/143/files#diff-c6863a517751fdfbb4cc9ebc293e2e1f0777ed9ef8a8ef7bb7c28a29d7f8b514R168), so all errors during the subscription request showed in the same panel, however, we were not handling unknown error messages well there. 

This makes it so we only display the errors in the email input screen when it's a known recoverable error (missing/wrong email + network error at the moment)


src/entities/errors.ts
```diff
@@ -112,6 +112,7 @@ export class ErrorCodeUtils {
   ): ErrorCode {
     switch (backendErrorCode) {
       case BackendErrorCode.BackendStoreProblem:
+      case BackendErrorCode.BackendPaymentGatewayGenericError:
```
tonidero, line 115: This is to try to assign this error code to a know error instead of `Unknown`. Seemed like `StoreProblemError` fit the best...

src/ui/rcb-ui.svelte
```diff
@@ -164,7 +164,7 @@
   };
 
   const handleError = (e: PurchaseFlowError) => {
-    if (state === "processing-auth-info") {
+    if (state === "processing-auth-info" && e.isRecoverable()) {
```
tonidero, line 167: This is the actual fix for the issue.

PR #186, opened 2024-09-18
Add additional error details and display

tonidero:
## Motivation / Description
This adds a new field to the public `PurchasesError` type, `extra` that will contain some extra information about certain types of errors. This is used now to pass the backend error code through the purchases flow and display it in the error screen.

image


nicfix (approved):
Looks great, thanks for going the extra mile with this!

PR #189, opened 2024-09-19
Add support for `placements`

tonidero:
## Motivation / Description
See https://www.revenuecat.com/docs/tools/targeting/placements

This adds a new API under `Purchases.getCurrentOfferingForPlacement` that allows to obtain an offering by placement.


tonidero:
Confirmed this attaches the information correctly and events are properly populated in the dashboard

src/helpers/offerings-parser.ts
```diff
@@ -0,0 +1,158 @@
...
+  if (offering == undefined) {
+    return null;
+  }
+
+  return {
+    ...offering,
+    packagesById: Object.fromEntries(
+      Object.entries(offering.packagesById).map(([packageId, rcPackage]) => [
+        packageId,
+        addPlacementContextToPackage(rcPackage, placementId),
```
tonidero, line 78: The placement information is dynamic depending on how the object is obtained... Lmk if you think of better ways of doing this!

src/tests/test-helpers.ts
```diff
@@ -1,10 +1,19 @@
 import { expect } from "vitest";
 import { PurchasesError } from "../entities/errors";
+import { type Offerings } from "../entities/offerings";
 
 export function failTest() {
   expect(true).toBeFalsy();
 }
 
+export function assertExpectedOfferings(
```
tonidero, line 9: Now, we can't do an equals directly in the offerings, since it has a function as well...

src/entities/offerings.ts
```diff
@@ -323,6 +328,12 @@ export interface Offerings {
    * Current offering configured in the RevenueCat dashboard.
    */
   readonly current: Offering | null;
+  /**
+   * Retrieves a specific offering by a placement identifier.
+   * For more info see https://www.revenuecat.com/docs/tools/targeting
+   * @param placementId - The placement identifier to retrieve the offering for.
+   */
+  getCurrentOfferingByPlacementId(placementId: string): Offering | null;
```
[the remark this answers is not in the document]
nicfix: Can we keep it as a simple data interface and populate it where we build instances?
```suggestion
  readonly currentOfferingByPlacementId: Offering | null;
```

src/helpers/offerings-parser.ts
```diff
@@ -0,0 +1,158 @@
...
+      ]),
+    ),
+    availablePackages: offering.availablePackages.map((rcPackage) =>
```
nicfix, line 81: Since we anyway do the mapping/enriching here, maybe we can just enrich the packages once with `addPlacementContextToPackage` and avoid doing it twice here and in the previous field, wdyt?

src/helpers/offerings-parser.ts
```diff
@@ -0,0 +1,158 @@
...
+    Logger.debugLog(
+      "Empty offerings. Please make sure you've configured offerings correctly in the " +
+        "RevenueCat dashboard and that the products are properly configured.",
+    );
+  }
+
+  return {
+    all: allOfferings,
+    current: currentOffering,
+    getCurrentOfferingByPlacementId(placementId: string): Offering | null {
```
nicfix, line 146: Why don't we pre-initialise a `byPlacementId: {[id:string]:Offering}` and pre-calculate it here?

Then anyone can just access it with:
```
offerings.byPlacementId[placementId]
```

src/tests/test-helpers.ts
nicfix: See previous proposal 🙂

src/helpers/offerings-parser.ts
tonidero: Well, it might make sense to do that, we will just increase the size of the offerings object I guess, but it would certainly be cleaner... cc @joshdholtz since you worked on the mobile SDK for these I think, in case you have any thoughts on this approach!

src/entities/offerings.ts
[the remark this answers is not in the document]
joshdholtz: @tonidero One suggestion I would make here is have this name match the name of the other JS SDK. RN uses 👇

`getCurrentOfferingForPlacement`

src/helpers/offerings-parser.ts
joshdholtz: Answered in Slack but there is fallback logic where using any placement identifier should return  fallback offering. That doesn't work well with a dictionary because a placement that doesn't exist would return null 🤷‍♂️

src/entities/offerings.ts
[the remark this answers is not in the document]
tonidero: Actually, that makes more sense... We can move this function to be part of the `Purchases` object, instead of part of the `offerings` interface... I will do that change, which will clean this up

src/main.ts
```diff
@@ -268,6 +227,47 @@ export class Purchases {
...
+      return null;
+    }
+    return findOfferingByPlacementId(
+      placementData,
+      offerings.all,
+      placementIdentifier,
+    );
+  }
+
+  private async getOfferingsForOfferingResponse(
```
tonidero, line 266: This is me trying to avoid duplicating logic.

src/helpers/offerings-parser.ts
```diff
@@ -0,0 +1,147 @@
...
+const addPlacementContextToPackage = (
+  rcPackage: Package,
+  placementId: string,
+): Package => {
+  return {
+    ...rcPackage,
+    rcBillingProduct: {
+      ...rcPackage.rcBillingProduct,
+      presentedOfferingContext: {
+        ...rcPackage.rcBillingProduct.presentedOfferingContext,
```
nicfix, line 27: Can the `rcPackage.rcBillingProduct.presentedOfferingContext` be null/undefined?

src/main.ts
nicfix, line 266:
```suggestion
  private async getAllOfferings(
```

or

```suggestion
  private async fetchOfferings(
```

src/helpers/offerings-parser.ts
tonidero: Good question! And no, since currently, all products can only be obtained through an offering, we at least fill the `presentedOfferingContext` with the presented offering identifier (the other fields may be null if not using targeting/placements. Thanks for double checking!

PR #191, opened 2024-09-23
Support non renewable products.

tonidero:
## Motivation / Description
This PR adds support for non-subscription products.


tonidero:
This is close to done, just need to test it against the backend and add some E2E tests

examples/rcbilling-demo/src/tests/main.test.ts
```diff
@@ -79,6 +118,19 @@ test.describe("Main", () => {
...
+  // Perform purchase
+  const cardButton = card.getByRole("button");
+  await cardButton.click();
+
+  await enterEmailAndContinue(page, userId);
+  await enterCreditCardDetailsAndContinue(page);
+
+  // Confirm success page has shown.
+  const successText = page.getByText("Purchase successful");
+  await expect(successText).toBeVisible({ timeout: 10000 });
```
tonidero, line 131: I was getting some failures with the standard (5000ms) so I increased it to 10000... Maybe we can optimize it in the future when it becomes a problem.

tonidero:
UI changes coming in a followup PR but this is ready for review

examples/rcbilling-demo/src/tests/main.test.ts
antoniobg: Does this only happen for non-subscription purchases?

tonidero: Yeah those were the failures... Not sure if the backend was taking longer for some reason for those 🤔

antoniobg: That's weird, it shouldn't take longer than subscriptions.

tonidero: Right... FWIW, it's taking longer to display the `Purchase successful` text. I'm able to reproduce quite easily locally (subscriptions works almost always and consumables fail quite often).

From running manually (I've done 3 subscriptions and 3 consumable purchases), subscriptions take ~3 poll attempts while consumables ~4-5.

api-report/purchases-js.api.json
```diff
@@ -3154,7 +3214,7 @@
             {
               "kind": "PropertySignature",
               "canonicalReference": "@revenuecat/purchases-js!Product#currentPrice:member",
-              "docComment": "/**\n * Price of the product. In the case of subscriptions, this will match the default option's base phase price.\n */\n",
+              "docComment": "/**\n * Price of the product. This will match the default option's base phase price in subscriptions or price in non subscriptions.\n */\n",
```
antoniobg, line 3217:
```suggestion
              "docComment": "/**\n * Price of the product. This will match the default option's base phase price in subscriptions or the price in non-subscriptions.\n */\n",

```

api-report/purchases-js.api.json
```diff
@@ -3210,7 +3302,7 @@
             {
               "kind": "PropertySignature",
               "canonicalReference": "@revenuecat/purchases-js!Product#defaultSubscriptionOption:member",
-              "docComment": "/**\n * The default subscription option for this product. Null if no subscription options are available like in the case of consumables and non-consumables.\n */\n",
+              "docComment": "/**\n * The default subscription option for this product. Null in the case of no subscriptions.\n */\n",
```
antoniobg, line 3305: I think the description was clearer before? Another question, should we mark it as deprecated or will we do that later?

api-report/purchases-js.api.json
```diff
@@ -3468,6 +3588,86 @@
...
+                "startIndex": 1,
+                "endIndex": 2
+              },
+              "releaseTag": "Public",
+              "name": "NonConsumable"
+            },
+            {
+              "kind": "EnumMember",
+              "canonicalReference": "@revenuecat/purchases-js!ProductType.Subscription:member",
+              "docComment": "/**\n * A product that is a subscription.\n */\n",
```
antoniobg, line 3651: Maybe
```suggestion
              "docComment": "/**\n * A product that is an auto-renewing subscription.\n */\n",

```

examples/rcbilling-demo/src/tests/main.test.ts
antoniobg: I'll take a look, thanks for point it out 👍

antoniobg (approved):

api-report/purchases-js.api.json
tonidero: > I think the description was clearer before?

Hmm yeah I will revert this.

> Another question, should we mark it as deprecated or will we do that later?

My idea was to not deprecate it. This would become a convenience method to avoid the dev having to cast down the `PurchaseOption` themselves if they need to access the `SubscriptionOption` data (like to get all pricing phases for trials/intro offers).

antoniobg: Got it

PR #193, opened 2024-10-01
Update UI to support consumables and non consumables

tonidero:
## Motivation / Description
Most changes to support OTP are done in #191.
 
This adapts the purchase flow UI to better support consumables and non consumables. Changes are on the side bar with the product detail and the purchase success screen
| Product detail | Purchase success |
| --- | --- |
| image | image |

### DIFF
--- a/src/ui/states/state-present-offer.svelte
+++ b/src/ui/states/state-present-offer.svelte
@@ -1,43 +1,58 @@
 <script lang="ts">
     import ModalSection from "../modal-section.svelte";
     import { getRenewsLabel, getTrialsLabel } from "../../helpers/price-labels";
-    import type { Product, PurchaseOption, SubscriptionOption } from "../../entities/offerings";
+    import {
+        type NonSubscriptionOption,
+        type Product,
+        ProductType,
+        type PurchaseOption,
+        type SubscriptionOption,
+    } from "../../entities/offerings";
 
     export let productDetails: Product;
     export let purchaseOption: PurchaseOption;
 
+    const isSubscription = productDetails.productType === ProductType.Subscription;
     const subscriptionOption: SubscriptionOption | null | undefined =
       purchaseOption as SubscriptionOption;
-    const trial = subscriptionOption?.trial;
-    const basePrice = subscriptionOption?.base?.price;
+    const nonSubscriptionOption: NonSubscriptionOption | null | undefined =
+      purchaseOption as NonSubscriptionOption;
+    const subscriptionTrial = subscriptionOption?.trial;
+
+    const subscriptionBasePrice = subscriptionOption?.base?.price;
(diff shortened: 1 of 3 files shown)


src/ui/states/state-present-offer.svelte
```diff
@@ -1,43 +1,58 @@
...
     export let productDetails: Product;
     export let purchaseOption: PurchaseOption;
 
+    const isSubscription = productDetails.productType === ProductType.Subscription;
     const subscriptionOption: SubscriptionOption | null | undefined =
       purchaseOption as SubscriptionOption;
-    const trial = subscriptionOption?.trial;
-    const basePrice = subscriptionOption?.base?.price;
+    const nonSubscriptionOption: NonSubscriptionOption | null | undefined =
+      purchaseOption as NonSubscriptionOption;
```
tonidero, line 19: I would really love to have something like sealed/final classes in typescript but I couldn't find anything :(. Happy to receive feedback if there are better ways to do this! cc @guido732

src/ui/states/state-success.svelte
```diff
@@ -14,5 +18,7 @@
   {onContinue}
 >
   <IconSuccess slot="icon" />
-  Your plan is now active.
+  {#if isSubscription}
+    Your plan is now active.
```
tonidero, line 22: I'm wondering if this should also be shown for non-consumables... If they are used as lifetimes, I think it could make kind of sense, but might be safer to not show it to cover more scenarios... cc @jefago

src/ui/states/state-present-offer.svelte
guido732: hmmmm that'd be useful but there's not an easy way to do so. I did some research and there are workarounds but nothing integrated that could be used like this.

What's the end goal you're trying to reach in this specific case? why would you want to lock the class?

tonidero: Right, so in Kotlin/Swift you can use sealed classes/enum classes to do something like this:
```
sealed class PurchaseOption {
  data class SubscriptionOption(...): PurchaseOption
  data class NonSubscriptionOption(...): PurchaseOption
}
```
Then, later in your code:
```
when (purchaseOption) {
  is SubscriptionOption -> { /* Here the compiler ensures it's a SubscriptionOption */ }
  is NonSubscriptionOption -> { /* Here the compiler ensures it's a NonSubscriptionOption */ }
}
```

It's very useful for polymorphic classes since it ensures you handle the different types enforcedly by the compiler. 

But I believe this doesn't exist in typescript since most things are simple interfaces and we only check for conformance to those interfaces... With the current code I have, if we added a new subtype, the typescript compiler wouldn't fail and we would need to remember to update it. Probably not a huge deal and we can live with it, but just wondering if you had better ideas as for how to organize the code.

antoniobg (approved):

PR #194, opened 2024-10-02
Change endpoint to use the new `/purchase`

tonidero:
## Motivation / Description
We are changing the endpoint name to be more precise now that we will support non subscriptions.


src/tests/helpers/purchase-operation-helper.test.ts
```diff
@@ -22,7 +22,7 @@ describe("PurchaseOperationHelper", () => {
   let purchaseOperationHelper: PurchaseOperationHelper;
 
   const operationSessionId = "test-operation-session-id";
-  const successSubscribeBody: SubscribeResponse = {
+  const successSubscribeBody: PurchaseResponse = {
```
antoniobg, line 25:
```suggestion
  const successPurchaseBody: PurchaseResponse = {
```

antoniobg (approved):

PR #201, opened 2024-10-09
Add support for getting a single offering

antoniobg:
## Motivation / Description

This adds a new API under Purchases.getOffering that allows to obtain an offering by its identifier.

### DIFF
--- a/examples/rcbilling-demo/src/pages/paywall/index.tsx
+++ b/examples/rcbilling-demo/src/pages/paywall/index.tsx
@@ -93,7 +93,7 @@ const PaywallPage: React.FC = () => {
   const { purchases, offering } = usePurchasesLoaderData();
 
   if (!offering) {
-    console.error("No current offering found");
+    console.error("No offering found");
     return <>No offering found!</>;
   }
   const packages: Package[] = offering?.availablePackages || [];
--- a/examples/rcbilling-demo/src/pages/paywall/index.tsx
+++ b/examples/rcbilling-demo/src/pages/paywall/index.tsx
@@ -1,6 +1,6 @@
 import { Offering, Package, PurchasesError } from "@revenuecat/purchases-js";
 import React from "react";
-import { useNavigate, useSearchParams } from "react-router-dom";
+import { useNavigate } from "react-router-dom";
 import { usePurchasesLoaderData } from "../../util/PurchasesLoader";
 import Button from "../../components/Button";
 import LogoutButton from "../../components/LogoutButton";
@@ -90,20 +90,13 @@ export const PackageCard: React.FC<IPackageCardProps> = ({
 
 const PaywallPage: React.FC = () => {
   const navigate = useNavigate();
-  const { purchases, offerings } = usePurchasesLoaderData();
-  const [searchParams] = useSearchParams();
+  const { purchases, offering } = usePurchasesLoaderData();
 
-  const offeringId = searchParams.get("offering_id");
-
-  const selectedOffering = offeringId
-    ? offerings.all[offeringId] || null
-    : offerings.current;
-
-  if (!selectedOffering) {
+  if (!offering) {
     console.error("No current offering found");
     return <>No offering found!</>;
   }
-  const packages: Package[] = selectedOffering?.availablePackages || [];
+  const packages: Package[] = offering?.availablePackages || [];
   if (packages.length == 0) {
     console.error("No packages found in current offering.");
   }
@@ -148,7 +141,7 @@ const PaywallPage: React.FC = () => {
               <PackageCard
                 key={pkg.identifier}
                 pkg={pkg}
-                offering={selectedOffering}
+                offering={offering}
                 onClick={() => onPackageCardClicked(pkg)}
               />
             ) : null,
--- a/examples/rcbilling-demo/src/pages/paywall/index.tsx
+++ b/examples/rcbilling-demo/src/pages/paywall/index.tsx
@@ -1,6 +1,6 @@
 import { Offering, Package, PurchasesError } from "@revenuecat/purchases-js";
 import React from "react";
-import { useNavigate } from "react-router-dom";
+import { useNavigate, useSearchParams } from "react-router-dom";
 import { usePurchasesLoaderData } from "../../util/PurchasesLoader";
 import Button from "../../components/Button";
 import LogoutButton from "../../components/LogoutButton";
@@ -91,13 +91,19 @@ export const PackageCard: React.FC<IPackageCardProps> = ({
 const PaywallPage: React.FC = () => {
   const navigate = useNavigate();
   const { purchases, offerings } = usePurchasesLoaderData();
+  const [searchParams] = useSearchParams();
 
-  const currentOffering = offerings.current;
-  if (!currentOffering) {
+  const offeringId = searchParams.get("offering_id");
+
+  const selectedOffering = offeringId
+    ? offerings.all[offeringId] || null
+    : offerings.current;
+
+  if (!selectedOffering) {
     console.error("No current offering found");
     return <>No offering found!</>;
   }
-  const packages: Package[] = currentOffering?.availablePackages || [];
+  const packages: Package[] = selectedOffering?.availablePackages || [];
   if (packages.length == 0) {
     console.error("No packages found in current offering.");
   }
@@ -142,7 +148,7 @@ const PaywallPage: React.FC = () => {
               <PackageCard
                 key={pkg.identifier}
                 pkg={pkg}
-                offering={currentOffering}
+                offering={selectedOffering}
                 onClick={() => onPackageCardClicked(pkg)}
               />
             ) : null,
--- a/examples/rcbilling-demo/src/tests/main.test.ts
+++ b/examples/rcbilling-demo/src/tests/main.test.ts
@@ -35,13 +35,20 @@ test.describe("Main", () => {
     browserName,
   }) => {
     const userId = getUserId(browserName);
-    const offeringId = "e2e_test_specific_offering"; // The offering ID you expect to filter by
+    const offeringId = "default_download";
     const page = await setupTest(browser, userId, offeringId);
 
     const packageCards = await getAllElementsByLocator(page, CARD_SELECTOR);
 
-    expect(packageCards.length).toBe(1);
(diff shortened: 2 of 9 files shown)


src/main.ts
```diff
@@ -262,6 +262,26 @@ export class Purchases {
...
+   * @param params - The parameters object to customise the offerings fetch. Check {@link GetOfferingsParams}
+   */
+  public async getOffering(
+    offeringIdentifier: string,
+    params?: GetOfferingsParams,
+  ): Promise<Offering | null> {
+    const appUserId = this._appUserId;
+    const offeringsResponse = await this.backend.getOfferings(appUserId);
+
+    const offerings = await this.getAllOfferings(
```
nicfix, line 277: I think that this method won't solve the problem we are trying to address.

If you check inside the `getAllOfferings` method [here](https://github.com/RevenueCat/purchases-js/blob/540b45eb6f89136d473d95439ec7f6e7e4bd8142/src/main.ts#L270-L278) you can see that:
1. it extracts the products from the packages of all offerings
2. calls `/products` with all the product_ids from all offerings

We need to call the `/products` endpoint with the product_ids that are only from the offering we are considering.
So we probably need to build an equivalent of what `getAllOfferings` is doing but with a `offering:Offering` parameter so that it will download only the products from that offering.

antoniobg: Of course, thanks for catching it 👍

src/main.ts
```diff
@@ -301,6 +302,42 @@ export class Purchases {
...
 
+  private async getSingleOffering(
+    offeringsResponse: OfferingsResponse,
+    appUserId: string,
+    offeringIdentifier: string,
+    params?: GetOfferingsParams,
+  ): Promise<Offering | null> {
+    const isCurrentOffering =
+      offeringsResponse.current_offering_id === offeringIdentifier;
+    const offeringResponse = offeringsResponse.offerings.find(
```
nicfix, line 313: Nit for readability:
```suggestion
    const selectedOfferingResponse = offeringsResponse.offerings.find(
```

examples/rcbilling-demo/src/pages/paywall/index.tsx
nicfix, line 1: The changes to this file are not really using the new API.
I don't think we need them.

nicfix: As mentioned in the inline comment, we need to change the `usePurchasesLoaderData` hook to accept the offeringId parameter in order to use the new sdk method.

examples/rcbilling-demo/src/pages/paywall/index.tsx
```diff
@@ -91,13 +91,19 @@ export const PackageCard: React.FC<IPackageCardProps> = ({
 const PaywallPage: React.FC = () => {
   const navigate = useNavigate();
   const { purchases, offerings } = usePurchasesLoaderData();
```
nicfix, line 93: In order to use the new sdk `getOffering` method we need to update the `usePurchasesLoaderData` hook to accept the `offering_id` parameter.

examples/rcbilling-demo/src/tests/main.test.ts
```diff
@@ -31,6 +31,27 @@ test.describe("Main", () => {
     );
   });
 
+  test("Displays specific offering when offering_id is provided", async ({
```
nicfix, line 34: I think this test makes sense only if we actually use the new method (through the `usePurchasesLoaderData` hook) as mentioned above, otherwise is not really testing the new functionality we added.

src/main.ts
```diff
@@ -262,6 +263,26 @@ export class Purchases {
     );
   }
 
+  /**
+   * Retrieves a specific offering by its identifier.
+   * @param offeringIdentifier - The identifier of the offering to retrieve.
+   * @param params - The parameters object to customise the offerings fetch. Check {@link GetOfferingsParams}
+   */
+  public async getOffering(
+    offeringIdentifier: string,
```
tonidero, line 272: I wonder if this could instead be a parameter inside the `GetOfferingParams`... so we keep it to a single method...

examples/rcbilling-demo/src/pages/paywall/index.tsx
antoniobg: Yep, I'm working on it.

src/main.ts
```diff
@@ -262,6 +263,26 @@ export class Purchases {
     );
   }
 
+  /**
+   * Retrieves a specific offering by its identifier.
+   * @param offeringIdentifier - The identifier of the offering to retrieve.
+   * @param params - The parameters object to customise the offerings fetch. Check {@link GetOfferingsParams}
+   */
+  public async getOffering(
```
tonidero, line 271: I do have the concern of people using this and hardcoding the offering id instead of using the current offering... I wonder if we should also have a `getCurrentOffering` to access the current one...

tonidero (commented):
Couple of comments with some concerns... but probably not blockers

src/main.ts
JayShortway: Good point! Maybe `GetOfferingParams` could contain a field that contains either `Current` or an offering identifier? Not sure what the best way to describe that is in TypeScript though, but I believe we do a similar thing for `PurchasesAreCompletedBy`. 

This would prevent yet another new offerings method.

nicfix: Yeah, we can add an optional `offeringId` field to the `GetOfferingParams` [interface](https://github.com/RevenueCat/purchases-js/blob/main/src/entities/get-offerings-params.ts#L5) and use it [here](https://github.com/RevenueCat/purchases-js/blob/main/src/main.ts#L229).

antoniobg: I've given it a shot adding a new parameter to `GetOfferingsParams` named `offeringIdentifier`, which could be a string or a OfferingKeyword (an enum which only value is `Current`). Let me know what you think. I'm not a fan of the `OfferingKeyword` name, but couldn't find anything better.

src/entities/get-offerings-params.ts
```diff
@@ -8,4 +18,10 @@ export interface GetOfferingsParams {
    * If not specified, the default currency will be used.
    */
   readonly currency?: string;
+
+  /**
+   * The identifier of the offering to fetch.
+   * Can be a string identifier or one of the predefined keywords.
+   */
+  readonly offeringIdentifier?: string | OfferingKeyword;
```
antoniobg, line 26: One thing to bear in mind is that `GetOfferingsParams` is also used in [getCurrentOfferingForPlacement](https://github.com/RevenueCat/purchases-js/pull/201/files#diff-4fab5baaca5c14d2de62d8d2fceef376ddddcc8e9509d86cfa5643f51b89ce3dL242), but it's completely ignore in that function. Not sure how do you feel about it.

examples/rcbilling-demo/src/pages/paywall/index.tsx
```diff
@@ -90,14 +90,13 @@ export const PackageCard: React.FC<IPackageCardProps> = ({
 
 const PaywallPage: React.FC = () => {
   const navigate = useNavigate();
-  const { purchases, offerings } = usePurchasesLoaderData();
+  const { purchases, offering } = usePurchasesLoaderData();
 
-  const currentOffering = offerings.current;
-  if (!currentOffering) {
+  if (!offering) {
     console.error("No current offering found");
```
nicfix, line 96: Nit:
```suggestion
    console.error("No offering found");
```

nicfix, line 96: Nit:
```suggestion
    console.error("No offering found");
```

src/entities/get-offerings-params.ts
```diff
@@ -8,4 +18,10 @@ export interface GetOfferingsParams {
    * If not specified, the default currency will be used.
    */
   readonly currency?: string;
+
+  /**
+   * The identifier of the offering to fetch.
+   * Can be a string identifier or one of the predefined keywords.
```
nicfix, line 24:
```suggestion
   * Can be a string identifier or one of the predefined keywords.
   * This parameter is ignored by the getCurrentOfferingForPlacement method.
```

src/entities/get-offerings-params.ts
nicfix: Added a suggestion on how to handle it.

src/main.ts
```diff
@@ -230,6 +236,21 @@ export class Purchases {
     validateCurrency(params?.currency);
     const appUserId = this._appUserId;
     const offeringsResponse = await this.backend.getOfferings(appUserId);
+
+    if (params?.offeringIdentifier) {
```
nicfix, line 240: I guess you can skip this first level if, given that you do the one at line 246.

nicfix (approved):
Just a couple of suggestions, but LGTM.

src/main.ts
```diff
@@ -75,7 +78,10 @@ export type { PurchasesErrorExtra } from "./entities/errors";
 export type { Period, PeriodUnit } from "./helpers/duration-helper";
 export type { HttpConfig } from "./entities/http-config";
 export { LogLevel } from "./entities/log-level";
-export type { GetOfferingsParams } from "./entities/get-offerings-params";
+export type {
+  GetOfferingsParams,
+  OfferingKeyword,
```
tonidero, line 83: Maybe I’m wrong but I think the enum should be exported normally (not just the type)

src/main.ts
```diff
@@ -230,6 +236,19 @@ export class Purchases {
...
+      params?.offeringIdentifier === OfferingKeyword.Current
+        ? offeringsResponse.current_offering_id
+        : params?.offeringIdentifier;
+
+    if (offeringIdFilter) {
+      offeringsResponse.offerings = offeringsResponse.offerings.filter(
+        (offering: OfferingResponse) =>
+          offering.identifier === offeringIdFilter,
+      );
+    }
```
tonidero, line 250: One consequence of this is that if the given offering id is not the current, the response of the getOfferings will have a null `current` offering. This is probably fine, but maybe we should add it to the docs for the `current` field

tonidero (approved):
Just a comment about some docs, but I think it looks good!

src/main.ts
antoniobg: Done

src/main.ts
antoniobg: Good catch. Done.

PR #206, opened 2024-10-10
ECO-1336: Revamping storybook to make it easier to implement UI changes in the sdk

nicfix:
## Motivation / Description
We need to implement some appearance changes due to customisation.
This PR sets up again storybook (updated) to create directly stories in svelte and test them directly instead of doing back and forth with the demo app.

## Changes introduced
* Updated storybook
* Fixed the main.ts setup
* Started working on a story for the components affected by customisation


guido732 (commented):
looks good, although when I install dependencies and try to run the storybook command I get this when trying to view a story 
image

guido732:
also, we should add `storybook-static` to the `.gitignore` file

nicfix:
The one you picked is the old story we had and it doesn't work anymore, I think we should not use storybook that way, I'll remove it (we still have it in git if we want in the future).

nicfix:
Done both!

guido732 (approved):

PR #209, opened 2024-10-13
Customization: Added shapes, background colors and generated text color.

nicfix:
## Motivation / Description
This PR implements the tickets ECO-1336, ECO-1526 and ECO-1535.

ECO-1336: Support for rounded, pill and rectangle shapes
ECO-1526: Support for custom background colors and automatically generated text color (black on white and vice versa)
ECO-1535: Additional visual improvements:
* Updated the base text colors
* Added the accent color to the lock, success, and error icons
* Removed the close button and added a X button on the top right corner

## Changes introduced
This is kind of a big PR, I recorded this loom to walk through it and explain what I changed and how.

https://www.loom.com/share/07defb2a69ec4503a851a208d7809953?sid=56f31f23-a3bf-480d-8f5d-9719a258de27

In general:
* Added several stories to test the customization on all the sdk components
* Refactored the code to generate the customization. It was designed to generate one customization, however we needed to customize differently the 2 panels (different backgrounds, different text colors). Now it can be used in every component to customize the "sub-tree"
* Removed the Shell component and moved the Modal inside the Aside and Main components, this allowed embedding the customization inside them
* Changed the re-coloring of the SVG icons using the fill="currentColor" property, in this way we can do it just with CSS and we don't need to store the branding globally
* Removed the global brandingAppearance store, it was causing issues when multiple stories were loaded since it was remaining dirty when changing components. We don't need to store it anymore thanks to the Theme class.
* Moved all the customization code inside /src/ui so that it doesn't mix with the sdk code.


src/helpers/process-style-overrides.ts
nicfix, line 1: The content of this file has been moved to src/ui/theme/utils.ts or refactored.

src/store/store.ts
nicfix, line 1: We don't need anymore to store the appearance globally.

The theme class can calculate id dynamically and we don't have anymore the button or the icons that need to change style in JS. All customizations are implemented using CSS vars.


This was also creating issues leaving the customization stale when having multiple pages in storybook.

src/ui/assets/icon-cart.svelte
nicfix, line 1: Moved to src/ui/icons/icon-cart.svelte

src/ui/assets/icon-error.svelte
nicfix, line 1: Moved to src/ui/icons/icon-cart.svelte

src/ui/assets/icon-lock.svelte
nicfix, line 1: Moved to src/ui/icons/icon-lock.svelte

src/ui/assets/icon-success.svelte
nicfix, line 1: Moved to src/ui/icons/icon-success.svelte

src/ui/icons/icon-cart.svelte
```diff
@@ -0,0 +1,7 @@
+<script>
+  import Icon from "../../assets/cart.svg?raw"; // Load SVG as a raw string
+</script>
+
+<div style="color:var(--rc-color-gray-text-dark);">
```
nicfix, line 5: Here I modified the svg file to use `fill="currentColor"`.

In this way the css style can change the color of the resulting icon.
That's how fontawesome svg icons work as well.

src/ui/icons/icon-error.svelte
```diff
@@ -0,0 +1,15 @@
+<script>
+  import Icon from "../../assets/error.svg?raw"; // Load SVG as a raw string
+</script>
+
+<div style="color:var(--rc-color-error);" class="rcb-ui-asset-icon">
```
nicfix, line 5: Here I modified the svg file to use `fill="currentColor"`.

In this way the css style can change the color of the resulting icon.
That's how fontawesome svg icons work as well.

src/ui/icons/icon-lock.svelte
```diff
@@ -0,0 +1,14 @@
+<script>
+  import Icon from "../../assets/lock.svg?raw"; // Load SVG as a raw string
+</script>
+
+<div style="color:var(--rc-color-accent);" class="rcb-ui-asset-icon">
```
nicfix, line 5: Here I modified the svg file to use `fill="currentColor"`.

In this way the css style can change the color of the resulting icon.
That's how fontawesome svg icons work as well.

src/ui/icons/icon-success.svelte
```diff
@@ -0,0 +1,15 @@
+<script>
+  import Icon from "../../assets/success.svg?raw"; // Load SVG as a raw string
+</script>
+
+<div style="color:var(--rc-color-accent);" class="rcb-ui-asset-icon">
```
nicfix, line 5: Here I modified the svg file to use `fill="currentColor"`.

In this way the css style can change the color of the resulting icon.
That's how fontawesome svg icons work as well.

src/ui/layout/aside-block.svelte
```diff
@@ -1,9 +1,17 @@
-<script>
-  export let style = "";
+<script lang="ts">
+  import { type BrandingAppearance } from "../../networking/responses/branding-response";
+  import Modal from "../modal.svelte";
+  import { Theme } from "../theme/theme";
+
+  export let brandingAppearance: BrandingAppearance | undefined = undefined;
+  let style = new Theme(brandingAppearance).productInfoStyleVars;
```
nicfix, line 7: Style for the product info panel calculated from the brandingAppearance

src/ui/layout/main-block.svelte
```diff
@@ -1,9 +1,18 @@
-<script>
-  export let style = "";
+<script lang="ts">
+  import { toFormStyleVar } from "../theme/utils.ts";
+  import { type BrandingAppearance } from "../../networking/responses/branding-response";
+  import Modal from "../modal.svelte";
+  import { Theme } from "../theme/theme";
+
+  export let brandingAppearance: BrandingAppearance | undefined = undefined;
+  let style = new Theme(brandingAppearance).formStyleVars;
```
nicfix, line 8: Style for the form panel calculated from the brandingAppearance

src/ui/layout/main-block.svelte
nicfix, line 1: I plan to rename this file/component into `form-panel.svelte`, I'll do it in a follow up PR to introduce less confusion

src/ui/layout/aside-block.svelte
nicfix, line 1: I plan to rename this file/component into `product-info-panel.svelte`, I'll do it in a follow up PR to introduce less confusion

src/ui/modal.svelte
```diff
@@ -1,47 +1,40 @@
 <script lang="ts">
-  export let dark = false;
```
nicfix, line 2: There's no more a dark and light modal, each one of them can be customised with custom colors.

src/ui/modal.svelte
```diff
@@ -1,47 +1,40 @@
...
+        color: var(--rc-color-grey-text-dark);
+        padding: 40px;
+        flex-direction: column;
+        display: flex;
+    }
 
-  .rcb-modal-dark {
-    background-color: var(--rc-color-grey-text-dark);
-    color: var(--rc-color-white);
-    min-height: 0px;
```
nicfix, line 30: This rule was moved to the panel that uses the modal (Aside)

src/ui/rcb-ui.svelte
```diff
@@ -194,75 +193,72 @@
...
-              {:else}
-                <IconCart />
-              {/if}
-            </ModalHeader>
-            {#if productDetails && purchaseOptionToUse}
-              <StatePresentOffer
-                {productDetails}
-                purchaseOption={purchaseOptionToUse}
-              />
+        <Aside brandingAppearance={brandingInfo?.appearance}>
```
nicfix, line 196: Passing down the brandingAppearance to allow customization

src/ui/rcb-ui.svelte
```diff
@@ -194,75 +193,72 @@
...
-          {#if state === "loading"}
-            <StateLoading />
-          {/if}
-          {#if state === "error"}
-            <StateError
-              {brandingInfo}
-              lastError={lastError ??
+        </Aside>
+      {/if}
+      <Main brandingAppearance={brandingInfo?.appearance}>
```
nicfix, line 213: Passing down the brandingAppearance to allow customization

src/ui/sandbox-banner.svelte
```diff
@@ -3,7 +3,7 @@
 <style>
   div {
     background-color: var(--rc-color-warning);
-    color: var(--rc-color-grey-text-dark);
+    color: rgba(0,0,0,1);
```
nicfix, line 6: Making sure the color is not affected by the customization

src/ui/states/state-error.svelte
```diff
@@ -38,12 +38,11 @@
   {brandingInfo}
   {onContinue}
   type="error"
+  closeButtonTitle="Try Again"
```
nicfix, line 41: Updated the action label as per figma design

src/ui/states/state-needs-auth-info.svelte
```diff
@@ -33,7 +34,10 @@
 </script>
 
 <form on:submit|preventDefault={handleContinue}>
-  <ModalHeader>Billing email address</ModalHeader>
+  <ModalHeader>
+    <span>Billing email address</span>
+    <CloseButton on:click={onClose} />
```
nicfix, line 39: Added the X to close the panel (ECO-1535)

src/ui/states/state-needs-auth-info.svelte
```diff
@@ -57,61 +64,61 @@
           Continue
         {/if}
       </Button>
-      <Button intent="secondary" on:click={onClose} disabled={processing}
```
nicfix, line 60: Removed the close button (ECO-1535)

src/ui/states/state-needs-payment-info.svelte
```diff
@@ -75,6 +76,11 @@
       onContinue();
     }
   };
+
+  const theme = new Theme(brandingInfo?.appearance);
+
+  let shapeCustomisation = theme.shape;
```
nicfix, line 82: Stripe elements to not support css variables, using directly the css rules calculated by the Theme.

src/ui/states/state-needs-payment-info.svelte
```diff
@@ -134,22 +151,24 @@
...
 <style>
-  .rcb-stripe-elements-container {
-    width: 100%;
-    margin-top: 32px;
-    margin-bottom: 24px;
-  }
+    .rcb-stripe-elements-container {
+        width: 100%;
+
+        /* The standard height of the payment form from Stripe */
```
nicfix, line 167: The card was resizing while loading, this fixed min-height prevents that

src/ui/states/state-success.svelte
```diff
@@ -16,6 +16,7 @@
   title="Purchase successful"
   {brandingInfo}
   {onContinue}
+  closeButtonTitle="Close"
```
nicfix, line 19: Updated the action label as per figma design

src/ui/theme/colors.ts
nicfix, line 1: This file includes the default colors definition for the form and the product info components.
It also includes mappings for the colors <=> branding appearance properties.

The text colors are calculated and not mapped based on the background color they will be used on.

src/ui/theme/shapes.ts
nicfix, line 1: Shapes and css rules definitions.

src/ui/theme/theme.ts
nicfix, line 1: The theme class, a shortcut class used in the components.

The logic is implemented in stateless functions included in the utils file. This is cleaner to use in the svelte files.

src/ui/theme/utils.ts
```diff
@@ -0,0 +1,190 @@
...
+    "grey-text-light": defaultColors["grey-text-light"],
+    "grey-ui-dark": defaultColors["grey-ui-dark"],
+    "grey-ui-light": defaultColors["grey-ui-light"],
+  };
+
+  if (!backgroundColors) {
+    return fallbackColors;
+  }
+
+  if (backgroundColors.startsWith("#")) {
```
nicfix, line 60: We should receive always hex colors from the branding appearance, but just to be sure this function always falls back to the fallbackColors

src/ui/theme/utils.ts
```diff
@@ -0,0 +1,190 @@
...
+
+  return Object.fromEntries(mappedColors) as Colors;
+};
+
+export const toColors = (
+  colorsMapping: any,
+  defaultColors: Colors,
+  brandingAppearance?: BrandingAppearance | undefined,
+): Colors => {
+  const mappedColors = mapColors(
```
nicfix, line 98: Here is where colors are mapped.

src/ui/theme/utils.ts
```diff
@@ -0,0 +1,190 @@
...
+    case "rectangle":
+      return RectangularShape;
+    case "pill":
+      return PillsShape;
+    default:
+      return DefaultShape;
+  }
+};
+
+export const toStyleVar = (prefix: string = "", entries: [string, string][]) =>
```
nicfix, line 151: Example outputs are:

```
--rc-shape-input-border-radius: 16px

--rc-color-accent: #576CDB
```

src/ui/transaction-overview.svelte
nicfix, line 1: This file was empty and not used

src/assets/colors.ts
nicfix, line 1: Moved to src/ui/theme/colors.ts

src/assets/cart.svg
nicfix, line 1: Now uses the `fill=currentColor` allowing css customization.

src/assets/error.svg
nicfix, line 1: Now uses the `fill=currentColor` allowing css customization.

src/assets/lock.svg
nicfix, line 1: Now uses the `fill=currentColor` allowing css customization.

src/assets/success.svg
nicfix, line 1: Now uses the `fill=currentColor` allowing css customization.

src/ui/icons/icon-cart.svelte
jefago: yeah, this is a good solution, I saw the SVG rewriting before and didn't like that very much either

src/ui/theme/utils.ts
```diff
@@ -0,0 +1,190 @@
...
+    ];
+  return null;
+};
+
+const rgbToTextColors = ([r, g, b]: [number, number, number]) => {
+  const sum = r + g + b;
+  const baseColor = sum <= 3 * 128 ? "255,255,255" : "0,0,0";
+
+  return {
+    "grey-text-dark": `rgba(${baseColor},0.8)`,
```
jefago: this should have 100% opacity, I pushed a fix

src/ui/theme/utils.ts
```diff
@@ -0,0 +1,190 @@
...
+      parseInt(color[1], 16),
+      parseInt(color[2], 16),
+      parseInt(color[3], 16),
+    ];
+  return null;
+};
+
+const rgbToTextColors = ([r, g, b]: [number, number, number]) => {
+  const sum = r + g + b;
+  const baseColor = sum <= 3 * 128 ? "255,255,255" : "0,0,0";
```
jefago, line 35: I will push the improved luminosity calculation

src/ui/states/state-needs-auth-info.svelte
```diff
@@ -32,86 +33,100 @@
...
+    <CloseButton on:click={onClose} />
+  </ModalHeader>
+  <form on:submit|preventDefault={handleContinue}>
+    <ModalSection>
+      <div class="form-container">
+        <div class="form-label"><label for="email">Email</label></div>
+        <div class="form-input {inputClass}">
+          <input
+            name="email"
+            placeholder="john@appleseed.com"
```
tonidero, line 48: NABD and it was already like this but should we change this to something else like `Email`?

src/ui/theme/utils.ts
```diff
@@ -0,0 +1,224 @@
...
+};
+
+const hexToRGB = (color: string): RGB | null => {
+  if (color.length == 7)
+    return {
+      r: parseInt(color.slice(1, 3), 16),
+      g: parseInt(color.slice(3, 5), 16),
+      b: parseInt(color.slice(5, 7), 16),
+    };
+  if (color.length == 4)
```
tonidero, line 30: Hmm should we always require a `#RRGGBB` format? Not sure if we need to support this as well...

src/ui/theme/shapes.ts
```diff
@@ -0,0 +1,25 @@
+export interface Shape {
+  "input-border-radius": string;
+  "input-button-border-radius": string;
+  "modal-border-radius": string;
```
tonidero, line 4: Hmm is it possible to name all these using camelCase instead? So we avoid using strings directly... Feels better to me to avoid strings, even if the compiler would still be able to catch most issues.

Edit, After more reviewing, I understand this is to make them compatible with CSS... I wonder if we should have a conversion method from the camelCase interface to a css-compatible map later on...

tonidero: I'm wondering if these should be numbers instead of strings. Just so it's easier to customize from the dashboard in the future if needed.

src/ui/theme/theme.ts
```diff
@@ -0,0 +1,31 @@
...
+import { Shape } from "./shapes";
+import { Colors } from "./colors";
+
+export class Theme {
+  constructor(
+    private readonly brandingAppearance?: BrandingAppearance | undefined,
+  ) {}
+
+  get shape(): Shape {
+    return toShape(this.brandingAppearance);
```
tonidero, line 17: Hmm this means these need to be recalculated every time they are called correct? Would it be possible to precalculate them on the constructor? That way we can just make `Theme` an interface.

src/ui/theme/utils.ts
```diff
@@ -0,0 +1,224 @@
...
+    return DefaultShape;
+  }
+  switch (brandingAppearance.shapes) {
+    case "rounded":
+      return RoundedShape;
+    case "rectangle":
+      return RectangularShape;
+    case "pill":
+      return PillsShape;
+    default:
```
tonidero, line 180: FFTI. Hmm I'm wondering if we could change the type of shapes to an `enum` so we can have exhaustive switch-cases... NABD right now though.

src/store/store.ts
```diff
@@ -1,18 +0,0 @@
...
-  color_form_bg: "",
-  color_page_bg: "",
-  font: "",
-  shapes: "default",
-  show_product_description: false,
-};
-
-// Store to hold appearance configuration object for custom overrides
-export const appearanceConfigStore =
-  writable<BrandingAppearance>(DEFAULT_STORE_VALUES);
```
tonidero, line 18: I think the intention behind using a store was to avoid passing the appearence around the widgets, but it's true that's basically changing from explicit dependencies to implicit, so I think this makes sense 👍

tonidero (approved):
Just some comments and questions but looking great. Thanks for doing this!

src/ui/theme/shapes.ts
nicfix: On the camelCase:
I personally don't like when I have to remember conventions where "some-thing" is actually changed by setting "someThing". These are style variables and css prefers `-` to camelCase.
I would keep it simple, the mapping is already complex as is 😅 

On the numbers:
I would avoid it since the sizes can actually be expressed in different values than `px`, as example `em` or `%`.
This keeps that here and makes it look a bit more like CSS

src/ui/theme/theme.ts
nicfix: Yeah, not a bit calculation for now, but we can definitely calculate it when initializing.

In the end we are using a property, so changing from calculation to pre-calculated field will be transparent to the client. They won't need to change the usage `theme.shape`.

I don't feel it that urgent (calculations are really limited here), but I can change it if needed.

src/ui/theme/utils.ts
nicfix: It's quite typical to shorthand a color with `#fff` I'd keep it if possible.

src/ui/theme/utils.ts
tonidero: Yup makes sense 👍

src/ui/theme/theme.ts
tonidero: No, I think it's not urgent, just an optimization IMO but can be addressed later

src/store/store.ts
nicfix: In the end we need to pass it only to:
1. The parent HTML element (Aside or Main) and the variables will be applied downwards
2. The purchase form since we need it to customize the stripe elements

There's not a lot of drill-down

PR #210, opened 2024-10-14
Fix locale to 'en'

jefago:
## Motivation / Description

Fix locale of the Stripe form to 'en'

Temporary fix until we support localization of the purchase flow


tonidero (approved):
Maybe we should remove the slack link from the description and mention this is temporary until we support localization, but changes look good!

PR #211, opened 2024-10-15
Fix security vulnerabilities

tonidero:
## Motivation / Description
This addresses some security vulnerabilities reported by dependabot:
- Removes outdated `rollup` dependency. Seems it wasn't used by anything and we already have a newer version of rollup as a dependency of `vite`.
- Update `tough-cookie`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@revenuecat/purchases-js",
-  "version": "0.9.0",
+  "version": "0.10.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@revenuecat/purchases-js",
-      "version": "0.9.0",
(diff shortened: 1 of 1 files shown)


package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "@revenuecat/purchases-js",
-  "version": "0.9.0",
+  "version": "0.10.0",
```
JayShortway, line 3: Did the release train miss this, and the same on line 9? (Not suggesting any changes for this PR.)

JayShortway (approved):
Thanks for taking care of this!

package-lock.json
tonidero: The current automations are not updating the lock files here and for the sample apps. Need to fix those to also update these on each release... It shouldn't affect anything though I believe.

PR #217, opened 2024-10-21
Fix handling unrecognized error codes in purchase operation

tonidero:
## Motivation / Description
We were not handling unrecognized error codes in the polling operation session endpoint correctly... It was basically getting stuck without returning a response to the developer. This fixes that so we handle unknown error codes by returning an unknown error.

### DIFF
--- a/src/helpers/purchase-operation-helper.ts
+++ b/src/helpers/purchase-operation-helper.ts
@@ -251,6 +251,30 @@ export class PurchaseOperationHelper {
           ),
         );
         return;
+      case CheckoutStatusErrorCodes.SetupIntentCompletionFailed:
+        reject(
+          new PurchaseFlowError(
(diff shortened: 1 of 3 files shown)


src/helpers/purchase-operation-helper.ts
```diff
@@ -251,6 +251,30 @@ export class PurchaseOperationHelper {
           ),
         );
         return;
+      case CheckoutStatusErrorCodes.SetupIntentCompletionFailed:
```
tonidero, line 254: I would appreciate a double check on whether we need any additional handling on the SDK for this error code

vicfergar: I think it’s fine. Since no action can be taken by the end customer, this seems correct.

vicfergar (approved):

PR #220, opened 2024-10-23
[ECO-1560] Add redemption info data on status response

rogersole:
## Motivation / Description

Added the parsing of the `redemption_info` returned by the backend on `success` purchase operation done by an anonymous subscriber.

Also used the `redeem_url` query parameter to show the `REDEEM` button link correctly on the demo page.


src/entities/redemption-info.ts
```diff
@@ -0,0 +1,23 @@
+import type { CheckoutStatusResponse } from "../networking/responses/checkout-status-response";
+
+/**
+ * This object gives you access to all purchase redemption data.
```
vicfergar, line 4: What do you mean by **all** purchase redemption data?
```suggestion
 * This object gives you access to the purchase redemption data.
```

vicfergar (commented):
I’m not an expert on the SDK, but from what I can see, it LGTM. However, I wouldn’t consider this review fully authoritative for approval, so it might be worth having someone more familiar with the SDK take a final look.

examples/rcbilling-demo/src/pages/success/index.tsx
```diff
@@ -3,19 +3,12 @@ import AppStoreButton from "../../components/AppStoreButton";
...
-
-const generateDeepLink = () => {
-  const appUserId = Purchases.getSharedInstance().getAppUserId();
-  return `revenuecatbilling://redeem_rcb_purchase?redemption_token=${encodeURIComponent(appUserId)}`;
-};
 
 const SuccessPage: React.FC = () => {
+  const queryString = window.location.search;
+  const urlParams = new URLSearchParams(queryString);
+  const redeemUrlParam = urlParams.get("redeem_url");
```
tonidero, line 10: there might be a better way, but since this is only for the demo app, I think it's ok to do it this way 👍

src/entities/redemption-info.ts
```diff
@@ -0,0 +1,23 @@
+import type { CheckoutStatusResponse } from "../networking/responses/checkout-status-response";
+
+/**
+ * This object gives you access to the purchase redemption data.
```
tonidero, line 4: Maybe a bit more detail like:
```suggestion
 * This object gives you access to the purchase redemption data when
 * the purchase can be redeemed to a mobile user, like in the case of anonymous users.
```

tonidero (approved):
Just some comments but nothing blocking. Looks great!

Note that this adds some new APIs, so I thought about whether to keep it on a branch for now... But I think it should be ok to merge for now but not deploy a new SDK version for now. cc @nicfix

nicfix:
@tonidero @rogersole I think the best for now is to ship the change but return the data from the backend only for apps in the feature flag.

Given that @rogersole is already working also on making the redeem url optional I think that the fastest is to return for now only behind feature flag and then return only when the app allows it.

The sdk should be ready for both situations

rogersole:
> @tonidero @rogersole I think the best for now is to ship the change but return the data from the backend only for apps in the feature flag.
> 
> Given that @rogersole is already working also on making the redeem url optional I think that the fastest is to return for now only behind feature flag and then return only when the app allows it.
> 
> The sdk should be ready for both situations

Created [this PR](https://github.com/RevenueCat/khepri/pull/10619) to return the `redemption_info` from backend behind the `RCBILLING_ANONYMOUS_PURCHASES` feature flag.

api-report/purchases-js.api.json
```diff
@@ -2272,7 +2272,7 @@
         {
           "kind": "Enum",
           "canonicalReference": "@revenuecat/purchases-js!OfferingKeyword:enum",
-          "docComment": "/**\n * Keywords for identifying specific offerings in the {@link Purchases.getOfferings} method.\n */\n",
+          "docComment": "/**\n * Keywords for identifying specific offerings in the {@link Purchases.getOfferings} method.\n *\n * @public\n */\n",
```
nicfix, line 2275: Was this addition to the doc intentional?

examples/rcbilling-demo/src/pages/paywall/index.tsx
```diff
@@ -111,14 +111,24 @@ const PaywallPage: React.FC = () => {
...
     // How do we complete the purchase?
     try {
-      const { customerInfo } = await purchases.purchase({
+      const { customerInfo, redemptionInfo } = await purchases.purchase({
         rcPackage: pkg,
         purchaseOption: option,
       });
 
       console.log(`CustomerInfo after purchase: ${customerInfo}`);
+      console.log(
```
nicfix, line 120: Do we need this console.log?

src/main.ts
```diff
@@ -345,9 +350,10 @@ export class Purchases {
    * @throws {@link PurchasesError} if there is an error while performing the purchase. If the {@link PurchasesError.errorCode} is {@link ErrorCode.UserCancelledError}, the user cancelled the purchase.
    */
   @requiresLoadedResources
-  public purchase(
-    params: PurchaseParams,
-  ): Promise<{ customerInfo: CustomerInfo }> {
+  public purchase(params: PurchaseParams): Promise<{
+    customerInfo: CustomerInfo;
+    redemptionInfo: RedemptionInfo | null;
+  }> {
```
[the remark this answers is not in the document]
nicfix: I agree with @tonidero's proposal, it would make things clearer from a documentation standpoint.

src/tests/helpers/purchase-operation-helper.test.ts
```diff
@@ -165,7 +165,7 @@ describe("PurchaseOperationHelper", () => {
     const getCheckoutStatusResponse: CheckoutStatusResponse = {
       operation: {
         status: CheckoutSessionStatus.Succeeded,
-        isExpired: false,
+        is_expired: false,
```
nicfix, line 168: Was this field ever exposed outside of the sdk?
I am guessing no.

nicfix (approved):
LG,

a couple of suggestions and questions but none of them is blocking

examples/rcbilling-demo/src/pages/paywall/index.tsx
rogersole: yup, or lint will fail for property not being used. We can remove it once we use it somewhere/somehow

src/tests/helpers/purchase-operation-helper.test.ts
rogersole: it's not used but we fixed with Toni anyway, since it was not parsed from the json response

api-report/purchases-js.api.json
rogersole: yes. It was a check failure running tests and we updated it with the `@public` tag

src/main.ts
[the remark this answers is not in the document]
rogersole: added

nicfix:
We marked this pr as breaking since the new fields returned by the purchase method would break apps that explicitly assigned them to a typed variable.

Check this conversation for more context:
https://revenuecat.slack.com/archives/C03UKCQLW1Z/p1729847077567529?thread_ts=1729843908.089619&cid=C03UKCQLW1Z

tonidero:
Ohh ok… the automations won’t make the release automatically then… but I can take care of that next week

PR #226, opened 2024-10-29
ECO-1595: Bumping to svelte 5 for the upcoming purchases-ui-web

nicfix:
## Motivation / Description
This PR bumps Svelte to version 5 in order to support the upcoming `purchases-ui-web` library.
It required quite a lot of troubleshooting but everything seems to be working fine now.
I needed to migrate the purchases-js project to yarn to be able to handle the dependencies for the `svelte-stripe` package.

`svelte-stripe` is the weakest of our dependencies, there are a couple of PRs to bump dependencies in their repo but still WIP since a while. We might want to build our own or fork to be less dependent on that project.

## Changes introduced
* Bumped several dependencies
* Added the dependency to `@revenuecat/purchases-web-ui` (will be renamed shortly)
* Migrated circleci to yarn to solve dependencies


api-report/purchases-js.api.md
```diff
@@ -312,13 +312,13 @@ export class PurchasesError extends Error {
     constructor(
     errorCode: ErrorCode,
     message?: string,
-    underlyingErrorMessage?: string | null | undefined,
+    underlyingErrorMessage?: (string | null) | undefined,
```
guido732, line 315: why are we doing this?

vitest.config.js
```diff
@@ -39,6 +39,10 @@ export default defineConfig({
     exclude: [...configDefaults.exclude, "examples/**"],
   },
 
+  resolve: {
+    conditions: mode === "test" ? ["browser"] : [],
+  },
+
```
guido732, line 45: can we add a comment explaining the reason for this?

api-report/purchases-js.api.md
nicfix: This is the result of running prettier after installing it with yarn.

Not sure what changes, I'll double check how to fix it.

examples/rcbilling-demo/.gitignore
```diff
@@ -29,3 +29,5 @@ src/.env
 /playwright-report/
 /blob-report/
 /playwright/.cache/
+
+yarn.lock
```
nicfix, line 33: We want to ignore the yarn.lock file only in the demo app.
The sdk is going to use yarn until we can resolve some dependencies issues with npm.

api-report/purchases-js.api.md
nicfix: Ok mystery solved, this was caused by the update of api-extractor.
I'll try to check if there's a way to configure this behaviour.

It's generated looking at the `purchase-operation-helper` `PurchaseFlowError` class.

Which has the following constructor
```
constructor(
    public readonly errorCode: PurchaseFlowErrorCode,
    message?: string,
    public readonly underlyingErrorMessage?: string | null,
    public readonly purchasesErrorCode?: ErrorCode,
    public readonly extra?: PurchasesErrorExtra,
  ) {
    super(message);
  }
```

The `(string|null)` is defined by the type and the `undefined` is added by the ?

.storybook/main.ts
```diff
@@ -6,7 +6,12 @@ const config: StorybookConfig = {
     "@storybook/addon-links",
     "@storybook/addon-essentials",
     "@storybook/addon-interactions",
-    "@storybook/addon-svelte-csf",
+    {
```
nicfix, line 9: Needed to bump storybook without having to rewrite all our stories

api-report/purchases-js.api.md
nicfix: I didn't find a way to configure the api extractor.
We should change the api to `underlyingErrorMessage?: string` so that the type won't be `(string | null)`.

Since these files are generated and we have already a lot of changes I wouldn't do it as part of this email.

vitest.config.js
nicfix: Done!

guido732 (approved):

PR #227, opened 2024-11-02
ECO-1606: @internal @experimental renderPaywall method

nicfix:
## Motivation
This PR adds the support for rendering paywalls through RCBilling.

## Changes

* Added the entities needed to load the paywall config
* Added the renderPaywall method
* Added a route in the demo app to render the paywall for an offering

## Note for the reviewer
This is an initial implementation marked as @experimental and @internal.
This will cause the developer's build to fail unless they explicitly mark the usage // @ts-expect-error.

Also if the feature flag is not enabled and no paywall is defined the `renderPaywall` method will fail for the offering passed as parameter returning an exception that explains that paywalls are coming soon.

This allows us to start shipping the initial implementation and refine as we go and at the same time puts in place enough safeguards to prevent the developers from using the method.


src/main.ts
```diff
@@ -230,6 +232,126 @@ export class Purchases {
...
+
+    let resolvedHTMLTarget =
+      htmlTarget ?? document.getElementById("rcb-ui-pw-root");
+
+    if (resolvedHTMLTarget === null) {
+      const element = document.createElement("div");
+      element.id = "rcb-ui-pw-root";
+      element.className = "rcb-ui-pw-root";
+      // one point less than the purchase flow modal.
+      element.style.zIndex = "1000000";
```
antoniobg, line 258: Should we add constants for these?

nicfix: On it

antoniobg (approved):

src/entities/render-paywall-params.ts
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * Parameters for the {@link Purchases.renderPaywall} method.
+ * @public
+ */
+export interface RenderPaywallParams {
+  /**
+   * The identifier of the offering to fetch the paywall for.
+   * Can be a string identifier or one of the predefined keywords.
+   */
+  readonly offering: Offering;
```
tonidero, line 12: We could make this optional as well (so the whole params could be optional) and just use the `current` offering by default. Can be done in future PRs though 👍

src/entities/render-paywall-params.ts
```diff
@@ -0,0 +1,27 @@
+import { Offering } from "./offerings";
+
+/**
+ * Parameters for the {@link Purchases.renderPaywall} method.
+ * @public
+ */
+export interface RenderPaywallParams {
+  /**
+   * The identifier of the offering to fetch the paywall for.
+   * Can be a string identifier or one of the predefined keywords.
```
tonidero, line 10: This seems outdated?

src/main.ts
```diff
@@ -48,6 +48,9 @@ import {
 import { type RedemptionInfo } from "./entities/redemption-info";
 import { type PurchaseResult } from "./entities/purchase-result";
 import { mount } from "svelte";
+import { RenderPaywallParams } from "./entities/render-paywall-params";
```
tonidero, line 51: We should also export this from here, so it's available to consumers and generated as part of the API.

src/entities/offerings.ts
```diff
@@ -353,6 +354,8 @@ export interface Offering {
    * Weekly package type configured in the RevenueCat dashboard, if available.
    */
   readonly weekly: Package | null;
+
+  readonly paywall_components: PaywallData | null;
```
tonidero, line 358: Hmm I think it would be good to have this be camelCase `paywallComponents`. We should also mark this as experimental so devs don't use it accidentally.

src/main.ts
```diff
@@ -230,6 +233,126 @@ export class Purchases {
...
+    const selectedLocale = paywallParams.selectedLocale || navigator.language;
+
+    return new Promise((resolve, reject) => {
+      mount(Paywall, {
+        target: certainHTMLTarget,
+        props: {
+          paywallData: offering.paywall_components!,
+          selectedLocale: selectedLocale,
+          onNavigateToClicked: navigateToUrl,
+          onVisitCustomerCenterClicked: onVisitCustomerCenterClicked,
```
tonidero, line 330: This seems currently unused correct?

src/entities/render-paywall-params.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  readonly htmlTarget?: HTMLElement;
+
+  readonly purchaseHtmlTarget?: HTMLElement;
+
+  readonly customerEmail?: string;
+
+  readonly onNavigateToUrl?: (url: string) => void;
+
+  readonly onBack?: () => void;
```
tonidero, line 22: Hmm these are new, I wonder if these should also be params for the `PurchaseParams` object?

tonidero (commented):
Left some comments that can be addressed in followup PRs.

PR #228, opened 2024-11-05
fix: Adjust the default luminance threshold from 0.2 to 0.37

vicfergar:
## Motivation / Description

I've increased the luminance threshold to **0.37** with the following results:
image


tonidero (approved):
Changes look good to me! I've modified the label to something I think fits the change better. Lmk if you think otherwise!

PR #232, opened 2024-11-12
Autogenerate anonymous app user ID

antoniobg:
## Motivation / Description
Introduce new config params to autogenerate anonymous App User IDs. 

Using overloads, developers could keep calling `configure` as they did
```tyspescript
Purchases.configure("myApiKey", "john-doe")
```

or explicitly asking the SDK to autogenerate anonymous IDs for them
```typescript
Purchases.configure("myApiKey", { appUserIDsAreProvidedBy: AppUserIDProvider.RevenueCat })
```

## Changes introduced
- Add new enum `AppUserIDProvider`
- Allow to initialize the SDK requesting to autogenerate anonymous IDs
- Autogenerate an anonymous app user ID when the SDK is configured with `appUserIDsAreProvidedBy == AppUserIDProvider.RevenueCat`
- Update demo app to use this RC provided IDs (previously was generating anonymous IDs before configuring the SDK)


examples/rcbilling-demo/src/tests/main.test.ts
```diff
@@ -178,11 +186,8 @@ async function getAllElementsByLocator(
   return await locatorResult.all();
 }
 
-async function enterEmailAndContinue(
-  page: Page,
-  userId: string,
-): Promise<void> {
-  const email = `${userId}@revenuecat.com`;
```
antoniobg, line 185: The user ID was only used to generate an email. I replaced it by a hardcoded email since we don't know the app user ID in the test when it's autogenerated.

src/entities/identity.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Mode for identifying users in the Purchases SDK.
+ * @public
+ */
+export enum IdentityMode {
```
antoniobg, line 5: Unsure about this name and the usage of the "identity" word.

src/main.ts
```diff
@@ -154,7 +157,7 @@ export class Purchases {
    */
   static configure(
     apiKey: string,
-    appUserId: string,
+    appUserId: string | IdentityMode,
```
antoniobg, line 160: The reason why we decided to specifically request to autogenerate anonymous user IDs instead of not passing an app user ID at all (like in the mobile SDKs) was so developers are more conscious about this choice.

src/entities/identity.ts
aboedo: I'm torn on it frankly. Part of me thinks "we do talk about identity as a concept [in our docs"](https://www.revenuecat.com/docs/customers/user-ids), but also, it's still an awkward and not intuitive thing to write when using the API. 

I wonder if things might look better if we instead take inspiration from the observer mode changes? 

We used to ask people to enable or disable "observer mode", and that introduced a lot of questions and awkwardness, so we updated the API to be much more explicit about what it means in practice: 

https://www.revenuecat.com/docs/migrating-to-revenuecat/sdk-or-not/finishing-transactions

So now it looks like this (note the last line): 

```swift
Purchases.configure(
    with: .init(withAPIKey: "<public_sdk_key>")
          .with(purchasesAreCompletedBy: .myApp, storeKitVersion: .storeKit2)
)
```

Maybe we could do something similar, like `appUserIDsAreProvidedBy: myApp | revenueCat`? WDYT?

src/main.ts
aboedo: this makes me feel like it's the way it should always have been in mobile too 😅

src/entities/identity.ts
aboedo: I guess it looks different / more awkward in that you currently pass in the appUserID itself... maybe if we called it `RevenueCatAutoGenerated` or something? Let's bring this up in `#guild-sdks` for some team-wide wisdom

aboedo (commented):
looking great so far!

src/main.ts
nicfix: I like it and dislike it at the same time 😅 
The lazy me would like to just avoid to pass the parameter.
The functional programming me (which is usually less lazy) agrees with this implementation.

Proposing this change:
```suggestion
    appUserId: string | IdentityMode = IdentityMode.Anonymous,
```

In this way we make everyone happy:
```
// Lazy me
purchases.configure(API_KEY)

// Less lazy me
purchases.configure(API_KEY, IdentityMode.Anonymous)
```

Wdyt?

nicfix: The most annoying part I see with this is the usual TS/JS limitation when using named parameters.

We will always require users to set a value for `appUserId` just to be able to set the following parameter.

```
Purchases.configure(API_KEY, IdentityMode.Anonymous, { ...options})
```

But not so much we can do about it without breaking the previous usages.

nicfix (commented):
Added a proposal to have a default for appUserId for the extra lazy devs.

src/entities/identity.ts
antoniobg: I've given it a shot, adding an additional parameter `appUserIDsAreProvidedBy` and using overloads to avoid breaking changes to the API so we can still call configure without the new parameter, and avoiding passing an app user ID when `appUserIDsAreProvidedBy == AppUserIDsProvider.RevenueCat`. 
I've not an expert in typescript but I think it works as expected. 
Let me know what you think!

src/main.ts
antoniobg: > In this way we make everyone happy:
>
> ```
>// Lazy me
>purchases.configure(API_KEY)
>
>// Less lazy me
>purchases.configure(API_KEY, IdentityMode.Anonymous)
>```
>Wdyt?

The thing is that we want developers to specifically opt in autogenerated anonymous IDs. In the web, is more risky since these anonymous IDs are not persisted in the client in any way AFAIK.

src/main.ts
```diff
@@ -147,18 +147,41 @@ export class Purchases {
...
-   * @param httpConfig - Advanced http configuration to customise the SDK usage {@link HttpConfig}.
+   * @param httpConfig - Optional. Advanced http configuration to customise the SDK usage.
    * @throws {@link PurchasesError} if the API key or user id are invalid.
+   * @remarks For RevenueCat-managed identifiers, use configure with AppUserIDProvider.RevenueCat
    */
   static configure(
     apiKey: string,
-    appUserId: string | IdentityMode,
-    httpConfig: HttpConfig = defaultHttpConfig,
+    appUserId: string,
```
nicfix, line 161: I think having 2 overrides for this method is overkill.

In the end we need to handle 2 cases:
1. The appUserId is passed
2. The appUserId is not passed.

Proposal
```

class Purchases{

  static final GeneratedByRevenueCat='Purchases.GeneratedByRevenueCat'

  static configure(
     apiKey:string,
     appUserId: string ,
     httpConfig?: HttpConfig
  ) {
     
     
     if(appUserId === Purchases.GeneratedByRevenueCat){
        // generate the app user id
     } else {
       // use whatever was passed.
     }
     
  }
```

Then we can just use it:
```
// anonymous
Purchases.configure(API_KEY, Purchases.GeneratedByRevenueCat)

// non anonymous
Purchases.configure(API_KEY, 'my_custom_app_user_id')
```

nicfix (commented):
I think the new approach is too complicated.
I proposed a different solution.

src/entities/identity.ts
antoniobg: Gave it another shot and now we have named params

antoniobg:
We'll following a different approach for now. Closing this.

src/main.ts
aboedo: Sorry for being so late to this.
The proposal makes sense at a high level, although I'd suggest making the name more explicit. 

`Purchases.GeneratedByRevenueCat` sounds like you're saying that "the purchases are generated by RevenueCat" instead of "the app user ids are generated by RevenueCat", so I'd try to make sure that the name appUserID is there explicitly, like `Purchases.AutoGeneratedAppUserID` or something.

PR #234, opened 2024-11-14
Change "plan" to "subscription" on success page copy

MrEdwardo:
## Motivation / Description
We never mention the word "plan" in the checkout flow; could be confusing for users (Dipsea feedback). "Subscription" is more obvious from a user perspective.


vicfergar:
I guess we should also update this example CTA from _Choose plan_ to _Subscribe_

https://github.com/RevenueCat/purchases-js/blob/main/examples%2Frcbilling-demo%2Fsrc%2Fpages%2Fpaywall%2Findex.tsx#L81

nicfix (approved):
Thanks!

This will be shipped with the next release train on Wednesday.
Let me know if you'd like to expedite.

antoniobg (approved):

PR #240, opened 2024-11-21
Eco 1701 rcbpurchui pass the variables from the sdk to the paywall

guido732:
## Motivation / Description
Parses offering packages into an object keyed by package Identifier and a variables object as value.
Passes this object into the paywall initialization.


src/helpers/variables-helpers.ts
```diff
@@ -0,0 +1,291 @@
...
+    total_price_and_per_month: "",
+    total_price_and_per_month_full: "",
+    sub_price_per_month: "",
+    sub_price_per_week: "",
+    sub_duration: "",
+    sub_duration_in_months: "",
+    sub_period: "",
+    sub_period_length: "",
+    sub_period_abbreviated: "",
+    sub_offer_duration: undefined, // doesn't apply (yet)
```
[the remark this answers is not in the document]
nicfix: I just added it and pushed it to this branch

nicfix (commented):
I think there are still some variables that we can calculate since we have trials.

I added a test with a full offering fixture so that it might be easier to iterate.

I also needed to change the import from `/src/main` since it was failing when running the tests locally.

src/helpers/paywall-variables-helpers.ts
```diff
@@ -0,0 +1,367 @@
...
+  if (period.unit === "month" && period.number > 1) {
+    return formatPrice(
+      price.amountMicros / period.number,
+      price.currency,
+      selectedLocale,
+    );
+  }
+  if (period.unit === "week") {
+    return formatPrice(
+      (price.amountMicros * 4) / period.number,
```
nicfix, line 152: Don't we need a check that `period.number > 1` also for this division?

src/helpers/paywall-variables-helpers.ts
```diff
@@ -0,0 +1,367 @@
...
+      price.currency,
+      selectedLocale,
+    );
+  }
+  if (period.unit === "day") {
+    return formatPrice(
+      (price.amountMicros * 30) / period.number,
```
nicfix, line 159: Don't we need a check that `period.number > 1` also for this division?

src/helpers/paywall-variables-helpers.ts
```diff
@@ -0,0 +1,367 @@
...
+    price.amountMicros,
+    price.currency,
+    selectedLocale,
+  );
+  if (!period) return fallback;
+
+  if (period.unit === "year") {
+    return formatPrice(price.amountMicros / 12, price.currency, selectedLocale);
+  }
+  if (period.unit === "month" && period.number > 1) {
```
nicfix, line 143: Maybe you can early return here
```suggestion
  if(period.number == 0)
  {
     return fallback;
  }

  if (period.unit === "month" && period.number > 1) {
```

And avoid the following checks

src/helpers/paywall-variables-helpers.ts
```diff
@@ -0,0 +1,367 @@
...
+  const periodString = full ? period.unit : AbbreviatedPeriod[period.unit];
+  if (period.unit === "year") {
+    pricePerMonth = formatPrice(
+      price.amountMicros / 12,
+      price.currency,
+      selectedLocale,
+    );
+    return `${price.formattedPrice}/${period.number}${periodString}($${pricePerMonth}/(${monthPeriodString})`;
+  }
+  if (period.unit === "month") {
```
nicfix, line 191: Also here `period.number` might be 0 causing a division by zero.

src/tests/helpers/paywall-variables-helper.test.ts
nicfix, line 1: I would add a test to validate that the `sub_relative_discount` is calculated well w.r.t. the highest price.

The current test will definitely cover this specific use case but having also this rule covered might be a good idea if we change some logic in the future.

nicfix (changes requested):
A couple of suggestions on how we handle the divisions by zero and another test that might be useful.

Marking as request changes to make sure that we wait to ship this before having the next version of the SDK online. We need to update the package.json to make sure that nothing breaks

nicfix (commented):
I'd just remove the copy of the VariablesDictionary class.

We can import it internally directly from `purchases-ui-js` and I don't think we need to export it publicly, if so we can just export the same type by doing

```
import type {VariablesDictionary as UIVariablesDictionary} from '@revenuecat/purchases-ui-js`

...

export const VariablesDictionary = UIVariablesDictionary
```

src/helpers/paywall-variables-helpers.ts
```diff
@@ -0,0 +1,366 @@
...
+  const productType = pkg.rcBillingProduct.productType;
+  switch (productType) {
+    case ProductType.Subscription:
+      return pkg.rcBillingProduct.defaultSubscriptionOption;
+    case ProductType.Consumable || ProductType.NonConsumable:
+      return pkg.rcBillingProduct.defaultNonSubscriptionOption;
+  }
+}
+
+export type VariableDictionary = {
```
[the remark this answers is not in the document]
guido732: it wasn't to export it to users of the SDK but rather to be used elsewhere in the app, I'll just export it from `purchases-ui-js` and reuse it :) thanks for the suggestion

src/helpers/paywall-variables-helpers.ts
```diff
@@ -0,0 +1,366 @@
...
+
+function getPricePerPeriod(
+  formattedPrice: string,
+  product: SubscriptionOption,
+  full: boolean = false,
+) {
+  return `${formattedPrice}/${product.base.period?.number}${full ? (product.base.period as Period).unit : AbbreviatedPeriod[(product.base.period as Period).unit as keyof typeof AbbreviatedPeriod]}`;
+}
+
+function getDurationInMonths(period: Period) {
```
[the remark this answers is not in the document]
guido732: 🫠  localizations are hard

nicfix (approved):

tonidero:
@guido732 Changed labels to `pr:other`, since this is not a user facing change yet, so better to assign it to the the rest of the changelog.

PR #241, opened 2024-11-22
ECO-1704: SDK Localization

nicfix:
## Motivation / Description
This PR introduces localization for the SDK

## Changes introduced
* Introduced the Translator class and the `<Localized>` component.
* Added initial translations for en, es, it
* Converted all the hardcoded labels to use Translator or `<Localized>`

# Supported Languages
* ar - Arabic
* ca - Catalan
* cs - Czech
* da - Danish
* de - German
* el - Greek
* en - English
* es - Spanish
* fi - Finnish
* fr - French
* he - Hebrew
* hi - Hindi
* hr - Croatian
* hu - Hungarian
* id - Indonesian
* it - Italian
* ja - Japanese
* ko - Korean
* ms - Malay
* nl - Dutch
* no - Norwegian
* pl - Polish
* pt - Portuguese
* ro - Romanian
* ru - Russian
* sk - Slovak
* sv - Swedish
* th - Thai
* tr - Turkish
* uk - Ukranian
* vi - Vietnamese
* zh_Hans - Chinese
* zh_Hant - Chinese


src/ui/localization/period.svelte
guido732, line 1: what's this for? is this one still WIP? 
would this be the equivalent of doing a react hook? it's a svelte component that returns a value instead of an html block, but can access state/props with runes

src/ui/localization/translator.ts
```diff
@@ -0,0 +1,114 @@
...
+import es from "./locale/es.json";
+import it from "./locale/it.json";
+import { PeriodUnit } from "../../helpers/duration-helper";
+
+export class Translator {
+  private readonly locales: Record<string, LocaleTranslations> = {};
+
+  private static instance: Translator;
+
+  public static getInstance(): Translator {
```
guido732, line 11: shouldn't this one be private?

src/stories/purchase-flow.stories.svelte
```diff
@@ -158,4 +158,12 @@
...
\ No newline at end of file
+       args={{ ...defaultArgs, brandingInfo:{...brandingInfo, appearance:colorfulBrandingAppearance} }} />
+
+<Story name='Italian' args={{...defaultArgs, selectedLocale:'it', brandingInfo:brandingInfo}} />
+<Story name='it_IT' args={{...defaultArgs, selectedLocale:'it_IT', brandingInfo:brandingInfo}} />
+<Story name='itDashIT' args={{...defaultArgs, selectedLocale:'it-IT', brandingInfo:brandingInfo}} />
+
+<Story name='Spanish' args={{...defaultArgs, selectedLocale:'es', brandingInfo:brandingInfo}} />
+<Story name='es_ES' args={{...defaultArgs, selectedLocale:'es_ES', brandingInfo:brandingInfo}} />
+<Story name='esDashES' args={{...defaultArgs, selectedLocale:'es_ES', brandingInfo:brandingInfo}} />
```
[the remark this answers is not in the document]
nicfix: Updated in the new story using context

src/ui/localization/localized.svelte
```diff
@@ -0,0 +1,22 @@
...
+  // TODO: Find another way to handle the default Locale
+  const defaultLocale = "en";
+  const translatedLabel = $derived(Translator.translate(selectedLocale || defaultLocale, labelId || "", variables));
+</script>
+
+
+{#if translatedLabel}
+  {translatedLabel}
+{:else}
+  {@render children?.()}
```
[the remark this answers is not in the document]
nicfix: I am not sure that works in svelte since they explicitly suggest to use {#if } when in the "root" of the file.
I can try.

src/ui/rcb-ui.svelte
```diff
@@ -44,6 +45,8 @@
   let productDetails: Product | null = null;
   let paymentInfoCollectionMetadata: PurchaseResponse | null = null;
   let lastError: PurchaseFlowError | null = null;
+  let selectedLocale: string = "en";
```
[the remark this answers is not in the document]
nicfix: Changed this behaviour initializing an explicit Translator object and allowing for a defaultLocale (fallback)

src/ui/localization/translator.ts
nicfix: Removed completely the singleton

src/entities/purchase-params.ts
```diff
@@ -21,4 +21,26 @@ export interface PurchaseParams {
    * The email of the user. If undefined, RevenueCat will ask the customer for their email.
    */
   customerEmail?: string;
+
+  /**
+   * Optional customization for the translations used in the purchase flow.
+   * The keys are the locale codes and the values are objects with the same keys as the default translations.
+   */
+  customTranslations?: Record<string, Record<string, string>>;
```
guido732, line 29: this type is repeated in a few places. we can make a definition for this

src/main.ts
```diff
@@ -474,7 +477,16 @@ export class Purchases {
    */
   @requiresLoadedResources
   public purchase(params: PurchaseParams): Promise<PurchaseResult> {
-    const { rcPackage, purchaseOption, htmlTarget, customerEmail } = params;
+    const {
+      rcPackage,
+      purchaseOption,
+      htmlTarget,
+      customerEmail,
+      selectedLocale = "en",
```
guido732, line 485: are we using simple locales or `en_US` format?

src/ui/localization/localized.svelte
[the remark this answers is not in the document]
guido732: nonono, I was just wondering. I do like the syntax, is super clear, I was just wondering if it had any difference

src/ui/localization/translator.ts
```diff
@@ -0,0 +1,128 @@
+import en from "./locale/en.json";
+import es from "./locale/es.json";
+import it from "./locale/it.json";
```
guido732, line 3: should we make a barrel import here to have something like `import { en, es, it } from "./locale" ?

src/ui/rcb-ui.svelte
```diff
@@ -209,6 +218,7 @@
           </ModalHeader>
           {#if productDetails && purchaseOptionToUse}
             <StatePresentOffer
+              {selectedLocale}
```
guido732, line 221: it's complaining `selectedLocale` doesn't exist in the type definition

examples/rcbilling-demo/src/pages/paywall/index.tsx
```diff
@@ -114,6 +114,11 @@ const PaywallPage: React.FC = () => {
       const { customerInfo, redemptionInfo } = await purchases.purchase({
         rcPackage: pkg,
         purchaseOption: option,
+        useBrowserLocale: true,
```
guido732, line 117: doesn't exist in type

src/main.ts
guido732, line 1: it doesn't allow me to comment on another line that wasn't changed but this import is complaining

```
import type { RenderPaywallParams } from "./entities/render-paywall-params";
```

src/stories/fixtures.ts
guido732, line 1:
```
import {
  type Product,
  ProductType,
  type SubscriptionOption,
} from "../entities/offerings";
```

src/ui/branding-info-ui.svelte
```diff
@@ -14,21 +15,21 @@
         srcWebp={buildAssetURL(brandingInfo.app_icon_webp)}
       />
     {/if}
-    <span class="app-title">{brandingInfo.seller_company_name}</span>
+    <span class="app-title"><Localized>{brandingInfo.seller_company_name}</Localized></span>
```
guido732, line 18: this is complaining that it's missing props

src/ui/localization/localized.svelte
```diff
@@ -0,0 +1,35 @@
+<script lang="ts">
+  import { Translator } from "./translator";
+  import { getContext } from "svelte";
+  import { translatorContextKey } from "./constants";
+
+
+  export interface LocalizedProps {
```
guido732, line 7: this should be defined elsewhere to be exported or we need to change this to a module

src/ui/states/state-present-offer.svelte
```diff
@@ -30,34 +37,71 @@
 
     {#if isSubscription}
             <span class="rcb-product-price">
+
                 {#if subscriptionTrial?.periodDuration}
-                    {getTrialsLabel(subscriptionTrial.periodDuration)} free trial
+                  <Localized
```
guido732, line 42: it's complaining about props on all of these 🤔

package.json
```diff
@@ -91,5 +91,6 @@
     "http-status-codes": "^2.3.0",
     "@revenuecat/purchases-ui-js": "0.0.14"
   },
-  "dependencies": {}
+  "dependencies": {},
+  "packageManager": "pnpm@9.12.3+sha512.cce0f9de9c5a7c95bef944169cc5dfe8741abfb145078c0d508b868056848a87c81e626246cb60967cbd7fd29a6c062ef73ff840d96b3c86c40ac92cf4a813ee"
```
nicfix, line 95: Need to clean this up

src/main.ts
nicfix: The Translator supports both, if it doesn't find `en_US` in the available locales it tries with `en`.

src/ui/localization/translator.ts
nicfix: they are different files, I don't see the need personally.
Also they are json so I wouldn't add typescript export just for that.

src/stories/fixtures.ts
nicfix: I need to find a way to fix my WebStorm auto import

src/ui/localization/translator.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  private getLocaleInstance(locale: string): LocaleTranslations | undefined {
+    const potentialLocaleCode = locale.split("_")[0].split("-")[0];
+    return (
+      Translator.getInstance().locales[locale] ||
+      Translator.getInstance().locales[potentialLocaleCode]
+    );
+  }
+
+  public static translate(
```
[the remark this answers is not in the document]
nicfix: This is translating a label, we could call it `translateLabel`.
However I'd like to avoid having `translatePeriodLabel`. I think translate it's fine.

src/entities/purchase-params.ts
nicfix: Created the CustomTranslations type

src/ui/branding-info-ui.svelte
nicfix: Shouldn't be complaining anymore, all props are optional now.

src/ui/localization/localized.svelte
nicfix: Fixed it, just me being clumsy with Svelte.

I just wanted it to be the props of this component, no need to export them.

src/ui/states/state-present-offer.svelte
nicfix: Updated all of these, now all the props are optional

src/helpers/price-labels.ts
```diff
@@ -24,56 +25,31 @@ export const formatPrice = (
...
-};
-
-export const getRenewsLabel = (duration: string): string => {
+export const getTranslatedPeriodFrequency = (
+  duration: string,
+  translator: Translator,
+): string => {
   const period = parseISODuration(duration);
   if (!period) {
     return "unknown";
```
tonidero, line 34: Shouldn't normally happen but does this need to be localized as well?

src/helpers/price-labels.ts
```diff
@@ -1,4 +1,5 @@
-import { parseISODuration, type Period, PeriodUnit } from "./duration-helper";
+import { parseISODuration } from "./duration-helper";
+import { Translator } from "../ui/localization/translator";
 
 export const priceLabels: Record<string, string> = {
```
tonidero, line 4: Is this unused now?

src/ui/states/state-error.svelte
```diff
@@ -1,49 +1,83 @@
...
     );
   });
 
-  function getErrorTitle(): string {
+  function getTranslatedErrorTitle(): string {
     switch (lastError.errorCode) {
       case PurchaseFlowErrorCode.AlreadyPurchasedError:
         if (productDetails?.productType === ProductType.Subscription) {
-          return "Already subscribed";
+          return translator.translate("state_error.error_title_already_subscribed");
```
tonidero, line 31: I've been thinking, would it be possible to use some kind of enum for the translation keys instead of plain strings? That way we get compile time safety that there won't be any typos among other benefits. 

If we want to keep the jsons as strings, we can have the enum have an associated string value corresponding to the JSON key and have a test to make sure everything matches in the jsons.

src/tests/locale/locale.test.ts
```diff
@@ -0,0 +1,23 @@
+import { describe, expect, test } from "vitest";
+import { Translator } from "../../ui/localization/translator";
+
+const eqSet = (xs: Set<any>, ys: Set<any>) =>
+  xs.size === ys.size && [...xs].every((x) => ys.has(x));
+
+describe("The Translator class", () => {
+  test("should have the same keys in all languages", () => {
```
tonidero, line 8: 🎉

src/ui/localization/translator.ts
```diff
@@ -0,0 +1,246 @@
...
+ * This example will override the default translation for the email step title in the English locale.
+ * ```typescript
+ * const customTranslations = {
+ *  en: {
+ *    "state_needs_auth_info.email_step_title": "Billing email",
+ *   }
+ * }
+ * ```
+ */
+export type CustomTranslations = Record<string, Record<string, string>>;
```
tonidero, line 21: I think this would also benefit from having some sort of enum, for the supported languages and localization keys.

src/ui/localization/translator.ts
```diff
@@ -0,0 +1,246 @@
...
+
+  public constructor(
+    customTranslations: CustomTranslations = {},
+    public readonly selectedLocale: string = "en",
+    public readonly defaultLocale: string = "en",
+  ) {
+    this.locales = {
+      en: new LocaleTranslations(en, "en"),
+      es: new LocaleTranslations(es, "es"),
+      it: new LocaleTranslations(it, "it"),
```
tonidero, line 63: Nitpick but should we have a `supportedLocales` array and create this dictionary dynamically, so it's easier to add new languages? We could also then have a test to make sure all supported locales have corresponding localization jsons.

src/ui/localization/locale/en.json
```diff
@@ -0,0 +1,56 @@
+{
+  "periods.week": "{{amount}} week",
+  "periods.month": "{{amount}} month",
+  "periods.year": "{{amount}} year",
+  "periods.day": "{{amount}} day",
+  "periods.weekShort": "{{amount}}wk",
```
tonidero, line 6: Hmm maybe it's ok as is but would it be ok to add a delimiter like
```suggestion
  "periods.weekShort": "{{amount}}/wk",
```
Same for the other short periods. Not sure if we want to go the really short way though.

src/ui/localization/locale/es.json
```diff
@@ -0,0 +1,56 @@
+{
+  "periods.week": "{{amount}} semana",
+  "periods.month": "{{amount}} mes",
+  "periods.year": "{{amount}} año",
+  "periods.day": "{{amount}} día",
+  "periods.weekShort": "{{amount}}wk",
+  "periods.monthShort": "{{amount}}mo",
+  "periods.yearShort": "{{amount}}yr",
+  "periods.dayShort": "{{amount}}d",
```
tonidero, line 9: Hmm these short periods weren't translated to spanish it seems... Not sure if this is the best translation since it's pretty close to the non-short version but a suggestion:
```suggestion
  "periods.weekShort": "{{amount}}/sem",
  "periods.monthShort": "{{amount}}/mes",
  "periods.yearShort": "{{amount}}/año",
  "periods.dayShort": "{{amount}}/día",
```

src/ui/localization/translator.ts
```diff
@@ -0,0 +1,246 @@
...
+  }
+
+  get fallbackLocale(): string {
+    return (
+      this.getLocaleInstance(this.defaultLocale)?.localeKey ||
+      this.simplifyLocaleString(this.defaultLocale)
+    );
+  }
+
+  private simplifyLocaleString(locale: string): string {
```
tonidero, line 106: Nitpick but how about renaming to `getLanguageCodeString` or something like that? I believe it would be more precise that way.

src/entities/purchase-params.ts
```diff
@@ -21,4 +22,26 @@ export interface PurchaseParams {
    * The email of the user. If undefined, RevenueCat will ask the customer for their email.
    */
   customerEmail?: string;
+
+  /**
+   * Optional customization for the translations used in the purchase flow.
+   * The keys are the locale codes and the values are objects with the same keys as the default translations.
+   */
+  customTranslations?: CustomTranslations;
```
tonidero, line 30: TBH I'm a bit reticent to provide this API unless it's something requested often by developers... I think I would prefer to have more control over the text and lower the constraints in the future if it's really needed. Wdyt?

src/entities/purchase-params.ts
```diff
@@ -21,4 +22,26 @@ export interface PurchaseParams {
...
+  /**
+   * The locale to use for the purchase flow. If not specified, the device's locale will be used.
+   */
+  selectedLocale?: string;
+
+  /**
+   * Specifies if the locale should be picked automatically using the navigator.language property.
+   * Defaults to false.
+   */
+  useBrowserLocale?: boolean;
```
tonidero, line 41: Hmm I'm wondering whether we need to support this... If the developer wants to enable this, they can just pass the browser locale to `selectedLocale` I think? Seems like an unneeded API though I'm likely missing some use case 😅

tonidero (commented):
Looking good! Excited we were able to do this without using libraries :). I left some comments but I'm mostly wondering if it would be ok to change to use enums instead of plain strings for all these localization keys. I believe it could be very error-prone otherwise, but maybe this is how it works on web...

src/entities/purchase-params.ts
```diff
@@ -21,4 +22,26 @@ export interface PurchaseParams {
...
+
+  /**
+   * The default locale to use if the device's locale is not available.
+   */
+  defaultLocale?: string;
```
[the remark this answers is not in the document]
nicfix: The problem with that is that we will receive the language to use from the paywall in the renderPaywall method.

So basically no matter what's the SDK config we need to follow which language the paywall data passes to us when it will invoke the `purchase` method internally.

src/helpers/price-labels.ts
nicfix: Yes!

Good catch, I'll fix it

src/helpers/price-labels.ts
nicfix: Most probably. cleaning it up!

src/ui/states/state-error.svelte
nicfix: We can, I think that now that I figured out all the labels it would be easier to write an enum for this.

The only problem I see is that we need to keep in sync keys in the enum and keys in the files.
I'll add a test for that!

src/ui/localization/locale/en.json
nicfix: I think those are used in a variable string that already has delimiters so they would look like
```
3$/{{amount}}wk
```

The {{amount}} is the number of weeks, not the money per week, maybe amount is not the best name for that variable.

src/ui/localization/locale/es.json
nicfix: Given the answer from above I will take the suggestion removing the `/`

src/ui/localization/translator.ts
nicfix: Will do!

src/helpers/paywall-variables-helpers.ts
```diff
@@ -148,87 +123,91 @@ function getTotalPriceAndPerMonth({
...
       price.currency,
-      selectedLocale,
     );
-    if (period.number > 1) {
-      return `${price.formattedPrice}/${period.number}${periodString}(${pricePerMonth}/${monthPeriodString})`;
-    } else {
-      return `${price.formattedPrice}`;
+
+    if (period.number === 1) {
+      return price.formattedPrice;
```
guido732, line 150: should we move this a bit above as an early return pre calling translator?

src/helpers/price-labels.ts
```diff
@@ -1,4 +1,5 @@
-import { parseISODuration, type Period, PeriodUnit } from "./duration-helper";
+import { parseISODuration } from "./duration-helper";
+import { Translator } from "../ui/localization/translator";
```
guido732, line 2:
```suggestion
import type { Translator } from "../ui/localization/translator";
```

src/stories/state-needs-auth-info.stories.svelte
guido732, line 1: this file seems like it's missing formatting

src/stories/utils/with-context.svelte
```diff
@@ -0,0 +1,17 @@
+<script lang="ts">
+
+  import { setContext, type Snippet } from "svelte";
+
+  interface Props {
+    children?: Snippet;
+    context: Record<string, any>;
```
guido732, line 7: should context be an optional prop? we're defaulting it to an empty object below and some stories (`src/stories/state-needs-auth-info.stories.svelte`) aren't passing context, just to make things type-safe

src/stories/state-needs-payment-info.stories.svelte
guido732, line 1: same here with formatting

src/stories/state-present-offer.stories.svelte
guido732, line 1: same here with formatting

src/stories/state-success.stories.svelte
guido732, line 1: and in this one

src/ui/app-icon.svelte
```diff
@@ -9,23 +9,23 @@
...
+    .rcb-app-icon-picture-container {
+        height: 40px;
+    }
 
-  .rcb-app-icon.loading {
-    background-color: gray;
-  }
+    .rcb-app-icon.loading {
+        background-color: gray;
+    }
```
guido732, line 30: these also aren't formatted correctly

src/helpers/price-labels.ts
nicfix: Removed

src/helpers/price-labels.ts
nicfix: Translated it

src/ui/localization/translator.ts
nicfix: Renamed

src/entities/purchase-params.ts
nicfix: Removed

src/entities/purchase-params.ts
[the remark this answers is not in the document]
nicfix: Here's the point in which the paywall "passes" the locale and default locale to the purchase:
https://github.com/RevenueCat/purchases-js/pull/241/files#diff-4fab5baaca5c14d2de62d8d2fceef376ddddcc8e9509d86cfa5643f51b89ce3dR308-R309

src/ui/states/state-error.svelte
```diff
@@ -1,49 +1,83 @@
...
-  closeButtonTitle="Try Again"
+  closeButtonTitle={translator.translate("state_error.button_try_again")}
 >
   <IconError slot="icon" />
 
-  {lastError.getPublicErrorMessage(productDetails)}
+  {getTranslatedErrorMessage()}
   {#if supportEmail}
-    <br>If this error persists, please contact <a href="mailto:{supportEmail}">{supportEmail}</a>.
+    <br>
```
guido732, line 77:
```suggestion
    <br />
```

src/ui/states/state-success.svelte
```diff
@@ -3,23 +3,31 @@
...
+  import { Translator } from "../localization/translator";
+  import Localized from "../localization/localized.svelte";
 
   export let productDetails: Product | null = null;
   export let brandingInfo: BrandingInfoResponse | null = null;
   export let onContinue: () => void;
 
   const isSubscription = productDetails?.productType === ProductType.Subscription;
+  const translator: Translator = getContext(translatorContextKey) || Translator.fallback();
+  // TODO: Continue from here
```
guido732, line 17: was this left behind?

guido732 (changes requested):
I noticed a lot of formatting differences, like prettier isn't running on save or not reading the set of rules correctly

src/entities/purchase-params.ts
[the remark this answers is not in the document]
tonidero: Hmm we could also move the locale for the `renderPaywall` method off that method and into configuration time I think?

src/entities/purchase-params.ts
tonidero: Right... If we end up moving to using enums, it would mean that we might want to make the enum public, and if we want to change that enum for any reason, it would be a breaking change... I guess we can keep this interface using strings, but as I mentioned, I think that's error prone... I guess we can keep it for now even if not the easiest of interfaces, but as mentioned in a different comment, I would suggest moving it to configure time to clear the parameters for the actual purchase/renderPaywall methods.

src/ui/states/state-success.svelte
nicfix: Removing

src/entities/purchase-params.ts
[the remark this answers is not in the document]
nicfix: Not really because the data we receive for the supported and default locale are in the offering data itself.

Also why forcing this to be at configuration time?
Why do we need/want to add this there?

src/entities/purchase-params.ts
nicfix: I already migrated all the labels to be an enum!

Re: custom overrides, I think it's fine, worst case they will be ignored (if a key is wrong or renamed).

We should consider changes to the labels breaking anyway so better to know exactly which ones they are with an enum.

src/helpers/paywall-variables-helpers.ts
nicfix: Done!

nicfix:
@guido732 mistery solved, we missed the prettier config and plugins for svelte.

Here's a PR (to be merged on this branch) that just adds the plugin and formats all svelte files:
https://github.com/RevenueCat/purchases-js/pull/251

src/helpers/price-labels.ts
nicfix: Please check the formatting PR here:
https://github.com/RevenueCat/purchases-js/pull/251

src/stories/state-needs-auth-info.stories.svelte
nicfix: Please check the formatting PR here:
https://github.com/RevenueCat/purchases-js/pull/251

src/stories/state-present-offer.stories.svelte
nicfix: Please check the formatting PR here:
https://github.com/RevenueCat/purchases-js/pull/251

src/stories/state-success.stories.svelte
nicfix: Please check the formatting PR here:
https://github.com/RevenueCat/purchases-js/pull/251

src/ui/app-icon.svelte
nicfix: Please check the formatting PR here:
https://github.com/RevenueCat/purchases-js/pull/251

src/ui/states/state-error.svelte
nicfix: Please check the formatting PR here:
https://github.com/RevenueCat/purchases-js/pull/251

src/main.ts
```diff
@@ -89,6 +90,10 @@ export { OfferingKeyword } from "./entities/get-offerings-params";
 export type { PurchaseParams } from "./entities/purchase-params";
 export type { RedemptionInfo } from "./entities/redemption-info";
 export type { PurchaseResult } from "./entities/purchase-result";
+export type {
+  TranslationVariables,
```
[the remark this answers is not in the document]
nicfix: Nope, removing

src/ui/localization/translator.ts
```diff
@@ -0,0 +1,304 @@
...
+        acc.replace(
+          `{{${key}}}`,
+          `${value === undefined || value === null ? "" : value}`,
+        ),
+      label,
+    );
+  }
+
+  public translate(
+    labelId: string,
```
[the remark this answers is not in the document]
nicfix: Renaming it

src/stories/utils/with-context.svelte
nicfix: Done

src/ui/states/state-success.svelte
nicfix: Removed

src/entities/purchase-params.ts
nicfix: I like that, I'll give it a go tomorrow since I am also renaming the labelId.

Thanks for all the feedback @tonidero!

nicfix:
Merged the PR with the introduction of prettier for svelte.

src/ui/localization/locale/es.json
```diff
@@ -0,0 +1,57 @@
...
+  "periods.dayFrequencyPlural": "cada {{amount}} días",
+  "state_present_offer.product_title": "{{productTitle}}",
+  "state_present_offer.product_description": "{{productDescription}}",
+  "state_present_offer.product_price": "{{productPrice}}",
+  "state_present_offer.free_trial_duration": "Prueba gratuita por {{trialDuration}}",
+  "state_present_offer.price_after_free_trial": "{{productPrice}} después del final de la prueba",
+  "state_present_offer.renewal_frequency": "Se renueva {{frequency}}",
+  "state_present_offer.continues_until_cancelled": "Continúa hasta que se cancele",
+  "state_present_offer.cancel_anytime": "Cancela en cualquier momento",
+  "state_needs_auth_info.email_step_title": "Contraseña email de facturación",
```
vicfergar, line 32: I don't know where `Contraseña` came from, but it's definitely nonsense 😄 
```suggestion
  "state_needs_auth_info.email_step_title": "Email para facturación",
```

src/ui/localization/locale/es.json
```diff
@@ -0,0 +1,57 @@
...
+  "periods.dayFrequency": "cada día",
+  "periods.unknownFrequency": "desconocida",
+  "periods.weekFrequencyPlural": "cada {{amount}} semanas",
+  "periods.monthFrequencyPlural": "cada {{amount}} meses",
+  "periods.yearFrequencyPlural": "cada {{amount}} años",
+  "periods.dayFrequencyPlural": "cada {{amount}} días",
+  "state_present_offer.product_title": "{{productTitle}}",
+  "state_present_offer.product_description": "{{productDescription}}",
+  "state_present_offer.product_price": "{{productPrice}}",
+  "state_present_offer.free_trial_duration": "Prueba gratuita por {{trialDuration}}",
```
vicfergar, line 27:
```suggestion
  "state_present_offer.free_trial_duration": "Prueba gratuita de {{trialDuration}}",
```

src/ui/localization/locale/es.json
```diff
@@ -0,0 +1,57 @@
...
+  "state_present_offer.price_after_free_trial": "{{productPrice}} después del final de la prueba",
```
vicfergar, line 28:
```suggestion
  "state_present_offer.price_after_free_trial": "{{productPrice}} tras el periodo de prueba",
```

guido732 (changes requested):

src/ui/localization/locale/es.json
```diff
@@ -0,0 +1,57 @@
...
+  "state_needs_payment_info.button_start_trial": "Comenzar la prueba",
+  "state_success.purchase_successful": "Adquisición exitosa",
+  "state_success.subscription_now_active": "Tu suscripción está ahora activa",
+  "state_success.button_close": "Cerrar",
+  "state_error.if_error_persists": "Si el error persiste, por favor contacta a ",
+  "state_error.error_title_already_subscribed": "Ya suscrito",
+  "state_error.error_title_already_purchased": "Ya comprado",
+  "state_error.error_title_other_errors": "Algo salió mal",
+  "state_error.error_message_already_subscribed": "Ya estás suscrito a este producto.",
+  "state_error.error_message_already_purchased": "Ya has comprado este producto.",
```
vicfergar, line 47:
```suggestion
  "state_error.error_title_already_subscribed": "Suscripción existente",
  "state_error.error_title_already_purchased": "Producto ya adquirido",
  "state_error.error_title_other_errors": "Algo salió mal",
  "state_error.error_message_already_subscribed": "Ya tienes una suscripción activa a este producto.",
  "state_error.error_message_already_purchased": "Ya has comprado este producto.",
```

src/ui/localization/locale/es.json
```diff
@@ -0,0 +1,57 @@
...
+  "state_error.error_message_missing_email_error": "Se requiere un email para completar la compra.",
+  "state_error.error_message_network_error": "Error de red. Por favor, verifica tu conexión a internet.",
+  "state_error.error_message_error_charging_payment": "Pago fallido.",
```
vicfergar, line 50:
```suggestion
  "state_error.error_message_error_charging_payment": "El pago no se completó.",
```

src/ui/localization/locale/en.json
```diff
@@ -0,0 +1,57 @@
...
+  "periods.yearFrequencyPlural": "every {{amount}} years",
+  "periods.dayFrequencyPlural": "every {{amount}} days",
+  "state_present_offer.product_title": "{{productTitle}}",
+  "state_present_offer.product_description": "{{productDescription}}",
+  "state_present_offer.product_price": "{{productPrice}}",
+  "state_present_offer.free_trial_duration": "{{trialDuration}} free trial",
+  "state_present_offer.price_after_free_trial": "{{productPrice}} after end of trial",
+  "state_present_offer.renewal_frequency": "Renews {{frequency}}",
+  "state_present_offer.continues_until_cancelled": "Continues until cancelled",
+  "state_present_offer.cancel_anytime": "Cancel anytime",
```
vicfergar, line 31: This feels kind of repetitive, doesn’t it? I'd suggest discarding `state_present_offer.continues_until_cancelled`

src/ui/localization/locale/en.json
```diff
@@ -0,0 +1,57 @@
...
+  "periods.yearFrequency": "yearly",
+  "periods.dayFrequency": "daily",
+  "periods.unknownFrequency": "unknown",
+  "periods.weekFrequencyPlural": "every {{amount}} weeks",
+  "periods.monthFrequencyPlural": "every {{amount}} months",
+  "periods.yearFrequencyPlural": "every {{amount}} years",
+  "periods.dayFrequencyPlural": "every {{amount}} days",
+  "state_present_offer.product_title": "{{productTitle}}",
+  "state_present_offer.product_description": "{{productDescription}}",
+  "state_present_offer.product_price": "{{productPrice}}",
```
vicfergar, line 26: what's the benefit of having this 3 keys?

src/ui/states/state-error.svelte
```diff
@@ -1,49 +1,116 @@
...
   <IconError slot="icon" />
 
-  {lastError.getPublicErrorMessage(productDetails)}
+  {getTranslatedErrorMessage()}
   {#if supportEmail}
-    <br>If this error persists, please contact <a href="mailto:{supportEmail}">{supportEmail}</a>.
+    <br />
+    <Localized labelId={LocalizationKeys.StateErrorIfErrorPersists}>
+      If this error persists, please contact
+    </Localized>
```
vicfergar, line 113: We should remove the content here, right?
```suggestion
    <Localized labelId={LocalizationKeys.StateErrorIfErrorPersists} />
```

src/ui/localization/locale/es.json
vicfergar: Actually, I’m wondering if just `Email` could be enough. By the context, it’s clear that it will be used to send receipts and information related to the purchase, isn’t it?

src/entities/purchase-params.ts
```diff
@@ -21,4 +22,21 @@ export interface PurchaseParams {
    * The email of the user. If undefined, RevenueCat will ask the customer for their email.
    */
   customerEmail?: string;
+
+  /**
+   * Optional customization for the translations used in the purchase flow.
+   * The keys are the locale codes and the values are objects with the same keys as the default translations.
+   */
+  customTranslations?: CustomTranslations;
```
tonidero, line 30: I was chatting about this with @aboedo today. I haven't been on the conversations that lead to this new API but one thought that came was that, instead of having a client-side API to override the strings, it might be a good idea to provide all texts from the backend, so they can be dynamically configured from there if needed. Specially with the localization changes we've been working on for paywalls, this might be much easier. 

I thought this idea could be great, but not sure if we would have the bandwidth to do it right now... Having said that, a new public API would be trickier to remove once it's published, so if we want to go that route, it might be worth considering not adding this parameter yet (not sure how much of a priority it has)

src/ui/localization/locale/en.json
nicfix: That the customer can customize them, doing stuff like

```
Price: {{productPrice}}
```

src/entities/purchase-params.ts
nicfix: I think there are pro and cons to forcing backend customization only.

Long term we want to allow developers to customize their labels, but on the mid-short term we won't be able to build an editor right now. I think this is a compromise that would unlock customers to customize the current version of the SDK now.

I see the backend and hardcoded version living together as follows
````
const finalLabels= {
   ...defaultTranslations,
   ...serverSideCustomTranslations,
   ...customTranslationsPassedAsParameter
}
```

In this way, if someone locally wants to generate something dynamically in their app with their logic for one label they can. I also think that who uses the sdk directly should be able to update their website relatively easily and fix the `customTranslations` object.

I see how the problem might be bigger on mobile where you don't have control on which version of the app is being executed on the phone even if you update it, but on web this is a lesser problem IMO.

I am fine removing this public parameter, however, this is a direct customer request that we had from Dipsea.
We do not need it for the Paywalls ship so we can definitely try to discuss it more and exclude it from this PR.

src/ui/localization/locale/es.json
nicfix: Maybe we can just use 
```suggestion
  "state_needs_auth_info.email_step_title": "Email de facturación",
```

wdyt @vicfergar ?

src/ui/localization/locale/zh_Hant.json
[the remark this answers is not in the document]
nicfix: @MrEdwardo had some AI help

src/ui/localization/localized.svelte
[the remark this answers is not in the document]
nicfix: The idea is, if you pass a label and the label is found it uses the label, otherwise it uses whatever you passed as children, if any.

src/ui/localization/locale/en.json
nicfix: The specific labels were repeated to allow developers to customize whatever was there.
That's why all the labels are there even if they say the same thing in our current choices, they might be customized to say something else by the customers.

Reusing the same label might cause one customization in one point to affect other ones.

src/entities/purchase-params.ts
nicfix: Removed the public `customTranslations?: CustomTranslations;` parameter for now.

We can iterate again when we'll work on customizing the WPL (which was the original request from Dipsea).

src/ui/localization/locale/en.json
vicfergar: But we're always showing these 3 lines, right?

src/ui/localization/localized.svelte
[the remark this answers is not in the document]
nicfix: I'll remove the children here!

src/ui/states/state-needs-payment-info.svelte
```diff
@@ -84,14 +93,25 @@
...
 <div>
   {#if stripe && clientSecret}
     <ModalHeader>
-      <div style="display: flex; align-items: center; justify-content: baseline;">
+      <div
+        style="display: flex; align-items: center; justify-content: baseline;"
+      >
         <IconLock />
-        <div style="margin-left: 10px">Secure Checkout</div>
+        <div style="margin-left: 10px">
```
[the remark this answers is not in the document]
nicfix: Will do

src/ui/localization/locale/en.json
nicfix: I think I am not understanding your comment. 

What do you mean by repetitive?
Do you want to remove the actual lines from the component or do you want to reuse 2 times the same label?

api-report/purchases-js.api.md
```diff
@@ -23,6 +23,9 @@ export interface CustomerInfo {
     readonly requestDate: Date;
 }
 
+// @public
+export type CustomTranslations = Record<string, Record<string, string>>;
```
tonidero, line 27: I think we don't need to export this anymore after removing the parameter?

src/main.ts
```diff
@@ -292,6 +302,8 @@ export class Purchases {
         rcPackage: pkg,
         htmlTarget: paywallParams.purchaseHtmlTarget,
         customerEmail: paywallParams.customerEmail,
+        selectedLocale: selectedLocale,
+        defaultLocale: offering.paywall_components?.default_locale || "en",
```
tonidero, line 306: Nitpick but might be worth extracting `en` to a constant? Like `const DEFAULT_LOCALE = "en"`. I see we also use this in a few places so probably reuse it in those.

src/ui/localization/translator.ts
tonidero: Bump to this one. Small cleanup but just wondering if we can build this dynamically, and maybe even have a test that checks that all locales in the `locale` folder are added to the `SUPPORTED_LOCALES` constant. This way we won't forget to add new locales here.

src/ui/localization/translator.ts
```diff
@@ -0,0 +1,371 @@
...
+import sk from "./locale/sk.json";
+import sv from "./locale/sv.json";
+import th from "./locale/th.json";
+import tr from "./locale/tr.json";
+import uk from "./locale/uk.json";
+import vi from "./locale/vi.json";
+
+import type { PeriodUnit } from "../../helpers/duration-helper";
+
+export type emptyString = "";
```
tonidero, line 37: Hmm I wonder if this should be uppercased
```suggestion
export type EmptyString = "";
```

src/entities/purchase-params.ts
```diff
@@ -21,4 +21,15 @@ export interface PurchaseParams {
    * The email of the user. If undefined, RevenueCat will ask the customer for their email.
    */
   customerEmail?: string;
+
+  /**
+   * The locale to use for the purchase flow. If not specified, the device's locale will be used.
+   */
+  selectedLocale?: string;
```
tonidero, line 28: It might be worth to write an E2E test verifying that localization is working, even if it's a basic one.

tonidero (approved):
Some comments but nothing really blocking. Looks great!

src/ui/localization/translator.ts
nicfix: > This way we won't forget to add new locales here.

Yeah, which is already the case given that @MrEdwardo added more locales 😅 

I'll create a reference list for now, I would avoid "scanning folders" or something similar just to make sure that we still bundle all these in the UMD and we don't rely on them being hosted somewhere.

src/ui/localization/translator.ts
nicfix: Yep, type not const. I refactored it from const to type and forgot to uppercase it, thanks!

api-report/purchases-js.api.md
nicfix: Yep, removing it.

nicfix: Done

src/main.ts
nicfix: Done, called it `englishLocale` just to avoid colliding with `defaultLocale`.

It's not a best practice to call constant variables with all caps in JS.

src/ui/localization/translator.ts
nicfix: I moved the supported languages in a dedicated file and initialized the `locales` attribute here using that object.

I wouldn't make this dynamic with the included files since this has to by resolved at build time.

I added a test to check that every file that is present in the `/locale` folder gets imported in the `supportedLanguages` object.

tonidero: Yeah agreed. I think adding a test is enough. Thanks!

guido732 (approved):

src/entities/purchase-params.ts
nicfix: I just added one!

Both with and without paywall!

tonidero:
Might be worth mentioning the supported locales in the PR description, since this PR will go to the changelog and some people might want to look at it (though we could also just add a link to our docs)

nicfix:
I agree @tonidero I'll list them down tomorrow before merging the PR.

@MrEdwardo did you start any work on the docs for this feature?
If not I'll follow up tomorrow

tonidero:
Note that we can always ship and add the link to the docs later, since the PR description is editable. No need to delay shipping due to that

vicfergar (approved):

PR #243, opened 2024-11-27
[ECO-1743] Make payment method "selected" color match accent color

vicfergar:
## Motivation / Description

Currently the "selected" color is a fixed blue, which doesn't work for a lot of themes. It should ideally match the accent color from customization.

## Changes introduced

95f4e75 - fix: Style of tabs and blocks
4c74268 - fix: Set text customized colors
6c9520c - feat: Pass `seller_company_name` to `StripeElements`
1d4d564 - fix: Close button color

| Before | After | 
|----------|----------|
| image | image |

| Before | After | 
|----------|----------|
| image | image |

| Before | After | 
|----------|----------|
| image | image |

| Before | After | 
|----------|----------|
| image | image |

### DIFF
--- a/src/ui/close-button.svelte
+++ b/src/ui/close-button.svelte
@@ -18,10 +18,13 @@
     padding: 0px;
     height: 24px;
     border-radius: 50%;
-    color: var(--rc-color-gray-text-dark);
+    color: var(--rc-color-grey-text-light);
   }
 
   .close-button:focus {
     outline: 2px solid var(--rc-color-focus);
+  }
+  .close-button:hover {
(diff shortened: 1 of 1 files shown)


src/ui/close-button.svelte
```diff
@@ -1,11 +1,13 @@
 <script lang="ts">
-  import Icon from "../assets/close.svg";
+  import Icon from "../assets/close.svg?raw";
```
antoniobg, line 2: Just curious, what does this mean?

vicfergar: It reads the svs as row data, I need to do this in order to change the fill color of the close button

PR #246, opened 2024-11-27
fix: Do not show error message if Google Pay or Apple Pay window is closed

vicfergar:

### DIFF
--- a/src/ui/states/state-needs-payment-info.svelte
+++ b/src/ui/states/state-needs-payment-info.svelte
@@ -67,8 +67,13 @@
     });
 
     if (result.error) {
-      // payment failed, notify user
       processing = false;
+
+      if (!result.error.setup_intent?.last_setup_error) {
+        return;
+      }
+
+      // payment failed, notify user
       onError(
         new PurchaseFlowError(
           PurchaseFlowErrorCode.StripeError,


src/ui/states/state-needs-payment-info.svelte
```diff
@@ -67,8 +67,13 @@
     });
 
     if (result.error) {
-      // payment failed, notify user
       processing = false;
+
+      if (result.error.type === "validation_error") {
```
vicfergar, line 72: `validation_error`s does not affect the setup_intent state so it's safe to keep using the same Stripe Elements instance to complete the purchase

src/ui/states/state-needs-payment-info.svelte
```diff
@@ -67,8 +67,13 @@
     });
 
     if (result.error) {
-      // payment failed, notify user
       processing = false;
+
+      if (!result.error.setup_intent?.last_setup_error) {
```
vicfergar, line 72: This ensures that the `setup_intent.setup_intent_failed` is still not sent, so it's safe to reuse the same Stripe Elements instance

tonidero: Hmm is this the recommended way to detect closure of the Google/Apple pay prompts? It would be great to have something more explicit, but not familiar with the Stripe API... Would it be possible to use [status](https://docs.stripe.com/api/setup_intents/object#setup_intent_object-status)?

tonidero: Also should we assume that `setup_intent will be there? As in maybe we can do:
```suggestion
      if (result.error.setup_intent !== undefined && !result.error.setup_intent?.last_setup_error) {
```
or something like this?

tonidero (approved):
Just a question, but I'm not too familiar with this, so maybe this is the best way to detect this. Looks good otherwise!

PR #252, opened 2024-12-03
adds test for puchase flow using rc_paywalls

guido732:
## Motivation / Description
Adds an E2E test for the purchase flow using the RC Paywalls

### DIFF
--- a/examples/rcbilling-demo/src/tests/main.test.ts
+++ b/examples/rcbilling-demo/src/tests/main.test.ts
@@ -3,6 +3,8 @@ import { Locator } from "playwright";
 
 const _LOCAL_URL = "http://localhost:3001/";
 const CARD_SELECTOR = "div.card";
+const PACKAGE_SELECTOR = "button.package";
+const RC_PAYWALL_TEST_OFFERING_ID = "e2e_tests_web_paywalls";
 
 test.describe("Main", () => {
   test.afterEach(({ browser }) => {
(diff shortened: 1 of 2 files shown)


examples/rcbilling-demo/src/tests/main.test.ts
```diff
@@ -3,6 +3,8 @@ import { Locator } from "playwright";
 
 const _LOCAL_URL = "http://localhost:3001/";
 const CARD_SELECTOR = "div.card";
+const PACKAGE_SELECTOR = "button.package";
+const RC_PAYWALL_TEST_OFFERING_ID = "e2e_tests_web_paywalls";
```
guido732, line 7: @tonidero do you think we should move this to an env variable? if so, can you help me with it? 

I don't think its strictly necessary since it's not something that exposes any sensible data, but I'll let you be the judge of that

tonidero: I think it's fine for the offering id to be hardcoded. Not really a secret. Could be worth moving to an ENV if we wanted to have different ones, but I don't think that's the case, so it's probably fine to leave here.

tonidero:
Hmm I don't see the changes anymore... Only a console.log in the tests? probably some bad merge?

nicfix (approved):
Please rollback the changes to the state-needs-payment-info, apart from that LGTM

PR #253, opened 2024-12-03
test

guido732:
## Motivation / Description
Adds proper ESlint config and autofix on save + pre-commit hook config for husky
This is working locally, though requiring some configuration for vscode to auto-fix on save, which using Webstorm I don't expect to work, might require extra work for that, but at least picks up errors on attempting to commit and shows them in the editor itself

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -1,8 +1,4 @@
-import {
-  type Offering,
-  type Offerings,
-  type Package,
-} from "./entities/offerings";
+import type { Offering, Offerings, Package } from "./entities/offerings";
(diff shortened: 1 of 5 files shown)


src/main.ts
```diff
@@ -517,7 +513,7 @@ export class Purchases {
 
     const certainHTMLTarget = resolvedHTMLTarget as unknown as HTMLElement;
 
-    const asModal = !Boolean(htmlTarget);
+    const asModal = !htmlTarget;
```
guido732, line 516: this one was automatically picked up and fixed, redundant boolean conversion rule

nicfix (approved):

PR #254, opened 2024-12-04
adds e2e for variable replacement

guido732:
## Motivation / Description
Add a test that checks variable replacement is working correctly in the SDK


examples/rcbilling-demo/src/tests/main.test.ts
```diff
@@ -77,6 +79,38 @@ test.describe("Main", () => {
     const title = page.getByText("E2E Tests for Purchases JS");
     await expect(title).toBeVisible();
   });
+  test("Can render an RC Paywall using variables", async ({
```
tonidero, line 82: We probably should start splitting tests into multiple files 😅 but that can come in a separate PR.

examples/rcbilling-demo/src/tests/main.test.ts
```diff
@@ -77,6 +79,38 @@ test.describe("Main", () => {
...
+    // Gets all packages
+    const packageCards = await getAllElementsByLocator(page, PACKAGE_SELECTOR);
+
+    // Get the purchase button as a Locator
+    const purchaseButton = (
+      await getAllElementsByLocator(page, "button.purchase-button")
+    )[0];
+
+    await expect(purchaseButton).toContainText(
+      "PURCHASE FOR $1.25/1wk($5.00/mo)",
```
tonidero, line 103: I'm thinking... Would this break with the localization changes? We probably should hardcode the locale for the E2E tests, not sure if we're doing that... In any case, it's ok if it's passing based off `main` :)

tonidero (approved):
LGTM!

examples/rcbilling-demo/src/tests/main.test.ts
guido732: agreed (eventually)

examples/rcbilling-demo/src/tests/main.test.ts
guido732: we're not doing that, but I agree it _could_ fail, although I'm guessing the fallback language for our CI pipe "browser" is english so it should be picking that up. Either way, we should have new localized tests soon that cover these cases :)

PR #258, opened 2024-12-09
upgrade dependencies

guido732:
## Motivation / Description
Upgrade dependencies to prepare for new version of `purchases-ui-js`


package.json
```diff
@@ -56,47 +56,47 @@
...
     "prettier-plugin-svelte": "^3.3.2",
-    "react": "^18.2.0",
-    "react-dom": "^18.2.0",
-    "storybook": "^8.3.6",
-    "svelte": "^5.5.0",
-    "svelte-check": "^4.0.6",
-    "svelte-stripe": "^1.1.2",
-    "typedoc": "^0.26.11",
-    "typescript": "^5.6.3",
+    "react": "^19.0.0",
```
nicfix, line 90: Why do we have the react dependency here?

nicfix (approved):
Just left a comment, we might need to cleanup react from the library devDependencies.

package.json
guido732: tbh I don't know, I thought it was intentional 🤔  I'll remove it and rebuild deps

PR #273, opened 2025-01-13
Track simple SDK_INITIALIZED event

nicfix:
## Description

Adds tracking for the SDK initialized event.

## Event Model

An event consists of:
- `id`: uuidv4 generated in the frontend for SDK events as soon as the event happens. Note it is technically redundant given we also have the `trace_id` and the `trace_index`.
- `type`: the type of event sent. Includes a prefix for Web Billing `rcb_`.  Follows same pattern as customer center events and paywall events.
- `trace_id`: uuidv4 generated in the frontend as soon as the SDK is initialized. All events of the same browser session will be sending this same `trace_id` until the page is reloaded or exited.
- ~`trace_index`: an incremental 0-based index that describes the partial order of the events in the same trace.~
- `timestamp`: an epoch in ms of when the event happened in the frontend using the local clock. Follows same pattern as customer center events and paywall events.

## Fake Timers

Vitest fake timers proofed a pretty powerful tool for testing the retry logic of tracking events. It helps give determinism and reduce the duration of the tests. No more arbitrary waits of X seconds.

## E2E

The e2e test currently fails given the backend does not accept the shape of the event. It will be solved once the backend PR is merged.

## Retry Logic

The events are sent right away as they happen. If there is an unexpected response from the browser or an error occurs with the request, the events will resent in batch with an exponential back-off. Starts with 4s and has a maximum interval of 60s. The back-off is reset upon successful submission of the events.

⚠️ Currently the events are sent an infinite amount of times until the request succeeds.
⚠️ Since events are fired right away if there have not been any errors there may be multiple requests in-flight that send events to the backend if none has yet failed.


src/behavioural-events/events-tracker.ts
```diff
@@ -0,0 +1,104 @@
...
+    console.debug(`Events tracker created for traceId ${this.traceId}`);
+
+    // This interval will flush the events every 5 seconds
+    // since we don't await the flush when calling trackEvent
+    // this interval will retry in case of failure
+    // using the mutex and posting only if there are events in the queue.
+    this.intervalHandle = setInterval(() => {
+      console.debug("Interval flush");
+      this.flushEvents();
+    }, 5000);
```
polmiro, line 28: Note: Maybe the retry should be exponential?

src/behavioural-events/events-tracker.ts
```diff
@@ -0,0 +1,110 @@
+import type { SDKInitializedEvent, TrackedEvent } from "./event-types";
+import { v4 as uuid } from "uuid";
+import { RC_ENDPOINT, VERSION } from "../helpers/constants";
+import { HttpMethods } from "msw";
+import { getHeaders } from "../networking/http-client";
+import { defaultHttpConfig, type HttpConfig } from "../entities/http-config";
+import { ExponentialInterval } from "../helpers/exponential-interval";
+
+const MIN_INTERVAL_RETRY = 2_000;
+const MAX_INTERVAL_RETRY = 60_000;
```
polmiro, line 10: Thoughts on these timings? The first one will always be skipped so it will be series 4s, 8s, 16s, 32s, 60s, 60s, 60s, ...

src/behavioural-events/events.ts
```diff
@@ -0,0 +1,46 @@
+import { v4 as uuidv4 } from "uuid";
+import { type Trace } from "./trace";
+
+type EventType = "rcb_sdk_initialized";
+
+export abstract class BaseEvent {
+  public readonly id: string;
```
polmiro, line 7: Note: Using `uuid.v4()` and generating it on the client for frontend events.

src/behavioural-events/events.ts
```diff
@@ -0,0 +1,46 @@
...
+  public readonly timestamp: number;
```
polmiro, line 8: Note: Using `timestamp` with ms from epoch time for consistency with customer center and paywall events.

src/behavioural-events/events.ts
```diff
@@ -0,0 +1,46 @@
...
+  public readonly type: string;
```
polmiro, line 9: Note: Using a single `type` field instead of `type` + `event_name` for consistency with customer center and paywall events.

src/behavioural-events/events.ts
```diff
@@ -0,0 +1,46 @@
...
+  public readonly traceId: string;
+  public readonly traceIndex: number;
```
polmiro, line 11: Sending the `traceIndex` in case some manipulation of the timestamp for partial ordering events in the same trace is needed.

src/behavioural-events/events-tracker.ts
```diff
@@ -0,0 +1,108 @@
...
+      return;
+    }
+    if (this.flushingMutex) {
+      console.debug("Already flushing");
+      return;
+    }
+
+    this.flushingMutex = true;
+    console.debug("Acquired flushing mutex");
+    if (this.eventsQueue.length > 0) {
```
vicfergar, line 66: is this check needed?

nicfix (commented):
The absolute base identifiers for the events look good to me, @estefaniarabadan @toppare wdyt?

These are the ids we aim to have in ALL events as mandatory, is there something we are missing here that you think we should have?

src/behavioural-events/events-tracker.ts
```diff
@@ -0,0 +1,108 @@
...
+      this.postEvents(this.eventsQueue)
+        .then((response) => {
+          if (response.status === 200 || response.status === 201) {
+            console.debug("Events flushed successfully");
+            this.eventsQueue.splice(0, this.eventsQueue.length);
+            this.intervalManager.resetInterval(5000);
```
antoniobg, line 72: Should we have a constant for `5000`?

src/behavioural-events/events-tracker.ts
polmiro: Good catch, updating.

src/behavioural-events/events-tracker.ts
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    this.flushingMutex = true;
+    console.debug("Acquired flushing mutex");
+    this.postEvents(this.eventsQueue)
+      .then((response) => {
+        if (response.status === 200 || response.status === 201) {
+          console.debug("Events flushed successfully");
+          this.eventsQueue.splice(0, this.eventsQueue.length);
+          this.intervalManager.resetInterval(5000);
```
vicfergar, line 71: Is this correct? I think it should be `MIN_INTERVAL_RETRY`
```suggestion
          this.intervalManager.resetInterval(MIN_INTERVAL_RETRY);
```

src/behavioural-events/events-tracker.ts
polmiro: Actually I should remove that and store the initial interval so the rest can just use it. Thanks!

src/helpers/exponential-interval.ts
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  public increaseInterval() {
+    clearInterval(this.intervalHandle);
+    this.interval = Math.min(this.interval * 2, this.maxInterval);
+    console.debug(`Increasing interval to ${this.interval} ms`);
+    this.startInterval();
+  }
+
+  public resetInterval(initialInterval: number) {
```
vicfergar, line 33: Is `initialInterval` really needed here?

polmiro: hehe I was just making the change. Thanks

toppare:
> The absolute base identifiers for the events look good to me, @estefaniarabadan @toppare wdyt?
> 
> These are the ids we aim to have in ALL events as mandatory, is there something we are missing here that you think we should have?

@nicfix how will we connect these events to actual users or projects?

src/behavioural-events/events-tracker.ts
```diff
@@ -0,0 +1,106 @@
...
+      console.debug("Already flushing");
+      return;
+    }
+
+    this.flushingMutex = true;
+    console.debug("Acquired flushing mutex");
+    this.postEvents(this.eventsQueue)
+      .then((response) => {
+        if (response.status === 200 || response.status === 201) {
+          console.debug("Events flushed successfully");
```
tonidero, line 69: I guess these are left for debugging? I think this is fine but if we want to leave any logs semi-permanently which may be useful, we can use `Logger.debugLog`. So they are only logged if logs are enabled.

polmiro: Good call, on it

examples/rcbilling-demo/src/tests/main.test.ts
```diff
@@ -271,8 +270,50 @@ test.describe("Main", () => {
       await expect(page.getByText(title)).toBeVisible();
     });
   });
+
+  test.only("Tracks event SDK Initialized", async ({
```
alfondotnet, line 274: make sure to remove the `only` 

note: add a an eslint rule for this if there is none yet.

polmiro: Ah, just noticed there is one for `playwright.config.ts` but I guess that does not apply here.

polmiro: Actually the default of `vitest` is already to not allow this in CI.
Screenshot 2025-01-27 at 10 26 59

But this test is actually from the e2e suite and the config in `playwright.config.ts` takes care of it. Notice the build actually failed because of it.

src/behavioural-events/events.ts
```diff
@@ -0,0 +1,46 @@
+import { v4 as uuidv4 } from "uuid";
+import { type Trace } from "./trace";
+
+type EventType = "rcb_sdk_initialized";
+
+export abstract class BaseEvent {
+  public readonly id: string;
+  public readonly timestamp: number;
+  public readonly type: string;
+  public readonly traceId: string;
```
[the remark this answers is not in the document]
estefaniarabadan: hey @polmiro , what this uuid is referring to? generating in the front end for what? like a session id?

src/behavioural-events/trace.ts
```diff
@@ -0,0 +1,17 @@
...
+  public readonly trace_id: string;
+  private trace_index: number;
+
+  constructor() {
+    this.trace_id = uuidv4();
+    this.trace_index = 0;
+  }
+
+  public nextTraceIndex(): number {
+    const current_index = this.trace_index;
```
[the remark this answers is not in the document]
polmiro: This has been addressed, I had to turn the method and also the trackEvent methods into async in order to wait for the possible mutex lock.

src/helpers/exponential-interval.ts
```diff
@@ -0,0 +1,45 @@
...
+    this.maxInterval = maxInterval;
+    this.callback = callback;
+    this.startInterval();
+  }
+
+  private startInterval() {
+    this.intervalHandle = setInterval(() => {
+      console.debug(`Interval flush after ${this.interval} ms`);
+      this.callback();
+    }, this.interval);
```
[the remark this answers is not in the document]
polmiro: This has been reimplemented using `setTimeout`.

src/behavioural-events/flush-manager.ts
```diff
@@ -0,0 +1,95 @@
...
+    this.clearTimeout();
+  }
+
+  public schedule(delay?: number) {
+    if (this.timeoutId !== undefined) {
+      Logger.log(`Already scheduled`);
+      return;
+    }
+
+    Logger.log(`Scheduling callback after ${this.currentDelay}ms delay`);
```
antoniobg, line 43: Should we use `debugLog` instead? (Applies to several lines in this file, and maybe other)

polmiro: Yes! 🤦🏻  going to review the rest in case there are any other.

polmiro:
For the record, `trace_index` was deemed redundant and removed given we have timestamp in miliseconds.

PR #275, opened 2025-01-17
chore: add console.error in example app when API is not set

alfondotnet:


nicfix:
Circleci is failing due to not having the correct tags.

Added `pr:other` this change doesn't require us to generate a new sdk version!

nicfix (approved):

PR #280, opened 2025-01-23
Fix/purchase-flow-need-auth-info accessibility improvements

Tuuben:
## Motivation / Description

Hey, I was going through the codebase to familiarize myself with it and noted some accessibility stuff is missing from inputs and buttons so thought I'd commit a small fix for it. Feel free to just disregard this PR incase it's not relevant / other internal changes incoming 😊 

## Changes introduced

- Fixes email input to display the correct keyboard type (email) on virtual keyboards.
- Fixes close button focus with screen readers.
- Improves email input descriptions with screen readers.
- Explicitly sets tabIndex for safari browsers to make buttons tab focusable, [issue described here](https://stackoverflow.com/questions/1848390/safari-ignoring-tabindex)

### DIFF
--- a/src/ui/button.svelte
+++ b/src/ui/button.svelte
@@ -3,14 +3,17 @@
   export let disabled = false;
   export let testId: string | undefined = undefined;
   export let type: "reset" | "submit" | "button" | null | undefined = undefined;
+  export let ariaBusy = false;
 </script>
 
 <button
+  tabindex={0}
   on:click
   class={`intent-${intent}`}
   {disabled}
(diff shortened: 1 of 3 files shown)


alfondotnet:
Hi @Tuuben!

Thanks for this PR! We are changing the views quite significantly, but I'll ping you to rebase as soon as we merge so we can get some improvements out! 

We are going to get rid of the e-mail view (state-needs-auth-info) and have an e-mail input alongside the payment form, so this might not be needed though

PR #281, opened 2025-01-23
WEB-1998: Allow metadata to be sent at purchase

nicfix:
## Motivation / Description
We want to allow developers to send metadata when invoking the `purchase` method.
We also want to pass through the commonly used `utm` parameters.

The metadata will be propagated to the Stripe customer and to the RevenueCat events.

This PR introduces the possibility of passing arbitrary string metadata and automatically sends along the `utm` parameters in the current query string. It's possible to opt-out of the automatic behaviour through Configuration.

## Changes introduced
* Added `metadata` to the PurchaseConfig
* Added `FlagConfig`
* Added an automatic mechanism to read the utm params from the query string and send them through in the `metadata` object.


src/networking/backend.ts
```diff
@@ -104,6 +107,10 @@ export class Backend {
       supports_direct_payment: true,
     };
 
+    if (purchaseMetadata) {
+      requestBody.purchase_metadata = purchaseMetadata;
```
tonidero, line 111: For the existing metadata in the POST /receipt endpoint in iOS, we accept a `metadata` field, of type `Record<string, string>`. So we only accept string values: https://github.com/RevenueCat/purchases-ios/blob/15c67454493cd0a483e2a3f445d7216675462893/Sources/Networking/Operations/PostReceiptDataOperation.swift#L144

I'm not sure if we want to reuse the same behavior, or have an entirely different mechanism for purchases-js... Note that any differences might make it more difficult to integrate with the hybrids later + we need to make sure the data matches the behavior in the backend, so I would try to consolidate the behavior unless there is a reason to keep it separate.

nicfix: Renamed it into `metadata`

src/entities/purchase-params.ts
```diff
@@ -1,4 +1,8 @@
-import { type Package, type PurchaseOption } from "./offerings";
+import {
+  type Package,
+  type PurchaseMetadata,
+  type PurchaseOption,
+} from "./offerings";
```
SimonGodefroid, line 5:
```suggestion
import type {
 Package,
 PurchaseMetadata,
 PurchaseOption,
} from "./offerings";
```

src/helpers/utm-params.ts
```diff
@@ -0,0 +1,21 @@
+export const autoParseUTMParams = () => {
+  const params = new URLSearchParams(window.location.search);
+
+  const possible_params = [
```
SimonGodefroid, line 4:
```suggestion
  const possibleParams = [
```

src/tests/helpers/utm-params.test.ts
```diff
@@ -0,0 +1,58 @@
+import { describe, expect, test } from "vitest";
+import { spyOn } from "@vitest/spy";
+import { autoParseUTMParams } from "../../helpers/utm-params";
+
+describe("autoParseUTMParams", () => {
+  test("should return an empty object if no utm param is set", () => {
```
SimonGodefroid, line 6: Most other tests in the file so not start with "should".

src/entities/purchase-params.ts
nicfix: Done, thanks!

src/helpers/utm-params.ts
nicfix: Done, thanks!

src/tests/helpers/utm-params.test.ts
nicfix: Removed the "should", thanks!

src/entities/offerings.ts
```diff
@@ -381,6 +381,13 @@ export interface Offerings {
   readonly current: Offering | null;
 }
 
+/**
+ * Metadata that can be passed to the backend when making a purchase.
+ * They will propagate to the payment gateway (i.e. Stripe) and to the RevenueCat.
```
tonidero, line 386:
```suggestion
 * They will propagate to the payment gateway (i.e. Stripe) and to RevenueCat.
```

api-report/purchases-js.api.md
```diff
@@ -277,6 +285,8 @@ export interface PurchaseParams {
     customerEmail?: string;
     defaultLocale?: string;
     htmlTarget?: HTMLElement;
+    // Warning: (ae-forgotten-export) The symbol "PurchaseMetadata" needs to be exported by the entry point Purchases.es.d.ts
```
tonidero, line 288: Seems we need to export the type from `main.ts`?

nicfix: Nice catch, fixed it!

PR #282, opened 2025-01-27
Revamped views

alfondotnet:
First pass at revamping our purchase flow.

Keep in mind this PR goes against a branch `fon/new-sdk-views`. I'd love feedback on what we have until now

Things left to do:

- @iga-gawronska Needs to figure out sizing adjustments to our Stripe form
- Need to implement floating labels for the e-mail form (but lower priority)
- Need to add controls to Storybook for the different appearances (and see if I can fix the fact that there's implementation duplication in the stories)
- Clean up the Stripe file onto a hook-like file
- Translate the texts I have changed to all other languages


src/main.ts
```diff
@@ -515,7 +515,6 @@ export class Purchases {
 
     const certainHTMLTarget = resolvedHTMLTarget as unknown as HTMLElement;
 
-    const asModal = !htmlTarget;
```
alfondotnet, line 518: @nicfix removing asModal here as I have removed backdrop etc (since it's always a 100vwx100vh page)

src/main.ts
```diff
@@ -525,6 +524,16 @@ export class Purchases {
     const localeToBeUsed = selectedLocale || defaultLocale;
 
     return new Promise((resolve, reject) => {
+      window.history.pushState({ checkoutOpen: true }, "");
+      const onClose = () => {
+        window.removeEventListener("popstate", onClose);
```
alfondotnet, line 529: so when pressing the back button on the browser we close our modal..

I have **not** implemented yet that for example if you are in the checkout page and press back, you go back to the e-mail page. Right now it will close the whole modal.

src/ui/layout/row-layout.svelte
```diff
@@ -13,7 +15,40 @@
...
+  .column.gap-small {
+    gap: var(--rc-spacing-gapSmall-mobile);
+  }
+
+  .column.gap-medium {
+    gap: var(--rc-spacing-gapMedium-mobile);
+  }
+
+  .column.gap-large {
+    gap: var(--rc-spacing-gapXXLarge-mobile);
```
alfondotnet, line 38: to review

src/ui/localization/locale/en.json
```diff
@@ -29,16 +29,16 @@
...
   "state_needs_auth_info.email_input_label": "Email",
   "state_needs_auth_info.email_input_placeholder": "john@appleseed.com",
   "state_needs_auth_info.button_continue": "Continue",
-  "state_needs_payment_info.payment_step_title": "Secure checkout",
-  "state_needs_payment_info.button_pay": "Pay",
+  "state_needs_payment_info.payment_step_title": "Secure checkout by RevenueCat",
+  "state_needs_payment_info.button_pay": "Pay now",
   "state_needs_payment_info.button_start_trial": "Start Trial",
-  "state_success.purchase_successful": "Purchase successful",
+  "state_success.purchase_successful": "Payment complete",
```
alfondotnet, line 39: need to translate these new values to all languages

src/ui/modal.svelte
```diff
@@ -1,43 +1,53 @@
...
+  export let delayFade = 130;
 </script>
 
-<div class="rcb-modal-container">
-  <main class="rcb-modal-main" {style}>
+<div class="rcb-modal-container" {style}>
+  <svelte:element
+    this={as}
+    class="rcb-modal-main"
+    transition:fade={{ duration: 500, delay: delayFade }}
```
alfondotnet, line 14: opening the modal directly felt a bit off. Adding a slight transition makes it look nice imo

src/main.ts
nicfix: @alfondotnet this was needed to distinguish when we are taking the entire screen from when we are showing it in an element passed by the developer.

We should not lose the functionality.
Using 100vw x 100vh would break whoever is using that mode.

alfondotnet:
Makes sense , will add a test for that

Alfonso Embid-Desmet
Staff Engineer 💻 @ revenuecat.com

src/main.ts
alfondotnet: as discussed IRL, will not tackle this for this PR as the mode is currently broken. have a WIP fixing this holistically and adding tests

PR #283, opened 2025-01-28
[WEB-2016] Rename Web Billing

antoniobg:
## Motivation / Description

We're renaming "RevenueCat Billing" to "Web Billing".

## Changes introduced
- Rename "RevenueCat Billing" to "Web Billing" in the documentation
- Introduce a new `webBillingProduct` property in the `Package` entity
- Deprecate the existing `rcBillingProduct` property, in favor of `webBillingProduct`

## TODOs:
- [ ] Update code snippets in the docs


examples/rcbilling-demo/README.md
tonidero, line 1: I guess we need to update the folder name as well. NABD though.

src/tests/helpers/paywall-variables-helper.test.ts
```diff
@@ -5,567 +5,582 @@ import { type VariableDictionary } from "@revenuecat/purchases-ui-js";
...
-              },
-              cycleCount: 1,
-              price: null,
-            },
-          },
-        },
-        defaultNonSubscriptionOption: null,
-      },
+      rcBillingProduct: trialProduct,
+      webBillingProduct: trialProduct,
```
tonidero, line 583: Nice cleanup!

nicfix:
I don't think we can call this a minor given that it changes a public interface

antoniobg:
> I don't think we can call this a minor given that it changes a public interface

I wasn't sure about that, to be honest, but since it's not a breaking change as the old property is still there I decided to go with a minor. Happy to change it though. 

@tonidero how does Core SDK handle these changes?

tonidero:
Well my rule of thumb is:
- If there is a breaking change, then it's a major
- If no breaking changes and there are new features (usually accompanied by new APIs, but not necessarily), a minor
- Anything else a patch

In this PR, I could go with minor since there is a new API but I would say there isn't new functionality, even if there is a new API. It's just a rename of an existing one. So I would consider this a patch IMO.

tonidero:
Actually let's follow Semver and make it a minor: https://semver.org/#spec-item-7

image

tonidero (approved):
LGTM!

nicfix (approved):
Nice move replicating the product in the public api.
Thanks for going the extra mile and changing also the public api.

PR #284, opened 2025-01-28
Quieter test logs

polmiro:
## Motivation / Description

There were some noisy logs happening.

## Changes introduced

- Use `Logger.debugLog` instead of `console.debug`.
- Remove redundant query parameters in request handler matcher

### DIFF
--- a/src/tests/test-responses.ts
+++ b/src/tests/test-responses.ts
@@ -258,8 +258,8 @@ export function getRequestHandlers(): RequestHandler[] {
     const body = offeringsResponsesPerUserId[userId]!;
     const url = `http://localhost:8000/v1/subscribers/${userId}/offerings`;
     requestHandlers.push(
-      http.get(url, () => {
-        APIGetRequest({ url: url });
+      http.get(url, ({ request }) => {
+        APIGetRequest({ url: request.url });
         return HttpResponse.json(body, { status: 200 });
       }),
     );
@@ -267,10 +267,10 @@ export function getRequestHandlers(): RequestHandler[] {
 
   Object.keys(productsResponsesPerUserId).forEach((userId: string) => {
     const body = productsResponsesPerUserId[userId]!;
-    const url = `http://localhost:8000/rcbilling/v1/subscribers/${userId}/products?id=monthly&id=monthly_2`;
+    const url = `http://localhost:8000/rcbilling/v1/subscribers/${userId}/products`;
     requestHandlers.push(
-      http.get(url, () => {
-        APIGetRequest({ url: url });
+      http.get(url, ({ request }) => {
+        APIGetRequest({ url: request.url });
         return HttpResponse.json(body, { status: 200 });
       }),
     );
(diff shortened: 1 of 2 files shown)


src/ui/localization/translator.ts
```diff
@@ -93,7 +94,7 @@ export class Translator {
         currency,
       }).format(price);
     } catch {
-      console.debug(
+      Logger.debugLog(
```
tonidero, line 97: Hmm I'm wondering if these should actually be `Logger.errorLog`... But we can keep as is for now.

src/tests/test-responses.ts
```diff
@@ -267,7 +267,7 @@ export function getRequestHandlers(): RequestHandler[] {
 
   Object.keys(productsResponsesPerUserId).forEach((userId: string) => {
     const body = productsResponsesPerUserId[userId]!;
-    const url = `http://localhost:8000/rcbilling/v1/subscribers/${userId}/products?id=monthly&id=monthly_2`;
+    const url = `http://localhost:8000/rcbilling/v1/subscribers/${userId}/products`;
```
tonidero, line 270: Hmm seems like the code still adds the products as query parameters to the url?  https://github.com/RevenueCat/purchases-js/blob/95743588d01dfdf30d595ce835f86efe0ab89a68/src/networking/endpoints.ts#L56

I guess tests may pass since url matching doesn't take into account query parameters I believe... But might be good idea to keep them here so it's clearer that the URL is supposed to have them? (we actually are not including the currency code, which is included in the code though)

tonidero (approved):
A couple of comments, but not really blocking. I'm ok with the changes

src/tests/test-responses.ts
polmiro: Well, that was my train of thought but I guess with a different opinion. Since the request matchers ignore the query parameters it seemed miss-leading to keep them. One may think it will only match if the query parameters are present but it is not true. But I can go ahead and update the behavior so that the code takes them into account inside of the match and returns some sort of failure response if they are not there. Curious to see if all the tests will be passing.

src/ui/localization/translator.ts
polmiro: I can go ahead and make the change to `Logger.errorLog` if you give me a 👍🏻 . I initially did not want to change the behavior so did not even consider it.

tonidero: Let's do the change then 👍. I believe it's an error, so it makes more sense to me!

src/tests/test-responses.ts
tonidero: Right.... I don't feel strongly TBH, just a thought... I'm ok keeping this as is!

polmiro: So, I tried to only return a successful response if those ids were passed, but that makes some tests fail since their requests do not send both ids or none at all.

For the scope of this PR I decided to:
- Remove the query parameters from the request matcher given that they are not ignored and may lead to thinking the request handler is more constrained when it really is not.
- Changed the url passed to the spy `APIGetRequest` so that it includes the full url received which does include the query parameters.

All tests kept working as they were.

polmiro:
@nicfix is it normal to get these Chromatic diffs even when I haven't changed anything related?

nicfix:
In this case yes, what's changing is the rendering of the stripe elements that we don't control perfectly in chromatic.
My guess there is that it was still loading the internal iframe when it took the snapshot.

PR #285, opened 2025-01-29
[WEB-1910] Track checkout session start event

polmiro:
## Motivation / Description

Adds tracking for checkout session start event.

I used the label `pr-other` given this is getting merged into `user-events` branch.

## Changes introduced

- Move tracker specific tests to from `events.test.ts` to `behavioral-events/events-tracker.test.js`.
- Add a helper to track event `CHECKOUT_SESSION_START`.
- Track event `CHECKOUT_SESSION_START` upon mounting purchases ui.


src/behavioural-events/events.ts
```diff
@@ -1,49 +1,79 @@
+import { camelToUnderscore } from "../helpers/camel-to-underscore";
 import { v4 as uuidv4 } from "uuid";
 
-type EventType = "web_billing_sdk_initialized";
+enum EventType {
+  SDK_INITIALIZED = "web_billing_sdk_initialized",
+  CHECKOUT_SESSION_START = "web_billing_checkout_session_start",
```
antoniobg, line 6: I just realized that in other enums we use `PascalCase`. Should do the same here?

antoniobg (approved):
Left a minor comment about the case format to use in enums

src/behavioural-events/events.ts
polmiro: Ah they where first constants, maybe that is why. I will review. Thanks

PR #287, opened 2025-01-30
[WEB-1911] Track billing email entry events

polmiro:
## Motivation / Description

Adds tracking events for the billing email entry screen:
- impression
- submission
- error
- close

## Changes introduced

- Add (new) UI tests for testing svelte component behaviors fast without a fully feature e2e.
- Add the tracking for the events.
- Renamed / Rearranged some of the tests files.


src/behavioural-events/events-tracker.ts
```diff
@@ -23,6 +27,12 @@ export interface IEventsTracker {
   trackCheckoutSessionStart(
     props: CheckoutSessionStartEventProps,
   ): Promise<void>;
+  trackBillingEmailEntryImpression(props: UserEventProps): Promise<void>;
+  trackBillingEmailEntrySubmit(props: UserEventProps): Promise<void>;
+  trackBillingEmailEntryDismiss(props: UserEventProps): Promise<void>;
+  trackBillingEmailEntryError(
+    props: BillingEmailEntryErrorEventProps,
+  ): Promise<void>;
```
tonidero, line 35: Hmm sorry I missed this on previous PRs but do these need to be part of the public API? I would try to avoid that as much as possible specially considering these seem to only be needed internally?

tonidero: Fortunately we haven't made a release with these new SDK methods, so I believe we're on time to remove them and avoid the breaking change.

polmiro: We don't! I fixed this in the upstream branch `web-1910` but I did not do the change on this one yet.

polmiro: No problem, thanks for reviewing!

But maybe wait for the reviews since I am still working on this draft PR and there may be larger changes. I will move to "Ready for review" when it's good to go from my side.

src/ui/states/state-needs-auth-info.svelte
```diff
@@ -12,22 +12,39 @@
...
   const handleContinue = async () => {
-    const verificationErrors = validateEmail(email);
-    if (verificationErrors) {
-      error = verificationErrors;
+    const emailError = validateEmail(email);
+    if (emailError) {
+      eventsTracker.trackBillingEmailEntryError({
+        appUserId: appUserId,
+        userIsAnonymous,
+        errorCode: 0,
```
polmiro, line 43: This needs to be fixed. The local handling of errors does not map to a `PurchaseFlowErrorCode`. Perhaps these error should only have a message and no error code? Thoughts?

tonidero: Hmm maybe we could expose the backend provided error directly and send that?

src/ui/rcb-ui.svelte
```diff
@@ -91,6 +96,10 @@
     new Translator(customTranslations, selectedLocale, defaultLocale),
   );
 
+  setContext(eventsTrackerContextKey, eventsTracker);
```
nicfix, line 99: Maybe is better to have a `TrackingContext` interface that contains all of those and set the context only once.

Something like
```
interface TrackingContext{
   tracker: EventsTracker;
   appUserId: string;
   userIsAnonymous: boolean;
}

[...]

setContext(trackingContextKey, {
   tracker: eventsTracker,
   appUserId: appUserId,
   userIsAnonymous: useIsAnonymous
})
```

In this way we have to propagate less keys down the nested components and we do less calls to getContext

nicfix (commented):
Left a couple of suggestions.

The most important is to reduce the number of `setContext` calls.

PR #288, opened 2025-01-30
Add translator script + update translations

alfondotnet:


src/ui/localization/locale/de.json
```diff
@@ -17,38 +16,40 @@
   "periods.monthFrequency": "monatlich",
   "periods.yearFrequency": "jährlich",
```
alfondotnet, line 17: @jefago does this look better to you?

PR #292, opened 2025-02-01
[WEB-2053] Add an internal public function to track events from WPL

polmiro:
## Motivation / Description

WPL will be tracking some events that need to be in the same trace as the rest of the `purchases-js` tracked events. There are two main options:
1. Expose a `_trackEvent` public method marked as `@internal` for `api-extractor`.
2. Expose the `_trace_id` public value marked as `@internal` for `api-extractor`.

This pull requests implements the first option which seems like the best one giving `@internal` tag is meant for such use cases and hides the `trace_id` implementation detail.

> The @internal modifier is one of the four release tags. It indicates that an API item is meant only for usage by other NPM packages from the same maintainer. Third parties should never use "internal" APIs. To emphasize this, an underscore prefix should be used in the name of a declaration with an (explicit) @internal tag. API Extractor validates this naming convention and will report [ae-internal-missing-underscore](https://api-extractor.com/pages/messages/ae-internal-missing-underscore/) if the underscore is missing.

See docs for [here](https://api-extractor.com/pages/tsdoc/tag_internal/).

## Changes introduced

- Expose a public `_trackEvent` marked `@internal` on `Purchases` class.
- Run `npm audit fix`.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -50,6 +50,7 @@ import { PaywallDefaultContainerZIndex } from "./ui/theme/constants";
 import { parseOfferingIntoVariables } from "./helpers/paywall-variables-helpers";
 import { Translator } from "./ui/localization/translator";
 import { englishLocale } from "./ui/localization/constants";
+import type { TrackEventProps } from "./behavioural-events/events-tracker";
 import EventsTracker, {
   type IEventsTracker,
 } from "./behavioural-events/events-tracker";
@@ -679,5 +680,14 @@ export class Purchases {
   public static generateRevenueCatAnonymousAppUserId(): string {
     const uuid = crypto.randomUUID();
     return `${ANONYMOUS_PREFIX}${uuid.replace(/-/g, "")}`;
+  }
+
+  /**
+   * Track an event.
+   * @param props - The properties of the event.
+   * @internal
+   */
+  public _trackEvent(props: TrackEventProps): void {
+    this.eventsTracker.trackEvent(props);
   }
 }
--- a/examples/rcbilling-demo/package-lock.json
+++ b/examples/rcbilling-demo/package-lock.json
@@ -414,14 +414,13 @@
       }
     },
     "node_modules/@esbuild/aix-ppc64": {
-      "version": "0.24.0",
-      "resolved": "https://registry.npmjs.org/@esbuild/aix-ppc64/-/aix-ppc64-0.24.0.tgz",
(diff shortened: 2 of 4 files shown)


src/main.ts
```diff
@@ -686,6 +687,15 @@ export class Purchases {
     return `${ANONYMOUS_PREFIX}${uuid.replace(/-/g, "")}`;
   }
 
+  /**
+   * Track an event.
+   * @param props - The properties of the event.
+   * @internal
+   */
+  public _trackEvent(props: TrackEventProps): void {
```
tonidero, line 695: Do we also need to export `TrackEventProps` as an internal API?

polmiro: Good call. Just updated. Although afaik, given those are not exposed it should not have a side-effect.

nicfix (approved):
I don’t think we have a more elegant solution for this, the only thing I would do is limit the type of events that can be tracked to the ones we expect from the WPL.

Another option is to move the EventsTracker class outside of the sdk and share only the track_id .
I think is better to just expose a method, it gives us much more control on how the track_id is used.

polmiro:
Great then 👌🏻 

In regards to constraining here which events can be sent from WPL, my preference would be to type those there based on constraining `TrackEventProps`. Otherwise in order to track more events from WPL we have to update the SDK and there is no real need for that. The backend endpoint already accepts an open format and is able to disregard events as it pleases needed be.

src/main.ts
tonidero: Related, but I'm not sure if the convention for interfaces is to also prefix it with an underscore like `_TrackEventProps`

tonidero:
One point in favor of constraining is that a developer could potentially start adding extra data/events we don't need, so I do like the idea of constraining those even if it requires a SDK update...

polmiro:
IMHO the @internal should be enough and such change would mostly cause us friction without safeguard us. Nothing really stops the developer from sending a request to the backend with a different `event_name` or any other properties. Even more so, if they are using plain JS, nothing will be type checking the parameters either.

The model of an event should be specified where it is sourced and that should be enough. This is the reason I generalized the types in the endpoint ingestion side. So that we did not have to go to Khepri necessarily and update the endpoint whenever a new field of an event or a new event is tracked.

I am happy to consider it tough if you feels strongly about it @tonidero . Let me know if I convinced you or not.

tonidero:
That's a good point yeah... Considering the backend needs to have the validation anyway, I agree with keeping it as is 👍

polmiro:
Screenshot 2025-02-03 at 17 44 59

So I tried to export `TrackEventProps` but I was getting this error in the screenshot above. I have digged a bit into it and already spent some time trying to fix it but looks like it may be related to a bug of the library. In any case, since the root function is already marked as `@internal` and all it does is avoid exposing it's types and avoid documenting it I think it's okay to keep on without it.

PR #294, opened 2025-02-03
[WEB-1916] Track CheckoutSessionEnd event

polmiro:
## Motivation / Description

Tracks `CheckoutSessionEnd` event. The start and end of the checkout are now track upon mounting and closing of the `.purchase(...)`.

## Changes introduced

- Move tracking of `CheckoutSessionStart` to `Purchases`.
- Add tracking of `CheckoutSessionEnd` to `Purchases`.


src/main.ts
```diff
@@ -546,15 +552,30 @@ export class Purchases {
 
     const localeToBeUsed = selectedLocale || defaultLocale;
 
+    const purchaseOptionToUse =
+      purchaseOption ?? rcPackage.rcBillingProduct.defaultPurchaseOption;
```
nicfix, line 556: the `rcBillingProduct` property was recently marked as deprecated. It's still there but might change in the future.
```suggestion
      purchaseOption ?? rcPackage.webBillingProduct.defaultPurchaseOption;
```

nicfix (commented):
It would be better to use the `webBillingProduct` property that should be filled with the very same value of the `rcBillingProduct`.

The `rcBillingProduct` has been deprecated and will be removed

src/main.ts
polmiro: The `user-events` branch is now a bit behind so `webBillingProduct` does not exist here. I will look into this once I rebase `user-events` from `main` or merge `main` into `user-events`.

src/behavioural-events/tracked-events.ts
```diff
@@ -56,6 +58,16 @@ export interface CheckoutSessionStartEvent extends IEvent {
   };
 }
 
+export interface CheckoutSessionEndEvent extends IEvent {
+  eventName: TrackedEventName.CheckoutSessionEnd;
+  properties: {
+    outcome: "finished" | "closed" | "errored";
```
tonidero, line 64: Hmm I'm wondering if this should be 3 separate events, instead of a property with 3 values... Since we have properties that are only needed for specific results. You might have talked about this already, in which case feel free to ignore.

tonidero (approved):
This is looking good to me! Just a question about maybe splitting the event into 3 to have a more defined schema.

src/behavioural-events/tracked-events.ts
polmiro: Good call @tonidero .

I took a bit of a middle-ground solution here and did not go all the way through. I think it is clearer from a user analytics perspective to have a `CheckoutSessionStart` and `CheckoutSessionEnd` with its outcome. But I have  tighten the types so that they are clearer and narrower on the SDK side.

Let me know if you had concerns on using the same type with different outcomes for the tracked event sent to the backend or this looks good to you otherwise.

Thanks!

tonidero (approved):
I like it more like this yeah thanks! As for having a single event with outcome vs 3 different events, I can see arguments for both TBH, but not a strong opinion, however data prefers it I guess :)

PR #299, opened 2025-02-04
[WEB-2056] Use a different domain for event tracking

polmiro:
## Motivation / Description

This is a safeguard in case at some point we were to run into domain blocking issues. It happened once in the past and it became problematic for some users.

### DIFF
--- a/.env.production
+++ b/.env.production
@@ -1,2 +1,3 @@
 VITE_RC_ENDPOINT=https://api.revenuecat.com
+VITE_RC_ANALYTICS_ENDPOINT=https://api.mimircat.com
 VITE_ASSETS_ENDPOINT=https://da08ctfrofx1b.cloudfront.net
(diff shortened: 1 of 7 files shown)


.env.production
```diff
@@ -1,2 +1,3 @@
 VITE_RC_ENDPOINT=https://api.revenuecat.com
+VITE_RC_ANALYTICS_ENDPOINT=https://api.mimircat.com
```
tonidero, line 2: FWIW, in mobile, we use a different subdomain for these events (`https://api-paywalls.revenuecat.com/` I know it says paywalls but it's used for all events right now)... But it makes more sense to have an entire separate domain 👍

src/main.ts
```diff
@@ -250,7 +250,6 @@ export class Purchases {
     }
     this.eventsTracker = new EventsTracker({
       apiKey: this._API_KEY,
-      httpConfig: httpConfig,
```
tonidero, line 253: I think it makes sense... Not sure if we would want to be able to inject headers in the future... but for now, this looks good!

tonidero (approved):
LGTM!

src/main.ts
polmiro: That is actually one of the questions I had for you. Do we want a configurable override for this? I could add two new fields to the existing `HttpConfig` such as `analyticsProxyUrl`, and `analyticsAdditionalHeaders`.

.env.production
polmiro: Thanks, good to know. I will check with coin.

PR #300, opened 2025-02-04
[WEB-1913] Add tracking for purchase step

polmiro:
## Motivation / Description

Adds tracking events for the payment entry step.


src/behavioural-events/sdk-events.ts
```diff
@@ -64,60 +72,83 @@ export interface CheckoutSessionStartEvent extends IEvent {
...
-  eventName: TrackedEventName.PurchaseSuccessfulDismiss;
+export interface PaymentEntryDismissEvent extends ISDKEvent {
+  eventName: SDKEventName.PaymentEntryDismiss;
+}
+
+export interface PaymentEntryErrorEvent extends ISDKEvent {
+  eventName: SDKEventName.PaymentEntryError;
+  properties: {
+    stripeErrorCode: string | null;
+    stripeErrorMessage: string | null;
```
vicfergar, line 135: Could we rename these fields to `gatewayError*` to make them more generic?
```suggestion
    gatewayErrorCode: string | null;
    gatewayErrorMessage: string | null;
```

polmiro: Yes, good call 👍🏻 !

src/tests/ui/states/state-needs-payment-info.test.ts
```diff
@@ -0,0 +1,113 @@
...
+//     registerAppInfo: vi.fn(),
+//     confirmPayment: vi.fn().mockResolvedValue({
+//       error: null,
+//     }),
+//     confirmSetup: vi.fn().mockResolvedValue({
+//       error: null,
+//     }),
+//   }),
+//   StripeElements: vi.fn(),
+// }));
```
vicfergar, line 48: As part of the Stripe Tax implementation, we're splitting the checkout flow in two steps and the clientSecret will no longer needed to render the payment form.

I believe it will be easier to recover these tests when these changes are deployed. 🤞

vicfergar (approved):
Left some minor suggestions but LGTM!

src/ui/states/state-needs-payment-info.svelte
```diff
@@ -258,6 +288,7 @@
                 },
               }}
               on:change={(event: OnChangeEvent) => {
+                selectedPaymentMethod = event.detail.value.type;
                 isPaymentInfoComplete = event.detail.complete;
```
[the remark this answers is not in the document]
polmiro: Safer indeed 👍🏻

src/tests/ui/states/state-needs-payment-info.test.ts
polmiro: Per our discussion, parking this for the large merge effort. I will add a TODO to the user events branch.

src/behavioural-events/event.ts
```diff
@@ -47,7 +47,7 @@ export class Event {
       type: this.EVENT_TYPE,
       event_name: this.data.eventName,
       app_user_id: this.data.appUserId,
-      context: buildEventContext(),
+      context: buildEventContext() as unknown as EventProperties,
```
nicfix, line 50: Can you move the `as` inside `buildEventContext` and declare it to return `EventProperties`?

In this way we don't need to use `as` in every place we use it.

nicfix (approved):
Left just a small suggestion but looks good to me

PR #302, opened 2025-02-05
User events

polmiro:
## Motivation / Description

Adds a baseline of events for the purchase flow.

## Changes introduced

- Event tracking.
- Events for the SDK initialization.
- Events for the entire checkout flow.
- Internal function to track event.

## TODO

- [x] Manual testing
- [x] ~Use `autoParseUTMParams` helper~ After looking into this I rather keep separate implementation paths, there is little advantage from using it and it is fairly straight-forward code.


polmiro:
@tonidero do you know how can I deal with the `danger` issue? I suspect it's the increase in size?

tonidero:
Yup that seems to be it. We added that as a warning to make sure we don't increase the size of the mobile SDKs repos, specially in iOS, where due to the nature of some package managers, it had to download the whole repo. If we're aware and want to continue, we can just add the label `danger-bypass-size-limit` to the PR, and it should pass on the next run

nicfix:
Maybe we can use another tool then to measure the bundle size so that we only really trigger it for changes that will be downloaded by the customers through npm.

vicfergar (approved):

tonidero:
> Maybe we can use another tool then to measure the bundle size so that we only really trigger it for changes that will be downloaded by the customers through npm.

That would indeed be a nice test to add yeah! Might be worth creating a ticket?

PR #304, opened 2025-02-05
WEB-2065: Adding automatically utm params to metadata

nicfix:
## Motivation / Description
We want to track UTM parameters automatically in the WPL.
However this might be useful also for the sdk itself.

## Changes introduced
* Added a autoParseUTMParams function
* Added tests
* Using the autoParseUTMParams function before executing the purchase.

## Additional comments
This is an implicit behaviour, we could avoid doing this and passing them only explicitly through the metadata (web-1998) but maybe is a good dev experience if they get picked automatically without the need of updating the code but just downloading the latest version of the sdk.

### DIFF
--- a/src/helpers/utm-params.ts
+++ b/src/helpers/utm-params.ts
@@ -1,0 +1,21 @@
+export const autoParseUTMParams = () => {
+  const params = new URLSearchParams(window.location.search);
+
+  const possible_params = [
+    "utm_source",
+    "utm_medium",
+    "utm_campaign",
+    "utm_term",
+    "utm_content",
+  ];
+  const utmParams: { [key: string]: string } = {};
+
+  possible_params.forEach((param) => {
+    const paramValue = params.get(param);
+    if (paramValue !== null) {
+      utmParams[param] = paramValue;
+    }
+  });
+
+  return utmParams;
+};
--- a/src/main.ts
+++ b/src/main.ts
@@ -50,6 +50,7 @@ import { PaywallDefaultContainerZIndex } from "./ui/theme/constants";
 import { parseOfferingIntoVariables } from "./helpers/paywall-variables-helpers";
 import { Translator } from "./ui/localization/translator";
 import { englishLocale } from "./ui/localization/constants";
+import { autoParseUTMParams } from "./helpers/utm-params";
 
 export { ProductType } from "./entities/offerings";
 export type {
@@ -524,6 +525,9 @@ export class Purchases {
 
     const localeToBeUsed = selectedLocale || defaultLocale;
 
+    const utmParamsMetadata = autoParseUTMParams();
+    const metadata = { ...utmParamsMetadata, ...(params.metadata || {}) };
+
     return new Promise((resolve, reject) => {
       mount(RCPurchasesUI, {
         target: certainHTMLTarget,
@@ -556,7 +560,7 @@ export class Purchases {
           purchaseOperationHelper: this.purchaseOperationHelper,
           asModal,
           selectedLocale: localeToBeUsed,
-          metadata: params.metadata,
+          metadata: metadata,
           defaultLocale,
         },
       });
--- a/src/tests/helpers/utm-params.test.ts
+++ b/src/tests/helpers/utm-params.test.ts
@@ -1,0 +1,58 @@
+import { describe, expect, test } from "vitest";
+import { spyOn } from "@vitest/spy";
+import { autoParseUTMParams } from "../../helpers/utm-params";
+
+describe("autoParseUTMParams", () => {
+  test("should return an empty object if no utm param is set", () => {
+    spyOn(URLSearchParams.prototype, "get").mockImplementation(() => null);
+    const utmParams = autoParseUTMParams();
+    expect(utmParams).toEqual({});
+  });
+
+  test("should return all utm params if set", () => {
+    const mockUTMParams: { [key: string]: string } = {
+      utm_source: "source",
+      utm_medium: "medium",
+      utm_campaign: "campaign",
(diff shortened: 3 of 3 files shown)


src/helpers/utm-params.ts
```diff
@@ -0,0 +1,21 @@
...
+    "utm_campaign",
+    "utm_term",
+    "utm_content",
+  ];
+  const utmParams: { [key: string]: string } = {};
+
+  possible_params.forEach((param) => {
+    const paramValue = params.get(param);
+    if (paramValue !== null) {
+      utmParams[param] = paramValue;
```
vicfergar, line 16: I wonder if we should set these params to have a common parent, `utm`, as we do in [the behavioral events](https://github.com/RevenueCat/purchases-js/blob/144adca71a6039d05ac66c5d72d4f5df634814ec/src/behavioural-events/event-context.ts#L16-L22)

tonidero: +1

tonidero (commented):
Looking good! I was thinking whether we would like to have an option to disable this auto-collection... But we can add it later if needed I think

src/helpers/utm-params.ts
nicfix: RE: Common parent

@vicfergar I suspect you mean
```
metadata = {
  'someKey' : 'someValue',
  
  'utm': {
     'utm_source':'...',
     'utm_...':'...'
  }
}
```

I wouldn't add that nesting, considering that this will reach 2 places:
1. Stripe metadata (flat)
2. Our metadata/events

Adding a nested object would complicate all the rest down the line and I don't think it's worth it.

nicfix: RE: Sharing the code

I agree @polmiro, I've seen that you already reused the function in one of your PRs.
I'll keep you posted if I merge first.

src/main.ts
```diff
@@ -556,7 +560,7 @@ export class Purchases {
           purchaseOperationHelper: this.purchaseOperationHelper,
           asModal,
           selectedLocale: localeToBeUsed,
-          metadata: params.metadata,
+          metadata: metadata,
```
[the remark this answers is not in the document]
nicfix: Sure thing!
On it.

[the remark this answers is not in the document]
nicfix: Added it in the demo app tests

nicfix:
@tonidero I added a flags object to opt out from automatically sending UTM params

src/entities/functionality-config.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * Flags used to enable or disable certain features in the sdk.
+ * @public
+ */
+export interface FlagsConfig {
```
tonidero, line 5: Hmm I'm not sure if I like the name... I guess it's too generic 😅... If we want to follow more the mobile SDKs pattern, I would say we should change the `configure` method to accept an object (something like `PurchasesConfiguration`), which can have all the required and optional parameters, then make `collectAutomaticallyUTMParamsAsMetadata` one of those top-level parameters with a default value. I know this would increase the number of parameters in that `PurchasesConfiguration` object, but it would avoid this very generic level of indirection. Wdyt?

Note: If we end up with that route, we should keep the existing configure function but deprecate it, so we don't make a breaking change.

src/helpers/utm-params.ts
vicfergar: Oh, I noticed that the `purchase` endpoint does not accept nested attributes, right?

src/entities/functionality-config.ts
nicfix: I don't have a strong opinion.
I'd also like to change towards a `configure` that has direct named parameters, however I think is too big for the context of this PR.

Considering that in any case we would need to deprecate this method I think it's fine to keep it as I proposed and make that change as a follow-up PR.
I think having the possibility to opt out is important for this feature.

nicfix: PS: I just renamed the file to keep it consistent with the class name

src/helpers/utm-params.ts
nicfix: Exactly, they should be flat so that they can be sent along to Stripe as metadata

src/entities/functionality-config.ts
tonidero: Hmm still feels like a weird API to me TBH... but as long as we are planning to move to a configuration with an object in the future, and not expand on this, I wouldn't be too opposed 👍

src/entities/flags-config.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * Flags used to enable or disable certain features in the sdk.
+ * @public
+ */
+export interface FlagsConfig {
+  /**
+   * If set to true, the SDK will automatically collect UTM parameters and store them as at the time of purchase.
+   * @defaultValue true
+   */
+  collectAutomaticallyUTMParamsAsMetadata?: boolean;
```
tonidero, line 10: Nitpick but maybe rename to `collectUTMParamsAsMetadataAutomatically`? Sounds a bit better to me 😅. Not a blocker though.

tonidero (approved):
Still not a fan of the naming but I don't think it's a ship stopper so :shipit:

src/entities/flags-config.ts
antoniobg: Or `autocollectUtmParamsAsMetadata`?

antoniobg: or simply `autoCollectUTMAsMetadata`

nicfix: Renamed to `autoCollectUTMAsMetadata`

PR #305, opened 2025-02-05
[WEB-2010] Split the checkout flow into two parts

vicfergar:


.storybook/main.ts
```diff
@@ -19,8 +19,6 @@ const config: StorybookConfig = {
     ...config,
     VITE_STORYBOOK_PUBLISHABLE_API_KEY:
       process.env.VITE_STORYBOOK_PUBLISHABLE_API_KEY || "",
-    VITE_STORYBOOK_RESTRICTED_SECRET:
-      process.env.VITE_STORYBOOK_RESTRICTED_SECRET || "",
```
vicfergar, line 23: This env var is no longer needed. With the new flow, we don't need a payment_intent or setup_intent to render the payment form 😄

package.json
```diff
@@ -92,7 +92,6 @@
     "storybook": "^8.5.0",
     "svelte": "^5.12.0",
     "svelte-check": "^4.1.1",
-    "svelte-stripe": "^1.3.0",
```
vicfergar, line 95: I removed this third-party library and initialized the stripe elements as [recommended by Stripe](https://docs.stripe.com/payments/payment-element#create-a-payment-element).

src/networking/backend.ts
```diff
@@ -72,17 +74,17 @@ export class Backend {
...
-  async postPurchase(
+  async postCheckoutStart(
     appUserId: string,
     productId: string,
-    email: string,
     presentedOfferingContext: PresentedOfferingContext,
     purchaseOption: PurchaseOption,
-  ): Promise<PurchaseResponse> {
-    type PurchaseRequestBody = {
+    email?: string,
```
vicfergar, line 82: `email` field is no longer required to initialize a purchase. This will allow us to move that field to the `state-needs-payment-info` and remove the `state-needs-auth-info`

src/tests/helpers/purchase-operation-helper.test.ts
```diff
@@ -225,41 +307,67 @@ describe("PurchaseOperationHelper", () => {
...
-  //   );
-  //   await expectPromiseToError(
-  //     purchaseOperationHelper.pollCurrentPurchaseForCompletion(),
-  //     new PurchasesError(
-  //       ErrorCode.UnknownError,
-  //       "Purchase status was not finished in given timeframe",
-  //     ),
-  //   );
-  // });
+  test("pollCurrentPurchaseForCompletion fails if poll returns in progress more times than retries", async () => {
```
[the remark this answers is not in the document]
polmiro: image

src/networking/backend.ts
polmiro: merge-conflict incoming!

src/ui/rcb-ui.svelte
```diff
@@ -102,25 +102,16 @@
...
-          lastError = null;
-          currentView = "needs-payment-info";
-          paymentInfoCollectionMetadata = result;
-          return;
-        }
-        if (result.next_action === "completed") {
-          lastError = null;
-          currentView = "success";
-          return;
-        }
```
vicfergar, line 123: This was unreachable code, since `next_action` is always set to `collect_payment_info` by the backend

src/ui/states/state-needs-payment-info.svelte
```diff
@@ -197,21 +111,52 @@
...
+    }
+
+    elements = stripe.elements({
+      loader: "always",
+      locale: localeToUse,
+      mode: "payment",
+      paymentMethodTypes: ["card"],
+      setup_future_usage: "off_session",
+      amount: productDetails.currentPrice.amountMicros / 10000,
+      currency: productDetails.currentPrice.currency.toLowerCase(),
```
vicfergar, line 155: Since these parameters should match the settings from the PaymentIntent/SetupIntent, I was considering including them in the `CheckoutStart` response. Wdyt @RevenueCat/web?

src/ui/states/state-needs-payment-info.svelte
```diff
@@ -275,71 +220,155 @@
...
+      } else {
+        onContinue();
+      }
+    }
+  };
+
+  const completeCheckout = async (stripe: Stripe, customerEmail: string) => {
+    if (!clientSecret) {
+      const checkoutCompleteResponse =
+        await purchaseOperationHelper.checkoutComplete(customerEmail);
```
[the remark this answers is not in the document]
vicfergar: That's correct, that's why I added the `if(!clientSecret)` safeguard.

Ideally, I would like to pass a `confirmationTokenId` to this call so that the `stripe.confirm*` call could be handled in the backend. Stripe docs [here](https://docs.stripe.com/payments/payment-element/migration-ct)

However, to create confirmation tokens, we need to add a new permission to our Stripe App. I'll open a thread in the channel to discuss about this approach

src/ui/states/state-needs-payment-info.svelte
vicfergar: I changed this to retrieve these values from the start purchase response so that we have more control over them from the backend

src/networking/endpoints.ts
```diff
@@ -81,6 +81,28 @@ export class GetBrandingInfoEndpoint implements Endpoint {
...
 
+export class CheckoutStartEndpoint implements Endpoint {
+  method: HttpMethodType = "POST";
+  name: string = "checkoutStart";
+  urlPath(): string {
+    return `${RC_BILLING_PATH}/checkout/start`;
+  }
+}
+
+export class CheckoutCompleteEndpoint implements Endpoint {
```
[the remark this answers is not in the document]
vicfergar: Good call. Renamed!

src/ui/rcb-ui.svelte
vicfergar: That's correct Nicola, the old endpoints (`/subscribe` and `/purchase` endpoints) are kept untouched and will return the field `collect_payment_info`.

I tested all of this in a canary using the published version of the SDK before shipping the BE changes.

nicfix: My question was related to the `completed` value.

When was that value returned?
In what scenario?

I understand that "recently" we updated to always return `collect_payment_info` my question is:

Did we break older version of the sdk that supported `completed` when we stopped sending it?

src/ui/states/state-needs-payment-info.svelte
```diff
@@ -90,256 +130,145 @@
...
+      );
+
+      paymentElement.mount("#payment-element");
+      paymentElement.on("change", (event) => {
+        isPaymentInfoComplete = event.complete;
+      });
+    } catch (error) {
+      console.error("Failed to initialize Stripe:", error);
+      modalErrorMessage = "Failed to initialize payment system";
     }
```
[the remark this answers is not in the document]
vicfergar: Thanks for the heads-up! I believe it should be resolved now.

polmiro (approved):

src/ui/rcb-ui.svelte
nicfix: As discussed in the 1:1 I was more worried about the potential issue happening due to the backend changes more than this change. Sorry for the confusion.

src/tests/test-helpers.ts
```diff
@@ -1,9 +1,7 @@
-import { expect } from "vitest";
+import { expect, assert } from "vitest";
 import { PurchasesError } from "../entities/errors";
 
-export function failTest() {
-  expect(true).toBeFalsy();
-}
+export function failTest() {}
```
alfondotnet, line 4: remove this?

src/ui/states/state-needs-payment-info.svelte
```diff
@@ -89,257 +122,170 @@
...
+
+        paymentElement.mount("#payment-element");
+        paymentElement.on(
+          "change",
+          (event: StripePaymentElementChangeEvent) => {
+            isPaymentInfoComplete = event.complete;
+          },
+        );
+      } catch (error) {
+        console.error("Failed to initialize Stripe:", error);
```
alfondotnet, line 167: Logger.error

src/ui/states/state-needs-payment-info.svelte
```diff
@@ -89,257 +122,170 @@
...
+        modalErrorMessage = "Failed to initialize payment system";
```
alfondotnet, line 168: we might not want to surface this error to the client, and in fact we should redirect to the error view or something

alfondotnet (approved):

alfondotnet (approved):

src/ui/states/state-needs-payment-info.svelte
vicfergar: Addressed

src/ui/states/state-needs-payment-info.svelte
vicfergar: Fixed. A non-recoverable generic error will be surfaced.

antoniobg (approved):

src/ui/states/state-needs-payment-info.svelte
```diff
@@ -89,257 +124,191 @@
...
 
-    stripe = await loadStripe(stripePk, {
-      stripeAccount: stripeAcctId,
-    });
+    return () => {
+      window.removeEventListener("resize", onResize);
+    };
+  });
+
+  onMount(() => {
```
tonidero, line 137: We have another `onMount` above, does it make sense to consolidate them?

tonidero (approved):
Big change! But it all makes sense  👍

src/ui/states/state-needs-payment-info.svelte
vicfergar: I think it's preferable to keep them separated

PR #307, opened 2025-02-07
Calculate button states

alfondotnet:
## Motivation / Description

https://www.loom.com/share/1e853343aeed429b9c54b686f70a1797

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
     "preview": "vite preview",
     "lint": "prettier --check .",
     "format": "prettier --write .",
+    "pack-build": "npm run build && npm pack",
     "test": "vitest",
     "test:typeCheck": "tsc --project tsconfig.test.json",
     "typecheck": "tsc --noEmit",
--- a/src/stories/components.stories.svelte
+++ b/src/stories/components.stories.svelte
@@ -79,27 +79,6 @@
   parameters={{ viewport: { defaultViewport: "mobile" } }}
 />
 
-<Story
-  name="Button (Secondary)"
-  args={{
-    component: ButtonComponent,
-    children: "Click me",
-    intent: "secondary",
-  }}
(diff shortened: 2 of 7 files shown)


polmiro (approved):

package.json
```diff
@@ -43,6 +43,7 @@
     "preview": "vite preview",
     "lint": "prettier --check .",
     "format": "prettier --write .",
+    "pack-build": "npm run build && npm pack",
```
vicfergar, line 46: I’m not opposed, but I was wondering why you added this script. Are you using it during development? If so, you might find [this section of the README](https://github.com/RevenueCat/purchases-js?tab=readme-ov-file#install-the-library-in-a-local-project) helpful! 😊

vicfergar (approved):

src/stories/components.stories.svelte
```diff
@@ -0,0 +1,110 @@
...
+<Story
+  name="Button (Secondary disabled)"
+  args={{
+    component: ButtonComponent,
+    children: "Click me",
+    intent: "secondary",
+    disabled: true,
+  }}
+  parameters={{ viewport: { defaultViewport: "mobile" } }}
+/>
```
vicfergar, line 101: Should we remove these two stories?

alfondotnet:
Removed the type secondary from intent so it should not be able to be
passed until implemented, will remove the story

Alfonso Embid-Desmet
Staff Engineer 💻 @ revenuecat.com

src/stories/components.stories.svelte
alfondotnet: ups, hadn't pushed a commit, sorry, just did

package.json
alfondotnet: packing is what I use to embed the library on the WPL PRs. I figured there was no harm to leave it there

PR #309, opened 2025-02-07
preload stripe on parent component

alfondotnet:
Preload stripe elements on parent component

@vicfergar can you test this against your changes to see if it improves the jankyness? I am not sure it does


vicfergar:
I tested this branch locally, also with my changes, but I didn't notice any improvement. In my branch, I extracted the Stripe logic into a dedicated service. Maybe we should preload not only Stripe but also the elements 🤔 .

I think we could give it a try once my PR is merged, wdyt?

alfondotnet:
yea no rush

PR #312, opened 2025-02-12
[WEB-2082] Append `trace_Id` and `checkout_session_id` to returned redeemUrl

polmiro:
## Motivation / Description

Given that we prefer to track the redemptions on the backend side, the `redeem_url` will now include `trace_id` and `checkout_session_id` when the url is a web one `http` or `https`. The future links will point to Khepri and redirect the user to the custom scheme.

They could also be append them to mobile links with custom schemes but given there is no expectation that they will make use of them I decided to skip them. @tonidero let me know if you prefer me to add them for those too.

### DIFF
--- a/src/tests/purchase.test.ts
+++ b/src/tests/purchase.test.ts
@@ -1,8 +1,22 @@
-import { describe, expect, test } from "vitest";
+import { beforeEach, describe, expect, test, vi } from "vitest";
 import { configurePurchases } from "./base.purchases_test";
 import { APIGetRequest, type GetRequest } from "./test-responses";
+import { mount } from "svelte";
+import { Purchases } from "../main";
+
+vi.mock("svelte", () => ({
+  mount: vi.fn(),
+}));
+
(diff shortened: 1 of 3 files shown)


src/tests/purchase.test.ts
```diff
@@ -22,6 +36,89 @@ describe("purchase", () => {
...
+
+    expect(redeemUrl.searchParams.get("trace_id")).toBe(
+      "c1365463-ce59-4b83-b61b-ef0d883e9047",
+    );
+    expect(redeemUrl.searchParams.get("checkout_session_id")).toBe(
+      "c1365463-ce59-4b83-b61b-ef0d883e9047",
+    );
+  });
+
+  test("does NOT append params to other protocols", async () => {
```
tonidero, line 96: 👍 I think this makes sense, no need to send it if it's going to be unused... If we decide we want to track it from the mobile SDKs we can think about adding it for these as well.

tonidero (approved):
LGTM!

PR #315, opened 2025-02-12
Some UI fixes

alfondotnet:
Some bug bash fixes:

- The arrow to "expand" should change when luminosity does
- Border of email input should be 1px 
- Fixes how a non monthly product shows and one with trial in present offer

I am waiting for design update of how to present the trial information

### DIFF
--- a/src/ui/badge.svelte
+++ b/src/ui/badge.svelte
@@ -1,0 +1,37 @@
+<script lang="ts">
+  import { type Snippet } from "svelte";
+
+  export let text: string | Snippet;
+</script>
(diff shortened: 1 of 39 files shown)


src/ui/badge.svelte
```diff
@@ -0,0 +1,37 @@
+<script lang="ts">
```
alfondotnet, line 1: I realized today that the designs on this part were outdated in Figma :( and the new designs are not ready yet.. so I'll keep this as it is until we can use or remove

PR #318, opened 2025-02-14
Remove checkout_session_id

polmiro:
## Motivation / Description

There is little value and some confusion added recording `checkout_session_id`. Given that every time a purchase flow is started an event is recorded `checkout_session_start` and another one when a purchase flow is ended `checkout_session_end` one can already derive the information given by `checkout_session_id` from the order of events.

## Changes introduced

- Completely remove `checkout_session_id`.


antoniobg (approved):

tonidero (approved):
I don't have much context on the underlying event processing, but code changes look good!

nicfix (approved):
Makes sense to remove the checkout_session_id, in the end we have the track_id that allows us to group all the events in a single session for each refresh.

PR #322, opened 2025-02-14
[Not merging yet] [WEB-2120] Put email view alongside payment element

alfondotnet:
This PRs:

- Puts the e-mail field alongside payment component
- Vertically centers the "continue button" on the success / failure views


src/helpers/purchase-operation-helper.ts
```diff
@@ -129,7 +129,6 @@ export class PurchaseOperationHelper {
     productId: string,
     purchaseOption: PurchaseOption,
     presentedOfferingContext: PresentedOfferingContext,
-    email?: string,
```
alfondotnet, line 132: @vicfergar to review

src/ui/rcb-ui.svelte
```diff
@@ -97,20 +93,12 @@
       currentView = "loading";
     }
 
-    if (!customerEmail) {
```
alfondotnet, line 100: @vicfergar to review

src/helpers/purchase-operation-helper.ts
vicfergar: It’s fine to remove this parameter if we want to maintain a simpler interface.

src/ui/rcb-ui.svelte
vicfergar: That's correct

src/ui/states/state-needs-payment-info.svelte
```diff
@@ -158,23 +161,41 @@
...
-            "Failed to load payment form",
-            event.error instanceof Error
-              ? event.error.message
-              : String(event.error),
-          );
-          handlePaymentError(purchaseError);
+          triggerError(event.error);
         });
       } catch (error) {
         const purchaseError = new PurchaseFlowError(
```
vicfergar, line 201: We could also call `triggerError(error)`, as there is no need for a different error message.

src/ui/states/state-needs-payment-info.svelte
```diff
@@ -158,23 +161,41 @@
           brandingInfo?.app_name,
         );
 
+        const linkAuthenticationElement = StripeService.createEmailElement(
+          elements,
+          customerEmail,
+        );
+
+        const triggerError = (error: StripeError) => {
```
vicfergar, line 169:
```suggestion
        const triggerLoadingError = (error: StripeError) => {
```

nicfix (changes requested):
I think we should wait a minute before jumping on this.

We didn't clarify how to handle the errors whenever there's the email's server validation and/or what should we show when is the developer providing the email.

Let's wait a minute and finalize what we already have in the new design branch and then we can come back to this.

nicfix:
Closing this PR for now (not deleting the branch)

PR #325, opened 2025-02-18
First proposal to accept branding appearance override

nicfix:

### DIFF
--- a/src/entities/branding.ts
+++ b/src/entities/branding.ts
@@ -1,0 +1,16 @@
+/**
+ * @public
+ * `BrandingAppearance` defines the appearance settings
+ *  of an app's branding configuration.
+ */
+export interface BrandingAppearance {
+  color_buttons_primary: string;
+  color_accent: string;
+  color_error: string;
+  color_product_info_bg: string;
+  color_form_bg: string;
+  color_page_bg: string;
+  font: string;
+  shapes: "default" | "rectangle" | "rounded" | "pill";
+  show_product_description: boolean;
+}
--- a/src/entities/purchase-params.ts
+++ b/src/entities/purchase-params.ts
@@ -50,7 +50,6 @@ export interface PurchaseParams {
    * using this method might break in future releases of `purchases-js`.
    *
    * @internal
-   * @type {BrandingAppearance | undefined}
    */
   brandingAppearanceOverride?: BrandingAppearance;
 }
--- a/src/entities/purchase-params.ts
+++ b/src/entities/purchase-params.ts
@@ -1,5 +1,6 @@
 import type { Package, PurchaseMetadata, PurchaseOption } from "./offerings";
-import type { BrandingAppearance } from "../networking/responses/branding-response";
+
+import type { BrandingAppearance } from "./branding";
 
 /**
  * Parameters used to customise the purchase flow when invoking the `.purchase` method.
--- a/src/entities/purchase-params.ts
+++ b/src/entities/purchase-params.ts
@@ -44,10 +44,12 @@ export interface PurchaseParams {
   /**
    * Defines an optional override for the default branding appearance.
    *
-   * This property can be used to customize the visual presentation or styling
-   * associated with the branding, replacing the default settings. When set, it
-   * supersedes the pre-configured appearance defined elsewhere in the system.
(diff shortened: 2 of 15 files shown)


src/entities/purchase-params.ts
```diff
@@ -39,4 +40,15 @@ export interface PurchaseParams {
...
   metadata?: PurchaseMetadata;
+
+  /**
+   * Defines an optional override for the default branding appearance.
+   *
+   * This property can be used to customize the visual presentation or styling
+   * associated with the branding, replacing the default settings. When set, it
+   * supersedes the pre-configured appearance defined elsewhere in the system.
+   *
+   * @type {BrandingAppearance | undefined}
```
tonidero, line 51: I don't think you need to specify the type like this?

src/networking/responses/branding-response.ts
```diff
@@ -13,8 +13,17 @@ export type BrandingAppearance = {
 export type BrandingInfoResponse = {
   app_icon: string | null;
   app_icon_webp: string | null;
-  appearance: BrandingAppearance;
+  appearance: BrandingAppearance | undefined;
```
tonidero, line 16: Hmm note that if we make this nullable, it could be a breaking change for existing SDK versions. The backend should keep sending this data for existing SDKs I think.

src/entities/purchase-params.ts
```diff
@@ -39,4 +40,15 @@ export interface PurchaseParams {
...
+   */
+  brandingAppearanceOverride?: BrandingAppearance;
```
tonidero, line 53: Hmm I'm not familiar with the need for overriding this in the SDK itself... I wonder if we should mark this as `@internal` if possible, unless we think it's going to be used by developers.

tonidero: Ahhh I just saw the loom recording 😅 . Makes more sense now. I do wonder if we shouldn't extract the UI part into another library that can be used by the SDK + the dashboard actually though... That would avoid us having to expose this API.

tonidero: That would probably be a bigger change though and would make it a bit more difficult to iterate on the SDK... So I'm not totally opposed with going with this option for now I think.

src/entities/purchase-params.ts
nicfix: I'll remove it!

nicfix: Done

src/entities/purchase-params.ts
```diff
@@ -39,4 +40,16 @@ export interface PurchaseParams {
...
+   * Defines an optional override for the default branding appearance.
+   *
+   * This property is used internally at RevenueCat to handle dynamic themes such
+   * as the ones coming from the Web Paywall Links. We suggest to use the Dashboard
+   * configuration to set up the appearance since a configuration passed as parameter
+   * using this method might break in future releases of `purchases-js`.
+   *
+   * @internal
+   */
+  brandingAppearanceOverride?: BrandingAppearance;
```
tonidero, line 54: Do we need to update the API and export `BrandingAppearance`? We should also mark it as `@internal` in that case.

tonidero (commented):
Just a comment about the API but looking good!

src/entities/purchase-params.ts
tonidero: I'm actually surprised our tests didn't catch the change in API actually... Not sure what happened..

nicfix: Weird, updating it.

Maybe is because we are not merging in main

tonidero (commented):
Just a couple more comments about the API files... Code changes look good though.

tonidero (approved):
A couple more questions but looks good to go IMO!

src/entities/branding.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * @public
```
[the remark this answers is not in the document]
nicfix: Marking this as internal was causing building issues, however adding nothing was requesting me to set at least something (internal public beta etc).

src/entities/purchase-params.ts
```diff
@@ -39,4 +41,16 @@ export interface PurchaseParams {
...
```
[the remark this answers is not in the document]
nicfix: I guess so.

PR #324, opened 2025-02-18
Re-added support to rendering the sdk in an element passed as parameter

nicfix:

PR #326, opened 2025-02-19
[WEB-2140] restore scroll after closing modal + fixing height

alfondotnet:
This PR:

- Fixes issue when closing modal. `onDestroy` is not being called so scroll is not returned to the app
- Fixes issue introduced by https://github.com/RevenueCat/purchases-js/pull/324/files#diff-890948e5715657ae559e6221fa767a9c6c771e193ddc37fc42add8c8ab491e6eL23 whereby the modal won't take the full screen when having scrolled down (since it's absolute 100%/100%). It fixes the issue by returning conditionally the previous styles, based on `isInElement`

### DIFF
--- a/src/ui/layout/container.svelte
+++ b/src/ui/layout/container.svelte
@@ -8,10 +8,17 @@
   let spacingStyle = new Theme(brandingAppearance).spacingStyleVars;
   let style = [textStyle, spacingStyle].join("; ");
 
+  export let isInElement: boolean = false;
+
   export let children: Snippet;
 </script>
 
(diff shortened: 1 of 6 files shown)


src/ui/layout/container.svelte
```diff
@@ -46,4 +48,18 @@
       sans-serif;
     overflow: auto;
   }
+
+  .rcb-ui-container.fullscreen {
```
alfondotnet, line 52: returns the previous styles conditionally if `fullscreen` class is defined

nicfix (approved):

PR #341, opened 2025-02-25
[WEB-2248] wpl-product-selection-prices-should-not-show-us

alfondotnet:
We can't use `currencyDisplay: narrowSymbol` as that introduces too much ambiguity in the cases of the Canadian dollar, Mexican dollar, yuan vs yen.. etc

We also do some DRY


nicfix:
This was merged with:
https://github.com/RevenueCat/purchases-js/pull/342

PR #342, opened 2025-02-25
Fix checkout min height + dedicated label for the "per {period}"

nicfix:
Created dedicated labels for the 'per {{period}}'


src/helpers/string-helpers.ts
```diff
@@ -0,0 +1,7 @@
+export const toUpperFirst = (value: string) => {
```
antoniobg, line 1: Maybe ?
```suggestion
export const capitalize = (value: string) => {
```

nicfix: Renamed

antoniobg (approved):

PR #343, opened 2025-02-26
Add `prepaid` entitlement period support

tonidero:
## Motivation / Description
This was added in the mobile SDKs, we should also support it in the `web` SDK since it can receive entitlements from Google.

### DIFF
--- a/src/entities/customer-info.ts
+++ b/src/entities/customer-info.ts
@@ -25,9 +25,10 @@ export type Store =
  * - "normal" If the entitlement is not under an introductory or trial period.
  * - "intro" If the entitlement is under an introductory period.
  * - "trial" If the entitlement is under a trial period.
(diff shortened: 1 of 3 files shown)


src/entities/customer-info.ts
```diff
@@ -25,9 +25,10 @@ export type Store =
  * - "normal" If the entitlement is not under an introductory or trial period.
  * - "intro" If the entitlement is under an introductory period.
  * - "trial" If the entitlement is under a trial period.
+ * - "prepaid" If the entitlement is a prepaid entitlement. Only for Google Play subscriptions.
  * @public
  */
-export type PeriodType = "normal" | "intro" | "trial";
+export type PeriodType = "normal" | "intro" | "trial" | "prepaid";
```
tonidero, line 31: These discreet types are probably not ideal since this could also be considered a breaking change in the API...

antoniobg (approved):

PR #344, opened 2025-02-26
WEB-2106: Using 16px input fontSize to avoid ios phones zooming in

nicfix:


nicfix:
@alfondotnet please feel free to merge it if it works on your ios device as well

PR #354, opened 2025-03-03
[Draft] [WEB-2232] add subscription renewal date and customer email to purchase result

alfondotnet:
We need this info in the purchase result to display in the WPL


src/helpers/purchase-result-helpers.ts
```diff
@@ -0,0 +1,25 @@
...
+export function getRenewalDateFromPackage(
+  customerInfo: CustomerInfo,
+  rcPackage: Package,
+): Date | null {
+  const productIdentifier = rcPackage.webBillingProduct.identifier;
+
+  const isActiveSubscription =
+    customerInfo.activeSubscriptions.has(productIdentifier);
+
+  if (!isActiveSubscription) {
```
alfondotnet, line 13: would be cool to actually refactor `getWillRenew` to ensure that this is a purchase that renews although for this use case the function will be called right after purchase so not sure if worth it, thoughts @tonidero ?

src/entities/purchase-result.ts
```diff
@@ -19,4 +19,15 @@ export interface PurchaseResult {
...
+  /**
+   * The email of the customer who made the purchase.
+   */
+  readonly customerEmail: string | null;
+
+  /**
+   * The date when the subscription will renew.
+   * This will be null for non-subscription purchases or lifetime subscriptions.
+   */
+  readonly renewalDate: Date | null;
```
tonidero, line 32: Wouldn't this be accessible by the `customerInfo` property already? While it's true that you would need to know the package/product just purchased, the client should be aware of that already right? Just thinking it seems like it's duplicating information...

src/helpers/purchase-result-helpers.ts
```diff
@@ -0,0 +1,25 @@
+import type { CustomerInfo } from "../entities/customer-info";
+import type { Package } from "../entities/offerings";
+
+export function getRenewalDateFromPackage(
```
tonidero, line 4: Hmm I think this could directly live in the WPL, unless we think this is something other users of the SDK will need... Which they can still access through the CustomerInfo as I mentioned.

src/helpers/purchase-result-helpers.ts
tonidero: Hmm not sure I understand what you mean... `getWillRenew` already checks for the expiration date but that of an entitlement, instead of a product directly: https://github.com/RevenueCat/purchases-js/blob/bd079f5ac814e86bf163b35373cc3c99d05ee8f5/src/entities/customer-info.ts#L325 . Note that there can be uncommon situations where more than one subscription gives the same entitlement, since it's an N to M relationship).

Entitlements are not recommended to be used with consumables, since the model doesn't really match though, so this is mostly only checked for subscriptions + non consumables.

Lmk if I misunderstood this though!

src/entities/purchase-result.ts
```diff
@@ -19,4 +19,15 @@ export interface PurchaseResult {
    * The operation session id of the purchase.
    */
   readonly operationSessionId: string;
+
+  /**
+   * The email of the customer who made the purchase.
+   */
+  readonly customerEmail: string | null;
```
tonidero, line 26: I think this makes sense... I was thinking whether there would be any PII concern, but from what I can think of, any concerns would already happen between introducing the email and the payment details. Also, I think this could be useful for devs in some situations yeah.

tonidero (commented):
Left some comments. Mostly I wonder whether we could get away with moving `renewalDate` to the WPL and avoid duplicating the information there. Lmk what you think!

PR #356, opened 2025-03-03
Make the size of the three dots in continue button smaller

alfondotnet:


.storybook/main.ts
```diff
@@ -6,6 +6,7 @@ const config: StorybookConfig = {
     "@storybook/addon-links",
     "@storybook/addon-essentials",
     "@storybook/addon-interactions",
+    "@storybook/addon-controls",
```
alfondotnet, line 9: useful for when typing args have configurable controls

polmiro (approved):

src/ui/processing-animation.svelte
```diff
@@ -1,8 +1,22 @@
...
+<div
+  class="processing"
+  style="--shadow-offset: {sizeMap[size].offset}; --width: {sizeMap[size]
+    .width};"
+></div>
 
 <style>
   .processing {
-    width: 12px;
+    width: var(--width, 12px);
```
nicfix, line 19: Mmmm, IIRC css variable names are not local, this could be set by someone outside this component. Also --width is quite generic so it's easy that it gets set.

Why do we need a var for this?

alfondotnet: > IIRC css variable names are not local
the CSS variable is being defined in the `style=` property, so they won't be accessible to the parents of this element, so no issue

> this could be set by someone outside this component.

In that case, it would be overriden by the `style` property, so no issue

>  Also --width is quite generic so it's easy that it gets set.

See above

> Why do we need a var for this?

To calculate, based on the size prop, both the widths and offsets of the dots with respect to each other dynamically, otherwise i'd have to create duplicate CSS classes and it would be much less maintainable imo.

nicfix (approved):

PR #358, opened 2025-03-05
Fix modal scroll on iOS

alfondotnet:


src/main.ts
```diff
@@ -614,6 +614,45 @@ export class Purchases {
...
+      };
+
+      const onError = (e: PurchaseFlowError) => {
+        const event = createCheckoutSessionEndErroredEvent({
+          errorCode: e.errorCode?.toString(),
+          errorMessage: e.message,
+        });
+        this.eventsTracker.trackSDKEvent(event);
+
+        if (component) {
```
alfondotnet, line 648: These two (onError and onFinished) are extracted to a function so we can call `unmount(component)` like `onClose`, so that we make sure to return the scroll to the caller in these paths also.

src/ui/layout/layout.svelte
```diff
@@ -17,6 +17,8 @@
     height: 100%;
     container-type: size;
     container-name: layout-query-container;
+    overflow-y: auto;
```
alfondotnet, line 20: overflow-y: auto needs to be set in here and in the child as well so that WebKit iOS does not clip the div to the parent height, and actually respects the overflow: auto of the parent (`rcb-ui-container.fullscreen`, the one with `position: fixed`)

src/ui/layout/layout.svelte
```diff
@@ -17,6 +17,8 @@
...
+    overscroll-behavior: none;
```
alfondotnet, line 21: this prevents iOS from overscrolling and showing the contents of the page underneath when you overshoot

src/ui/rcb-ui.svelte
```diff
@@ -66,13 +67,17 @@
 
   onMount(() => {
     if (!isInElement) {
-      document.body.style.overflow = "hidden"; // Prevents background scrolling
+      document.documentElement.style.height = "100%";
```
alfondotnet, line 70: trick to prevent ever having two scrolls (once you finish scrolling on the modal), if the caller page has to scroll also, goes along with overflow: hidden

src/ui/sandbox-banner.svelte
```diff
@@ -26,7 +26,7 @@
 
 <style>
   .rcb-ui-sandbox-banner {
-    position: absolute;
+    position: sticky;
```
alfondotnet, line 29: looks better to have the sandbox bar sticky

src/ui/rcb-ui.svelte
```diff
@@ -66,13 +67,17 @@
...
+      document.body.style.height = "100%";
```
nicfix, line 71: Modifying these makes me itchy, feels very invasive towards the hosting application and we risk that we cannot set them back.

Is there no other way to avoid changing the document?

nicfix (approved):
Left one comment, I think that if we can find a way to avoid modifying the document it would be better. But it should be good enough for now.

src/ui/rcb-ui.svelte
alfondotnet: Not really. 

The better solution would be to use a MIT dependency that does `position: fixed` saving the scrollTop of the page on iOS or overflow: hidden if not iOS, I have read:

https://github.com/willmcpo/body-scroll-lock

Although this dependency has problems since they shipped a bug in the latest version that doesn't work well, it seems folks are using now https://github.com/tuax/tua-body-scroll-lock for this kind of thing, but I figured we keep if simple. 

At the end, you gonna have to change something on the body, and in the case of `onDestroy()` not being called (hopefully we cover all cases with this PR, at least for our purchase flows, I didn't touch paywalls), you can always refresh the page worst case, but yeah, this is a known pitfall of position fixed modals

PR #359, opened 2025-03-05
Added the collectAnalyticsEvents parameter

nicfix:
Added the silent mode for the EventsTracker

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "@revenuecat/purchases-js",
   "description": "Web subscriptions made easy. Powered by RevenueCat",
   "private": false,
-  "version": "0.18.0",
+  "version": "0.18.1",
   "type": "module",
   "files": [
     "dist"
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "@revenuecat/purchases-js",
   "description": "Web subscriptions made easy. Powered by RevenueCat",
   "private": false,
-  "version": "0.18.1",
+  "version": "0.18.0",
   "type": "module",
   "files": [
     "dist"
--- a/src/behavioural-events/events-tracker.ts
+++ b/src/behavioural-events/events-tracker.ts
@@ -78,6 +78,10 @@ export default class EventsTracker implements IEventsTracker {
   }
 
   private trackEvent(props: TrackEventProps) {
+    if (this.isSilent) {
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@revenuecat/purchases-js",
   "description": "Web subscriptions made easy. Powered by RevenueCat",
   "private": false,
-  "version": "0.18.0",
+  "version": "0.18.1",
```
tonidero, line 5: NABD but this is outdated since we're now in 0.18.2 (probably outdated in the base branch)

src/behavioural-events/events-tracker.ts
```diff
@@ -96,6 +104,12 @@ export default class EventsTracker implements IEventsTracker {
 
     const eventsToFlush = [...this.eventsQueue];
 
+    if (this.isSilent) {
+      Logger.debugLog("Skipping events tracking, the EventsTracker is silent");
+      this.eventsQueue.splice(0, eventsToFlush.length);
```
tonidero, line 109: Hmm would it make more sense to not add the events to the queue in the first place instead? If we want to optimize it, we could even just not create the `EventTracker` in the first place, and make it optional inside `Purchases`.

This is considered this setting can't be changed after configuration.

package.json
nicfix: Thanks, we'll need to bump to 1.0.0 on fon/new-sdk-views, I'll roll it back to `0.18.0` and then we can bump in the base branch

src/behavioural-events/events-tracker.ts
tonidero: I was taking another look, and my idea to not create the EventTracker in the first place is trickier, since this class also generates the `traceId` which gets sent to the backend when starting a checkout session. I think it might be better to leave that, but lmk if we would prefer to stop sending that property as well

src/tests/behavioral-events/events-tracker.test.ts
```diff
@@ -43,6 +43,27 @@ describe("EventsTracker", (test) => {
     context.eventsTracker.dispose();
   });
 
+  test("does not track even if silent", async () => {
```
polmiro, line 46:
```suggestion
  test("does not track event if silent", async () => {
```

polmiro (approved):
Looks good to me, only things:
- Naming wise might be enough with `collectAnalyticsEvents`?
- Left a typo correction.

PR #361, opened 2025-03-06
"Details" button fixes

alfondotnet:
## Motivation / Description

- Increase the font to 14px in mobile (in desktop is already 14px)
- Move the "product description" back underneath the "Details" toggle as how it was (desktop still remains right underneath the title)
- Localize the string in all languages

### DIFF
--- a/src/ui/navbar-header.svelte
+++ b/src/ui/navbar-header.svelte
@@ -6,6 +6,7 @@
   import Localized from "./localization/localized.svelte";
 
   export let children: Snippet;
+  export let shouldShowDetailsButton = false;
   export let expanded = false;
   export let toggle;
 </script>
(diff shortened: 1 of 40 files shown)


src/ui/navbar-header.svelte
```diff
@@ -19,7 +21,7 @@
       aria-expanded={expanded}
       aria-controls="rcb-header-details-content"
     >
-      Details
+      <Localized key={LocalizationKeys.NavbarHeaderDetails} />
```
nicfix, line 24: :heart: Good catch

src/ui/navbar-header.svelte
```diff
@@ -31,15 +33,18 @@
     display: flex;
     align-items: center;
     cursor: pointer;
+    font-size: 14px;
```
nicfix, line 36: Shall we define a text style for this?
Something like `body2`, any name works, and even if it's just for this label it's fine, I just would like to have all the text styles in the text.ts file.

nicfix (approved):
Left a suggestion (creating a text style). But looks good otherwise.

src/ui/navbar-header.svelte
alfondotnet: yeah we talked about this and Iga said that while she was thinking on what new variation to create it was fine for a one off, but I'll create a "larger caption or something" in text

PR #362, opened 2025-03-06
Only show "Details" button when we need to

alfondotnet:
## Motivation / Description

Criteria is:

- If there is a product description AND "show product description" is checked
or
- If there are tax information (@vicfergar this is not ready yet right)
or
- If there's a trial

### DIFF
--- a/src/ui/rcb-ui-inner.svelte
+++ b/src/ui/rcb-ui-inner.svelte
@@ -11,7 +11,7 @@
   import StateSuccess from "./states/state-success.svelte";
   import { type CurrentView } from "./ui-types";
   import { type BrandingInfoResponse } from "../networking/responses/branding-response";
-  import type { Product, PurchaseOption } from "../main";
+  import type { Product, PurchaseOption, SubscriptionOption } from "../main";
   import StatePresentOffer from "./states/state-present-offer.svelte";
   import BrandingInfoUI from "./branding-info-ui.svelte";
   import {
@@ -36,6 +36,14 @@
   export let purchaseOperationHelper: PurchaseOperationHelper;
   export let isInElement: boolean = false;
 
+  const shouldShowDetailsButton =
+    (Boolean(brandingInfo?.appearance) &&
+      Boolean(productDetails?.description)) ||
+    Boolean(
+      (purchaseOptionToUse as SubscriptionOption)?.trial?.periodDuration,
+    ) ||
+    false;
+
   const viewsWhereOfferDetailsAreShown: CurrentView[] = [
     "present-offer",
     "needs-auth-info",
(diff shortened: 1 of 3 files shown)


src/ui/rcb-ui-inner.svelte
```diff
@@ -36,6 +36,14 @@
   export let purchaseOperationHelper: PurchaseOperationHelper;
   export let isInElement: boolean = false;
 
+  const shouldShowDetailsButton =
```
alfondotnet, line 39: point for refactor, don't like that this logic is coupled here

nicfix: Maybe add a comment for the potential refactoring?
Exposing this as a parameter of the component might be useful when we'll expose these components into `purchases-ui-js`

src/ui/rcb-ui-inner.svelte
```diff
@@ -36,6 +36,14 @@
...
+    (Boolean(brandingInfo?.appearance) &&
```
nicfix, line 40: We have an explicit flag to show/hide the product description.
So I would check explicitly for that flag.
```suggestion
    (Boolean(brandingInfo?.appearance?.show_product_description) &&
```

src/ui/rcb-ui-inner.svelte
```diff
@@ -36,6 +36,14 @@
...
+      Boolean(productDetails?.description)) ||
+    Boolean(
+      (purchaseOptionToUse as SubscriptionOption)?.trial?.periodDuration,
+    ) ||
+    false;
```
nicfix, line 45: I think we can avoid this false?

```suggestion
    );
```

nicfix, line 45: I usually prefer to split those complex conditions to make them more readable.
This is an opinionated approach so happy to not use it if you don't agree but I would split it as follows

```suggestion
  const shouldShowProductDescription= (Boolean(brandingInfo?.appearance) &&
      Boolean(productDetails?.description));
  
  const hasTrialInformation= Boolean(
      (purchaseOptionToUse as SubscriptionOption)?.trial?.periodDuration,
    )
  
  const shouldShowDetailsButton = shouldShowProductDescription || hasTrialInformation;
```

nicfix (approved):
Just a couple of nits, but overall it looks good.

src/ui/rcb-ui-inner.svelte
alfondotnet: sure I have have casting to boolean on the otherones I don't think I do, will remove

src/ui/rcb-ui-inner.svelte
alfondotnet: done

src/ui/rcb-ui-inner.svelte
alfondotnet: wow nice catch, I did have it like that and totally removed it without realizing

alfondotnet: done

src/ui/rcb-ui-inner.svelte
alfondotnet: yeah totally, that was my comment on refactoring, I think these "hasTrialInformation" "showShowProductDescription" etc should probably be in a context, but I'll change it here right now

alfondotnet: done

PR #363, opened 2025-03-06
[WEB-2291] sdk in the tablet the margins are too big

alfondotnet:


src/ui/layout/section-layout.svelte
```diff
@@ -53,6 +53,18 @@
     }
   }
 
+  @container layout-query-container (width >= 768px) and (width < 1024px) {
+    .layout-content {
```
alfondotnet, line 57: created https://linear.app/revenuecat/issue/WEB-2314/refactor-container-queries-in-purchases-js to track some work that's needed to improve complexity here

src/ui/theme/spacing.ts
```diff
@@ -1,45 +1,55 @@
...
     desktop: "0.375rem",
   },
   gapMedium: {
     mobile: "0.5rem",
+    tablet: "0.75rem",
     desktop: "0.75rem",
   },
   gapLarge: {
     mobile: "0.75rem",
+    tablet: "0.75rem",
```
alfondotnet, line 32: keeping all measures except outerPadding equal for tablet and desktop as they share layout, we can iterate if we need

src/ui/theme/spacing.ts
```diff
@@ -1,45 +1,55 @@
...
     mobile: string;
+    tablet: string;
     desktop: string;
   };
 }
 
 export const DEFAULT_SPACING: Spacing = {
   outerPadding: {
     mobile: "clamp(1.3125rem, 5.6vw, 1.5rem)",
+    tablet: "clamp(1.40625rem, 7.52vw, 3.25rem)",
```
alfondotnet, line 12: middle ground here

PR #364, opened 2025-03-06
Fon/review fixes

alfondotnet:
minor nits found when reviewing all the stories


src/assets/close.svg
alfondotnet, line 1: this is just a rename of a class

src/ui/sandbox-banner.svelte
```diff
@@ -16,17 +18,18 @@
...
+    <span class="rcb-sandbox-banner-text">Sandbox</span>
+    <div class="rcb-sandbox-banner-close-button-wrapper">
       <CloseButton on:click={closeBanner} />
     </div>
   </div>
 {/if}
 
 <style>
   .rcb-ui-sandbox-banner {
-    position: sticky;
```
alfondotnet, line 29: minor changes so we don't show the background below when it's closing. 

Sticky does this, and fixed cannot be applied in embedded as it jumps to the viewport

PR #368, opened 2025-03-10
Validate app user id on change

polmiro:
## Motivation / Description

While looking into [this reported issue](https://revenuecat.zendesk.com/agent/tickets/54693). I noticed the validation of `appUserId` is not happening when in the `changeUser` function. This is technically a breaking change but given it could be a source of bugs maybe it should be integrated anyways?

@tonidero wdyt?

### DIFF
--- a/src/tests/main.test.ts
+++ b/src/tests/main.test.ts
@@ -105,11 +105,16 @@ describe("Purchases.isEntitledTo", () => {
 });
 
 describe("Purchases.changeUser", () => {
-  test("can change user", () => {
+  test("throws error if given invalid user id", async () => {
+    const purchases = configurePurchases();
+    await expect(purchases.changeUser("")).rejects.toThrow(PurchasesError);
+  });
+
+  test("can change user", async () => {
     const newAppUserId = "newAppUserId";
     const purchases = configurePurchases();
     expect(purchases.getAppUserId()).toEqual(testUserId);
-    purchases.changeUser(newAppUserId);
+    await purchases.changeUser(newAppUserId);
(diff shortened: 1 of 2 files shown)


src/tests/main.test.ts
```diff
@@ -101,6 +101,11 @@ describe("Purchases.isEntitledTo", () => {
 });
 
 describe("Purchases.changeUser", () => {
+  test("throws error if given invalid user id", async () => {
+    const purchases = configurePurchases();
+    await expect(purchases.changeUser("")).rejects.toThrowError(PurchasesError);
```
tonidero, line 106: Does this need the `await`?

tonidero (approved):
Great catch! Yeah I think this is good. It does have a behavior change, but I think it wouldn't work correctly with the previous case any way, so this makes sense. Thank you!!

src/tests/main.test.ts
polmiro: If I do not add it the exception is happening outside of the test context and the test fails given that `changeUser` is returning a promise.

polmiro: But I can definitely write this in a better way. Give me a min.

antoniobg (approved):

src/tests/main.test.ts
tonidero: Ahh right, nvm then, I think it's ok!

polmiro: Actually I could not make it better. But I have updated the happy path test to also await to ensure there are no exceptions happening there.

tonidero (approved):
Makes sense, thank you!

PR #370, opened 2025-03-11
Upgrade vitest to 2.1.9

anperezrc:
Addresses CVE-2025-24964

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@revenuecat/purchases-js",
-  "version": "0.18.2",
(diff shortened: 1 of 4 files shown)


package-lock.json
```diff
@@ -1,12 +1,12 @@
 {
   "name": "@revenuecat/purchases-js",
-  "version": "0.18.2",
```
polmiro, line 3: I believe this change is just a leftover from the merge and release of the new version this Monday. It happened in my branch too earlier today so we can just keep it.

alfondotnet (approved):

PR #372, opened 2025-03-11
WEB-2327: Added a close and a back button when not in element

nicfix:
## Changes introduced
* Details are always shown now
* Details button has been replaced with the close button on mobile
* Added a close and back button (close on mobile, back on tablet/desktop)
* Generated the translations for the "Back" label


scripts/translate_locales.py
```diff
@@ -56,7 +56,7 @@ def translate_text(text, target_language, keys_context=None):
         system_prompt += "\n\nHere is additional context for some of the keys to help with translation:\n"
         for key, context in keys_context.items():
             if (
-                key in text
+                    key in text
```
alfondotnet, line 59: 👀

scripts/translate_locales.py
```diff
@@ -147,9 +147,9 @@ def process_json_files(directory, target_language, keys_to_update=None):
 
     for filename in os.listdir(directory):
         if (
-            filename.endswith(".json")
-            and filename != "en.json"
-            and filename != "keys_context.json"
+                filename.endswith(".json")
```
alfondotnet, line 150: 👀

src/ui/back-button.svelte
```diff
@@ -0,0 +1,29 @@
...
+  import Localized from "./localization/localized.svelte";
+  import { LocalizationKeys } from "./localization/supportedLanguages";
+
+  export let disabled = false;
+</script>
+
+<button on:click class="rcb-back-button" {disabled} data-testid="close-button">
+  <IconArrow className="back" />
+  <div>
+    <Localized key={LocalizationKeys.NavbarBackButton} />
```
alfondotnet, line 12: where is font-size defined here?

src/ui/back-button.svelte
```diff
@@ -0,0 +1,29 @@
...
+  </div>
+</button>
+
+<style>
+  .rcb-back-button {
+    border: none;
+    cursor: pointer;
+    background-color: transparent;
+    padding: 0;
+    height: 24px;
```
alfondotnet, line 22: should we otherwise define a variable for this?

src/ui/navbar-header.svelte
```diff
@@ -9,28 +11,57 @@
...
-        aria-expanded={expanded}
-        aria-controls="rcb-header-details-content"
-      >
-        <Localized key={LocalizationKeys.NavbarHeaderDetails} />
-        <IconArrow className={expanded ? "expanded" : "collapsed"} />
-      </button>
+  <div class="rcb-header-multiline-layout">
+    {#if showCloseButton}
+      <div class="rcb-back">
+        <BackButton
```
alfondotnet, line 22: should we not show the backbutton is `onClose` is not defined? (although it will always be defined I guess)

src/ui/navbar-header.svelte
```diff
@@ -66,5 +105,20 @@
...
+    }
+
+    .rcb-close {
+      display: none;
+    }
+
+    .rcb-header-multiline-layout {
+      all: unset;
+      display: flex;
+      flex-direction: column;
```
alfondotnet, line 120: these three are already defined with the same value above

alfondotnet (approved):
Some comments but overall looks good! 
I like it better always expanded, the app name alongside the logo looks good also.

Left a question in this Chromatic diff: https://revenuecat.chromatic.com/test?appId=6751ce4c50d424102499e900&id=67d07ea0c2900234447dc517

scripts/translate_locales.py
nicfix: Reverted, my IDE didn't take the setup for python in this project.

scripts/translate_locales.py
nicfix: Reverted, my IDE didn't take the setup for python in this project.

src/ui/back-button.svelte
nicfix: It's inherited by the button, I'll set it explicitly.

src/ui/back-button.svelte
nicfix: Removed it, this was needed when it was an icon button, now it has text.

src/ui/back-button.svelte
nicfix: Done

src/ui/navbar-header.svelte
nicfix: Removed

src/ui/navbar-header.svelte
nicfix: I think the current usage is safe.
I usually prefer to not assume on the existence of a parameter but rather explicitly mark as show/not show.

nicfix:
@alfondotnet the chromatic issue was connected to the theme I set up that had a black background, the button was becoming white due to that.
I brought back the theme to the previous color.

PR #374, opened 2025-03-12
Storybook with globals

polmiro:
## Motivation / Description

These are some improvements to the Storybook.

## Changes introduced

- Added global toolbar controls for:
    - Embedded viewport.
    - Locale (current PR)
    - Branding
- Consolidated the purchase flow stories.
- Added support for Chromatic modes.
- Simplified the translator fallback. Now set up once upon initialization of the main component.
- Components are now reactive to changes in the translator by using a Svelte store in the translator context.
- Froze the config of locale for stories that describe a specific locale.
- Upgrade storybook addons.
- Add a global decorator for stories to set the global context.

## Screenshots

Screenshot 2025-03-11 at 17 46 43


src/networking/responses/branding-response.ts
```diff
@@ -7,5 +7,4 @@ export type BrandingInfoResponse = {
   id: string;
   app_name: string | null;
   support_email?: string | null;
-  gateway_tax_collection_enabled: boolean;
```
nicfix, line 10: Why removing this?
Isn't it already returned by the backend?

nicfix (approved):
Left one question, but overall looks good.

One observation:
We increased a lot the usage of reactive variables (using $).

We should keep in in check to avoid having performance issues in the future.
By using the context we are splitting the "reactiveness" in each component and this might increase the re-rendering due to handling the change both in the parent and in the children.

Not a big deal but something to keep an eye on for the future.

src/networking/responses/branding-response.ts
polmiro: This is bizarre but I was getting type check errors in my runtime. I will review again.

polmiro: Reverted just in case, unsure why my env was complaining at the time.

PR #389, opened 2025-03-14
Reorganize components

polmiro:
## Motivation / Description

A couple of organization improvements:
- Organized svelte components into "./components" and "./layout".
- Rename `RcbUi` to `PurchasesUi` (and `RcbUiInner`).
- Rename `StatePresentOffer` to `ProductInfo` (because it always displayed in mobile name is more fitting).


antoniobg:
Looks good, but there are a lot of changes in the UI tests, it might be better to hardcode the date to avoid having the changes in every PR.

polmiro:
Good call, this was from the previous PR. I will address here.

antoniobg (approved):

PR #390, opened 2025-03-14
fix: improve anonymous AppUserID generation compatibility for mobile browsers

vicfergar:
## Motivation / Description

The previous implementation relied on `crypto.randomUUID()` which is not supported in some mobile browsers like Vivo browser on Android.

## Changes introduced

- Maintains `crypto.randomUUID()` as the preferred method when available
- Adds a fallback using `crypto.getRandomValues()` for broader compatibility
- Includes a final `Math.random()` fallback for maximum browser support

## Linear ticket (if any)

[WEB-2340](https://linear.app/revenuecat/issue/WEB-2340/%5Bsdk%5D-uuid-generation-fails-in-vivo-browser-on-android)

## Screenshots (representing the issue)

image


src/main.ts
```diff
@@ -771,8 +771,35 @@ export class Purchases {
    * @public
    */
   public static generateRevenueCatAnonymousAppUserId(): string {
-    const uuid = crypto.randomUUID();
-    return `${ANONYMOUS_PREFIX}${uuid.replace(/-/g, "")}`;
+    const generateUUID = (): string => {
```
tonidero, line 774: might be worth extracting into a utils file?

src/tests/main.test.ts
```diff
@@ -222,6 +222,36 @@ describe("Purchases.generateRevenueCatAnonymousAppUserId()", () => {
...
+    crypto.randomUUID = undefined as unknown as typeof crypto.randomUUID;
+
+    const getRandomValuesSpy = vi.spyOn(crypto, "getRandomValues");
+
+    const anonymousId = Purchases.generateRevenueCatAnonymousAppUserId();
+    expect(anonymousId).toMatch(/^\$RCAnonymousID:[0-9a-f]{32}$/);
+    expect(getRandomValuesSpy).toHaveBeenCalled();
+  });
+
+  test("falls back to Math.random when crypto methods are not available", () => {
```
tonidero, line 244: Hmm would it be worth adding some tests for each of these cases that calling the method multiple times, it generates different UUIDs? NABD though.

tonidero (approved):
Looks good! Thanks for fixing!

src/tests/main.test.ts
vicfergar: Addressed

src/main.ts
vicfergar: Done

antoniobg (approved):

nicfix (approved):

PR #397, opened 2025-03-14
Organize stories per atomic design

polmiro:
## Motivation / Description

Just to have some level of organization as we add more components. For now only organized stories but will map the components themselves in a different PR.

### DIFF
--- a/src/stories/fixtures.ts
+++ b/src/stories/fixtures.ts
@@ -175,13 +175,7 @@ export const checkoutStartResponse: CheckoutStartResponse = {
   gateway_params: {
     publishable_api_key: publishableApiKey,
     stripe_account_id: accountId,
-    elements_configuration: {
-      mode: StripeElementsMode.Payment,
-      payment_method_types: ["card"],
-      setup_future_usage: StripeElementsSetupFutureUsage.OffSession,
-      amount: 1000,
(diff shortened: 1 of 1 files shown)


src/stories/fixtures.ts
```diff
@@ -167,12 +167,21 @@ export const stripeElementsConfiguration = {
...
-    elements_configuration: stripeElementsConfiguration,
+    publishable_api_key: publishableApiKey,
+    stripe_account_id: accountId,
+    elements_configuration: {
+      mode: StripeElementsMode.Payment,
+      payment_method_types: ["card"],
+      setup_future_usage: StripeElementsSetupFutureUsage.OffSession,
+      amount: 1000,
+      currency: "usd",
+    },
```
vicfergar, line 184: Why not reusing the `stripeElementsConfiguration`?

polmiro: Oversight, thanks!

vicfergar (approved):
LGTM! There are a lot of `changes` without diffs in the UI tests. They all seem correct, but it would be better if someone familiar with the stories could review them.

polmiro:
The tests are only because of the rename of the components nesting based on atomic design groups (atoms, molecules, ...). So it's safe to approve.

PR #405, opened 2025-03-21
Refactor `PurchasesUIInner`

polmiro:
## Motivation / Description

Structural and naming improvements prior to adding tax calculation logic on top of `RCPurchasesUIInner`.

## Changes introduced

- Removed `present-offer` as a `CurrentView`. It is not a real page but a temporary value prior to deciding which page should be shown. Made it explicit by adding `initializing` as an extra option for `currentView` variable in `PurchasesUI` (outside of the `CurrentView` type used by `PurchasesUIInner`).
- Removed nullability of `productDetails` and `purchaseOption`. Is is not possible and it does not make sense to handle a purchase without a product or purchase option. This was propagating in a few places and strangely handled without reason and causing some confusion.
- Extracted a `Template` component responsible for composing the different layout items common to all pages.
- Moved the responsibility of calculating `colorVariables` to the `Template` component.
- Moved the responsibility of the fallback unknown error value from `PurchasesInnerUI` to the `StateError` page.
- Extracted `Loading` molecule component to be reused by both `StateLoading` and `StateNeedsPaymentInfo` to better separate different tier components.

### DIFF
--- a/src/stories/pages/purchase.stories.svelte
+++ b/src/stories/pages/purchase.stories.svelte
@@ -108,6 +108,6 @@
     defaultPurchaseOption: subscriptionOptionWithTrial,
   }}
 />
-<Story name="Loading" args={{ currentView: "loading" }} />
+<Story name="Loading" args={{ currentView: "loading-payment-page" }} />
 <Story name="Payment complete" args={{ currentView: "success" }} />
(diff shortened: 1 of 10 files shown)


nicfix (approved):
Nice cleanup with the template component!

Issue #271, opened 2025-01-08
Set custom user attributes


scamps86:
Hi,

I’m currently integrating your web SDK into my website, but I’m having trouble finding a way to set additional attributes for customers, similar to how I can do it with your Capacitor SDK in the app.

Is there a way to achieve this without resorting to your REST API?

Thanks in advance!

blazejkustra:
I'm facing the same problem, @scamps86 did you find a way to do it?

scamps86:
I couldn't find a way to implement it using the web SDK. I had to do it through their API.

rglanz-rc:
Unfortunately we don't have that implemented in the Web SDK just yet—so API is the required route for now.

(Here's the web SDK docs where you can find what's supported/not supported yet: https://revenuecat.github.io/purchases-js-docs/0.15.2/)

We are working to bring this SDK up to parity with the relevant features from our mobile SDKs (like setting customer attributes).

nicfix:
It is now possible to set Subscriber Attributes through `purchases-js`.

Download the latest version to start using it, docs [here](https://www.revenuecat.com/docs/customers/customer-attributes#setting-attributes) (you can see Web was added to the list of languages supported)

PR #416, opened 2025-03-28
[WEBF-4] Support subscriber attributes

antoniobg:
## Motivation / Description
This PR brings in support for setting [customer attributes](https://www.revenuecat.com/docs/customers/customer-attributes#setting-attributes).

## Changes introduced

- Added the `setAttributes` method to allow setting customer-specific attributes.
- Added unit tests
- Specified locale settings in the test configuration for better compatibility.
- Use customer attributes in demo app
- Added e2e tests


src/main.ts
```diff
@@ -694,6 +694,20 @@ export class Purchases {
...
   }
 
+  /**
+   * Sets attributes for the current user. Attributes are useful for storing additional, structured information on a customer that can be used elsewhere in the system.
+   * For example, you could store your customer's email address or additional system identifiers through the applicable reserved attributes, or store arbitrary facts like onboarding survey responses, feature usage, or other dimensions as custom attributes.
+   * @param attributes - A dictionary of attributes to set for the current user.
+   * @throws {@link PurchasesError} if there is an error while setting the attributes or if the customer doesn't exist.
+   */
+  public async setAttributes(attributes: {
+    [key: string]: string;
```
tonidero, line 704: We might want to allow for nullable values...
```suggestion
    [key: string]: string | null;
```
This would mimick how it works in the mobile SDKs.

src/main.ts
```diff
@@ -694,6 +694,20 @@ export class Purchases {
     return this._appUserId;
   }
 
+  /**
+   * Sets attributes for the current user. Attributes are useful for storing additional, structured information on a customer that can be used elsewhere in the system.
+   * For example, you could store your customer's email address or additional system identifiers through the applicable reserved attributes, or store arbitrary facts like onboarding survey responses, feature usage, or other dimensions as custom attributes.
+   * @param attributes - A dictionary of attributes to set for the current user.
+   * @throws {@link PurchasesError} if there is an error while setting the attributes or if the customer doesn't exist.
+   */
+  public async setAttributes(attributes: {
```
tonidero, line 703: Not for this PR! After this, we might want to add smaller convenience methods for some of the reserved attributes... or if we're concerned about making multiple calls for each individual call, we can think about other options.

src/main.ts
```diff
@@ -694,6 +694,20 @@ export class Purchases {
...
+    [key: string]: string;
+  }): Promise<void> {
+    // Ensure the customer exists by calling getCustomerInfo
+    await this.getCustomerInfo();
+    return await this.backend.setAttributes(this._appUserId, attributes);
```
tonidero, line 708: We might want to wrap any errors from the backend into some more user-friendly errors... At least for any "common" errors.

src/main.ts
```diff
@@ -694,6 +694,20 @@ export class Purchases {
     return this._appUserId;
   }
 
+  /**
+   * Sets attributes for the current user. Attributes are useful for storing additional, structured information on a customer that can be used elsewhere in the system.
+   * For example, you could store your customer's email address or additional system identifiers through the applicable reserved attributes, or store arbitrary facts like onboarding survey responses, feature usage, or other dimensions as custom attributes.
```
tonidero, line 699: Not sure if it's worth saying here that we won't cache nor retry sending these attributes... I'm a bit concerned with the difference in behavior in the hybrids making it more difficult to use there... I think something I will want to discuss with the SDK team.

tonidero (commented):
Looking great! Just some small comments.

src/main.ts
tonidero, line 703: In the mobile SDKs, we have a `syncAttributesAndOfferingsIfNeeded` that additionally to syncing attributes, it also fetches offerings in case targeting changed... We might want to do something like that here as well... probably less important since we are not caching the value here though.

src/main.ts
tonidero: Nvm this, we are already doing this handling inside the backend/http-client. Sorry for the noise!

src/main.ts
antoniobg: Since we're not caching anything, I think adding this method doesn't add lot of value, and if someone wants to get the offering after setting attributes, they could explicitly call `getOfferings`.

tonidero: Agreed 👍

src/entities/errors.ts
```diff
@@ -216,6 +218,7 @@ export enum BackendErrorCode {
   BackendInvalidOperationSession = 7877,
   BackendPurchaseCannotBeCompleted = 7878,
   BackendEmailIsRequired = 7879,
+  BackendSubscriberNotFound = 7259,
```
tonidero, line 221: NABD but should we move this a bit above so the codes are ordered?

src/networking/backend.ts
```diff
@@ -200,4 +201,41 @@ export class Backend {
       },
     );
   }
+
+  async setAttributes(
```
tonidero, line 205: One thing that I was considering, in the mobile SDKs, since we cache these, we basically don't send them again if the value hasn't changed, which can potentially save on requests. I can see how some devs might be calling this every time, and we would set the attributes every time. I think this is probably fine for now, but just wanted to call it out so we're aware.

tonidero (approved):
Changes LGTM! Might be worth getting another review from @nicfix as well, to make sure we all agree.

examples/webbilling-demo/src/pages/paywall/index.tsx
```diff
@@ -95,6 +95,31 @@ const PaywallPage: React.FC = () => {
...
+  const nickname = searchParams.get("nickname");
+
+  useEffect(() => {
+    const setAttributes = async () => {
+      const attributes: { [key: string]: string } = {};
+      if (displayName) {
+        attributes["$displayName"] = displayName;
+      }
+      if (nickname) {
+        attributes["nickname"] = nickname;
```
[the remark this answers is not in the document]
antoniobg: Yes, there are some attributes that are [reserved and defined by RevenueCat](https://www.revenuecat.com/docs/customers/customer-attributes#reserved-attributes), which start with `$`, and others that are custom, defined by developers.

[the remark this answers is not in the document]
tonidero: Yeah I do wonder if we might want specific setters for those reserved attributes... But I guess that could result in increased calls, depending on the developer implementation. We can add those in a separate PR later though.

alfondotnet (approved):

examples/webbilling-demo/src/pages/paywall/index.tsx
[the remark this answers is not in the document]
vicfergar: Maybe just adding an enum with the reserved attribute names would be enough? I think we do [the same](https://github.com/RevenueCat/purchases-ios/blob/224b15f73152979222a490d823284bd736ffd1a0/Sources/SubscriberAttributes/ReservedSubscriberAttributes.swift#L22-L25) in other SDKs, right?

src/main.ts
vicfergar: I mentioned this in a comment above, couldn't we include an `enum` with reserved attribute names and do something like:

```
export enum ReservedSubscriberAttribute {
    email = "$email",
    phoneNumber = "$phoneNumber",
    displayName = "$displayName",
    pushToken = "$apnsTokens"
    
    ...
}

...

type AttributeKey = ReservedSubscriberAttribute | (string & {});

public async setAttributes(attributes: Record<AttributeKey, string | null>): Promise<void> {
}
```

vitest.config.js
```diff
@@ -50,6 +50,12 @@ export default defineConfig(() => ({
     // Other configurations like shims for APIs or polyfills can also be included
     // ...
     exclude: [...configDefaults.exclude, "examples/**"],
+
+    // Set environment variables for tests
+    env: {
+      LC_ALL: "en_US.UTF-8",
+      LANG: "en_US.UTF-8",
```
vicfergar, line 57: ❤️

src/main.ts
```diff
@@ -694,6 +694,25 @@ export class Purchases {
...
+   */
+  public async setAttributes(attributes: {
+    [key: string]: string | null;
+  }): Promise<void> {
+    /*
+     * Ensure the customer exists by calling getCustomerInfo as setAttributes will throw an error
+     * if the customer doesn't exist.
+     *
+     * Note: We may want to optimize this in the future by prefetching customer info during SDK initialization.
+     */
```
alfondotnet, line 711: or at least we could store a boolean in the instance if we've called `getCustomerInfo()` once, so subsequent attr calls don't have to call it again. 

NBD anyway as it will be a fast call

alfondotnet (approved):

src/tests/main.test.ts
```diff
@@ -326,3 +332,92 @@ describe("Purchases.purchase()", () => {
...
+      http.post(
+        "http://localhost:8000/v1/subscribers/someAppUserId/attributes",
+        () => {
+          return HttpResponse.json({}, { status: StatusCodes.OK });
+        },
+      ),
+    );
+
+    await purchases.setAttributes({ name: "John", age: "30" });
+  });
```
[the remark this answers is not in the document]
antoniobg: Done

alfondotnet:
Fixes: #271

src/entities/attributes.ts
```diff
@@ -0,0 +1,194 @@
...
+  /**
+   * The phone number of the customer.
+   */
+  PhoneNumber = "$phoneNumber",
+  /**
+   * iOS advertising identifier UUID for the customer.
+   */
+  IDFA = "$idfa",
+  /**
+   * iOS vender identifier UUID for the customer.
```
vicfergar, line 23: typo?
```suggestion
   * iOS vendor identifier UUID for the customer.
```

antoniobg: https://github.com/RevenueCat/docs/pull/821

examples/webbilling-demo/src/pages/paywall/index.tsx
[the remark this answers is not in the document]
antoniobg: Done

src/entities/errors.ts
antoniobg: Done

src/main.ts
antoniobg: Added a note

vicfergar (approved):

src/main.ts
antoniobg: I'm going to leave this optimization for later

PR #422, opened 2025-04-01
Wrapping up the first implementation

nicfix:

### DIFF
--- a/src/stripe/stripe-service.ts
+++ b/src/stripe/stripe-service.ts
@@ -15,6 +15,7 @@ import type {
   GatewayParams,
   StripeElementsConfiguration,
 } from "../networking/responses/stripe-elements";
+
 export class StripeService {
   private static FORM_VALIDATED_CARD_ERROR_CODES = [
     "card_declined",
@@ -170,6 +171,11 @@ export class StripeService {
       layout: {
         type: "tabs",
       },
+      fields: {
+        billingDetails: {
+          address: "auto",
+        },
+      },
       terms: {
         applePay: "never",
         auBecsDebit: "never",
@@ -184,6 +190,10 @@ export class StripeService {
         usBankAccount: "never",
       },
     });
+  }
+
+  static createAddressElement(elements: StripeElements) {
+    return elements.create("address", { mode: "billing" });
   }
 
   static async confirmIntent(
--- a/src/ui/molecules/stripe-payment-elements.svelte
+++ b/src/ui/molecules/stripe-payment-elements.svelte
@@ -4,6 +4,7 @@
     Appearance,
     ConfirmationToken,
     Stripe,
+    StripeAddressElement,
     StripeElementLocale,
     StripeElements,
     StripeError,
@@ -24,7 +25,8 @@
     type PaymentElementError,
     PaymentElementErrorCode,
   } from "../types/payment-element-error";
-  import { type TaxCustomerDetails } from "../ui-types";
+  import { type PriceBreakdown, type TaxCustomerDetails } from "../ui-types";
+  import type { StripeAddressElementChangeEvent } from "@stripe/stripe-js/dist/stripe-js/elements/address";
 
   export let gatewayParams: GatewayParams;
   export let brandingInfo: BrandingInfoResponse | null;
@@ -41,6 +43,7 @@
     customerDetails: TaxCustomerDetails,
   ) => void;
   export let stripeLocale: StripeElementLocale | undefined = undefined;
+  export let priceBreakdown: PriceBreakdown;
 
   export async function submit() {
     if (!elements) return;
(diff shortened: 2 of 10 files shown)


src/ui/purchases-ui.svelte
```diff
@@ -215,22 +216,27 @@
...
+        taxCalculation.total_excluding_tax_in_micros;
+      priceBreakdown.taxBreakdown =
+        taxCalculation.pricing_phases.base.tax_breakdown;
+      priceBreakdown.pendingReason = null;
+      gatewayParams.elements_configuration =
+        taxCalculation.gateway_params.elements_configuration;
+    } catch (e) {
+      priceBreakdown.taxCalculationStatus = "pending";
+      priceBreakdown.pendingReason = "needs_complete_billing_address";
+      priceBreakdown.taxCalculationBasedOnFullAddress = true;
```
vicfergar, line 238: why do we need this flag if we already have a specific `pending_reason`?

src/ui/molecules/stripe-payment-elements.svelte
```diff
@@ -232,82 +235,117 @@
...
+      addressElement = StripeService.createAddressElement(elements);
+      addressElement.mount("#address-element");
+      addressElement.on(
+        "change",
+        async (event: StripeAddressElementChangeEvent) => {
+          if (taxCollectionEnabled && event.complete && event.isNewAddress) {
+            await triggerTaxDetailsUpdated();
+          }
+        },
+      );
```
vicfergar, line 338: Would it make sense to move the `initialize` and `destroy` logic for this element into `onMount()`? Then here, we’d just call `addressElement.mount("#address-element")` and `addressElement.unmount()` as needed. Curious if that might help simplify things a bit.

src/ui/molecules/stripe-payment-elements.svelte
```diff
@@ -232,82 +235,117 @@
...
+      addressElement?.destroy();
+      addressElement = null;
+    }
+  })();
 </script>
 
-<div id="payment-element"></div>
+<div
+  style="display: flex; flex-direction: column; align-items: stretch; gap: 21px;"
+>
```
vicfergar, line 348: Why is this needed?

src/ui/pages/payment-entry-page.svelte
```diff
@@ -167,6 +167,7 @@
           bind:submit={stripeSubmit}
           bind:confirm={stripeConfirm}
           bind:stripeLocale
+          bind:priceBreakdown
```
vicfergar, line 170: A `bind` should not be needed in this case, we can pass eighter the entire `priceBreakdown` or a new field as we did for `taxCollectionEnabled`

src/stripe/stripe-service.ts
```diff
@@ -170,6 +171,11 @@ export class StripeService {
       layout: {
         type: "tabs",
       },
+      fields: {
+        billingDetails: {
+          address: "auto",
+        },
+      },
```
vicfergar, line 178: Since this is the default value, I believe we shouldn’t need to set it explicitly

src/ui/purchases-ui.svelte
nicfix: We need the flag since when the address is resolved correctly the status moves to "complete" and the pendingReason becomes null.

If we don't have the flag the address elements will be hidden.

src/ui/molecules/stripe-payment-elements.svelte
vicfergar: What I mean is, initializing the address element only once on mount and call only mount/unmount (add the element to the DOM) at this point.

src/ui/molecules/stripe-payment-elements.svelte
vicfergar: Ok, I thought this was automatically handled by Stripe

src/ui/purchases-ui.svelte
vicfergar: Got it, thanks for the explanation!

src/ui/molecules/stripe-payment-elements.svelte
vicfergar: I looks like we need to create this element on demand to set [defaultValues](https://docs.stripe.com/js/elements_object/create_address_element#address_element_create-options-defaultValues). So lets keep your initial approach

src/stripe/stripe-service.ts
```diff
@@ -186,6 +192,10 @@ export class StripeService {
     });
   }
 
+  static createAddressElement(elements: StripeElements) {
+    return elements.create("address", { mode: "billing" });
+  }
```
vicfergar, line 197: We could improve the customer experience by passing tax customer details as default values:
```suggestion
  static createAddressElement(
    elements: StripeElements,
    country: string,
    postalCode?: string,
  ) {
    return elements.create("address", {
      mode: "billing",
      defaultValues: {
        address: {
          country: country,
          postal_code: postalCode,
        },
      },
    });
  }
```

src/ui/molecules/stripe-payment-elements.svelte
nicfix: I am refactoring the implementation that got too complex.

We'll have something like

```
{#if elements}
 <PaymentElement {...}/>
 {#if priceBreakdown.taxCalculationBasedOnFullAddress}
 <AddressElement {...}/>
 {/if}
{/if}


```

src/ui/ui-types.ts
```diff
@@ -33,6 +35,10 @@ export type PriceBreakdown = {
 };
 
 export type TaxCustomerDetails = {
-  countryCode: string | undefined;
-  postalCode: string | undefined;
+  country: string | undefined;
```
nicfix, line 38: Extended these details to support all the needed information to request calculation for the full address.

src/ui/pages/payment-entry-page.svelte
```diff
@@ -163,10 +162,11 @@
   >
     <div class="rc-checkout-form-container" hidden={!!modalErrorMessage}>
       <div class="rc-payment-element-container">
-        <StripePaymentElements
+        <StripeElements
           bind:submit={stripeSubmit}
           bind:confirm={stripeConfirm}
           bind:stripeLocale
+          {priceBreakdown}
```
vicfergar, line 169: Since we're passing `priceBreakdown`, lets remove the field `taxCollectionEnabled`

PR #424, opened 2025-04-03
This adds an API to set PlatformInfo information to be used by hybrids integrating the JS SDK

tonidero:
## Motivation / Description
We have some extra headers to identify when an SDK is used through a hybrid SDK. This allows the hybrid SDK to provide this information so we can get accurate data.

### DIFF
--- a/api-report/purchases-js.api.md
+++ b/api-report/purchases-js.api.md
@@ -243,6 +243,12 @@ export enum PeriodUnit {
 }
 
 // @public
+export interface PlatformInfo {
+    readonly flavor: string;
+    readonly version: string;
+}
+
+// @public
 export interface PresentedOfferingContext {
     readonly offeringIdentifier: string;
     readonly placementIdentifier: string | null;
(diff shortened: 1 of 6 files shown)


vicfergar (approved):

api-report/purchases-js.api.md
```diff
@@ -242,6 +242,12 @@ export enum PeriodUnit {
     Year = "year"
 }
 
+// @public
+export interface PlatformInfo {
```
JayShortway, line 246: I guess there's no way to mark this as internal, or do something like `@InternalRevenueCatAPI` / `@_spi(Internal)`? (Same for `setPlatformInfo()`.)

JayShortway (approved):
Good one! Just a question.

api-report/purchases-js.api.md
tonidero: We can mark it with `@internal` but then it can't be used externally... The other thing would be to mark it with `@experimental`... but also doesn't feel like an experimental API, just something people shouldn't normally use 😅 . We do have the same API in iOS/Android, so I think it's probably okish... Lmk if you feel otherwise!

tonidero:
Will merge this for now. Lmk if you have other concerns and happy to address in a follow-up

api-report/purchases-js.api.md
JayShortway: Hmm yea, I guess we'd need something that acts like `@experimental` but has another name? I'd also be fine with `@experimental` as that _does_ discourage usage, even if it's not an accurate name. Not a blocker for me!

PR #425, opened 2025-04-03
[WEB-2367] Handle tax calculation errors (FF)

vicfergar:
## Motivation / Description

Handle the following error that could be returned as a tax calculation result:

- Tax collection is not enabled in Stripe
- The tax [origin address](https://docs.stripe.com/tax/set-up#origin-address) is either missing or invalid (as per Stripe configuration)
- Required permissions are missing
- Stripe account is in test mode only (Sandbox environment only)
- [Tax location is invalid](https://docs.stripe.com/tax/custom#handling-customer-location-errors)

## Changes introduced

ca47be6 - chore: Update API Specs
8d5a604 - feat: Represent Underlying ErrorMessage in Sandbox mode
bdab2a1 - feat: Handle tax calculation errors
1e1fe78 - chore: Remove unused code

## Linear ticket (if any)

[WEB-2367](https://linear.app/revenuecat/issue/WEB-2367/%5Bwebsdk%5D-handle-possible-tax-calculation-errors)

### DIFF
--- a/src/ui/purchases-ui.svelte
+++ b/src/ui/purchases-ui.svelte
@@ -223,8 +223,8 @@
           priceBreakdown.taxCalculationStatus = null;
           break;
         case TaxCalculationError.InvalidLocation:
-          // TODO: Show an error message
           priceBreakdown.taxCalculationStatus = "pending";
+          priceBreakdown.pendingReason = "invalid_postal_code";
           break;
         default:
           handleError(
--- a/src/ui/localization/locale/ar.json
+++ b/src/ui/localization/locale/ar.json
@@ -50,6 +50,7 @@
   "error_page.error_message_network_error": "خطأ في الشبكة. يرجى التحقق من اتصال الإنترنت.",
   "error_page.error_message_error_charging_payment": "فشل الدفع.",
   "error_page.error_message_error_setting_up_purchase": "لم يبدأ الشراء بسبب خطأ. رمز الخطأ: {{errorCode}}",
+  "error_page.error_message_invalid_tax_location": "لم نتمكن من التحقق من عنوان الفوترة الخاص بك. يرجى التحقق مرة أخرى من الرمز البريدي والبلد للتأكد من صحتهما.",
   "error_page.error_message_unknown_error": "حدث خطأ غير معروف. رمز الخطأ: {{errorCode}}",
   "error_page.button_try_again": "حاول مرة أخرى",
   "paywall_variables.price_per_period": "{{formattedPrice}}/{{period}}",
--- a/src/ui/localization/locale/ca.json
+++ b/src/ui/localization/locale/ca.json
@@ -50,6 +50,7 @@
   "error_page.error_message_network_error": "Error de xarxa. Comprova la teva connexió a internet.",
   "error_page.error_message_error_charging_payment": "El pagament ha fallat.",
   "error_page.error_message_error_setting_up_purchase": "La compra no s'ha iniciat a causa d'un error. Codi d'error: {{errorCode}}",
+  "error_page.error_message_invalid_tax_location": "No hem pogut verificar la vostra adreça de facturació. Si us plau, verifiqueu de nou el codi postal i el país per assegurar-vos que són correctes.",
   "error_page.error_message_unknown_error": "S'ha produït un error desconegut. Codi d'error: {{errorCode}}",
   "error_page.button_try_again": "Torna a intentar-ho",
   "paywall_variables.price_per_period": "{{formattedPrice}}/{{period}}",
--- a/src/ui/localization/locale/cs.json
+++ b/src/ui/localization/locale/cs.json
@@ -50,6 +50,7 @@
   "error_page.error_message_network_error": "Síťová chyba. Zkontrolujte prosím své internetové připojení.",
   "error_page.error_message_error_charging_payment": "Platba selhala.",
   "error_page.error_message_error_setting_up_purchase": "Nákup nebyl zahájen kvůli chybě. Kód chyby: {{errorCode}}",
+  "error_page.error_message_invalid_tax_location": "Nemohli jsme ověřit vaši fakturační adresu. Zkontrolujte prosím, zda je poštovní směrovací číslo a země správné.",
   "error_page.error_message_unknown_error": "Došlo k neznámé chybě. Kód chyby: {{errorCode}}",
   "error_page.button_try_again": "Zkusit znovu",
   "paywall_variables.price_per_period": "{{formattedPrice}}/{{period}}",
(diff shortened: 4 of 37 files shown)


src/entities/errors.ts
```diff
@@ -36,10 +36,16 @@ export enum ErrorCode {
...
+  RateLimitError = 39,
+  TaxCollectionNotEnabledError = 40,
+  TaxLocationCannotBeDeterminedError = 41,
+  InvalidTaxLocationError = 42,
+  GatewaySetupError = 43,
+  SandboxModeOnlyError = 44,
 }
 
 export class ErrorCodeUtils {
-  static getPublicMessage(errorCode: ErrorCode): string {
```
vicfergar, line 42: Not every `ErrorCode` should have a public message

src/ui/pages/error-page.svelte
```diff
@@ -115,7 +121,11 @@
   {/snippet}
 
   {#snippet message()}
-    {getTranslatedErrorMessage()}
+    {#if isSandbox && error.underlyingErrorMessage}
+      {error.underlyingErrorMessage}
```
vicfergar, line 125: We have some e2e broken because of this change. I can move this to another PR since it affects non-taxes errors too.

I also think we should have storybooks stories covering this

src/entities/errors.ts
```diff
@@ -36,10 +36,16 @@ export enum ErrorCode {
...
+  static getPublicMessage(errorCode: ErrorCode): string | undefined {
```
vicfergar, line 48: Not every ErrorCode should have a public message. Hence the `undefined`

src/entities/errors.ts
```diff
@@ -36,10 +36,16 @@ export enum ErrorCode {
   CustomerInfoError = 28,
   SignatureVerificationError = 36,
   InvalidEmailError = 38,
+  RateLimitError = 39,
+  TaxCollectionNotEnabledError = 40,
+  TaxLocationCannotBeDeterminedError = 41,
+  InvalidTaxLocationError = 42,
+  GatewaySetupError = 43,
+  SandboxModeOnlyError = 44,
```
tonidero, line 44: Oh wait... I think we got these wrong at some point... Ideally, these error codes are shared accross the mobile SDKs so the hybrids (once we support web in the hybrids) can map these errors to their own errors correctly. I should have totally added something here to notify 🤦 

So these codes are already being used by the mobile SDKs like [iOS](https://github.com/RevenueCat/purchases-ios/blob/ea9afca69c8701387fe85d555e72cea7691bdbd8/Sources/Error%20Handling/ErrorCode.swift#L20) or [Android](https://github.com/RevenueCat/purchases-android/blob/89c7d571f4a259fd96c12c2f67ac6a3c4bb0d827/purchases/src/main/kotlin/com/revenuecat/purchases/errors.kt#L53). And looks like a couple of errors above do not match the ones in the mobile SDKs. We might need to change that, to avoid errors when we support this in the hybrids.

We need to either:
- Add these with a higher version than the numbers used in iOS (which has the highest error code numbers right now), and add them to the android and iOS SDKs repos (though ideally we should add them to all the hybrids 😅 )
- Not expose these as error codes themselves, but as a message only and fallback to one of the error codes above.

So I think the decision depends on whether we want the developer to take a specific action on these errors, or just fall

src/entities/errors.ts
tonidero: I wonder about this... I think these developer facing error codes should all have a message. In what situation we wouldn't want to have a message?

tonidero (commented):
Some questions about the error handling (which admittedly, I think it's wrong right now 😞)

src/entities/errors.ts
vicfergar: Oh, I thought `PublicMessage` referred to an end-customer-facing message. I also found that `PurchasesError.message` can be `undefined`, so I thought that might be the case as well.

src/ui/pages/error-page.svelte
vicfergar: I dropped this change, we can include this as part of a different PR

api-report/purchases-js.api.json
```diff
@@ -1,7 +1,7 @@
 {
   "metadata": {
     "toolPackage": "@microsoft/api-extractor",
-    "toolVersion": "7.50.1",
```
nicfix, line 4: @vicfergar can you bump the version of the api-extractor tool on your machine and run again the generation?

src/helpers/purchase-operation-helper.ts
```diff
@@ -66,30 +64,6 @@ export class PurchaseFlowError extends Error {
     );
   }
 
-  getPublicErrorMessage(productDetails: Product | null): string {
```
nicfix, line 69: Was this function not used?

src/ui/purchases-ui.svelte
```diff
@@ -207,30 +213,49 @@
...
+        case TaxCalculationError.Pending:
+          priceBreakdown.taxCalculationStatus = "pending";
+          break;
+        case TaxCalculationError.Disabled:
+          priceBreakdown.taxCollectionEnabled = false;
+          priceBreakdown.taxCalculationStatus = null;
+          break;
+        case TaxCalculationError.InvalidLocation:
+          // TODO: Show an error message
+          priceBreakdown.taxCalculationStatus = "pending";
```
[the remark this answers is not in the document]
nicfix: TL:DR; we can merge this PR as is.

The long version :) 
Is this showing a "Invalid Zip Code" in the Stripe Elements form?

If not then we need to do something to let the user understand that, even if they inserted a zip code, we were not able to calculate taxes and therefore they cannot purchase.

That's why I think we should follow up with the address since it moves this behaviour to Stripe Elements and allows for a better input of the actual address using the Google Maps api.

We can go on for this PR and then I'll follow up with my PR that uses the Stripe Address as soon as I'll finish storing the address in the Stripe Customer rather than in our DB.

nicfix (commented):
Just a couple of questions.

src/entities/errors.ts
vicfergar: We agreed to map the new backend errors to existing `ErrorCodes` (option 2 above).

api-report/purchases-js.api.json
vicfergar: Since I removed the new `ErrorCodes`, there are no changes in the API report

src/helpers/purchase-operation-helper.ts
vicfergar: Nope, we are already using the translator for this

api-report/purchases-js.api.json
vicfergar: BTW I noticed that we're pinging an older version in our `packages.json`, that's why it gets changed in this case

> "@microsoft/api-extractor": "^7.48.0",

src/ui/purchases-ui.svelte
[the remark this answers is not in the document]
vicfergar: Done, the following message will be shown in this case:
image

nicfix (approved):

PR #428, opened 2025-04-06
Add pricePerWeek/pricePerMonth/pricePerYear APIs to PricingPhase

tonidero:
## Motivation / Description
This adds some convenience info to the `PricingPhase` type, that includes the price for different periods.


ajpallares (approved):
I think it looks good!

src/entities/offerings.ts
```diff
@@ -128,6 +132,21 @@ export interface PricingPhase {
...
+  /**
+   * The price converted to monthly rate, based on the period information.
+   * Null in case of trials.
+   */
+  readonly pricePerMonth: Price | null;
+  /**
+   * The price converted to yearly rate, based on the period information.
+   * Null in case of trials.
+   */
+  readonly pricePerYear: Price | null;
```
JayShortway, line 149: I think we should document that these are approximations.

src/entities/offerings.ts
```diff
@@ -399,14 +418,91 @@ const toPrice = (priceData: PriceResponse): Price => {
...
+function getPriceConversionFactor(period: Period): {
+  toWeek: number;
+  toMonth: number;
+  toYear: number;
+} {
+  const { number, unit } = period;
+
+  const DAYS_PER_WEEK = 7;
+  const DAYS_PER_MONTH = 30.44;
+  const DAYS_PER_YEAR = 365.25;
```
JayShortway, line 477: Should we use the same values as we do in other platforms? In Android, for instance: 
```kotlin
private object PeriodConstants {
    const val DAYS_PER_WEEK = 7.0
    const val DAYS_PER_MONTH = 30.0
    const val DAYS_PER_YEAR = 365.0
    // ...
}
```

JayShortway (commented):
Thanks for adding these! Just 2 small comments on consistency and documentation.

src/entities/offerings.ts
tonidero: You're right... we should be consistent on this. Also checked iOS and these are the values we use there. Thanks!!

src/entities/offerings.ts
tonidero: Changed in https://github.com/RevenueCat/purchases-js/pull/428/commits/325d106b28e5637281805cff378ca5e0eeb7befd

JayShortway (approved):
Thanks for making the changes!

PR #429, opened 2025-04-07
Fix PeriodUnit enum not being exported correctly

tonidero:
## Motivation / Description
We were not exporting the `PeriodUnit` enum correctly, which can cause issues in consumers if trying to use that type directly. For enums, we need to export the enum itself as well, not just the type.


ajpallares (approved):
TIL! Thanks for fixing!

PR #432, opened 2025-04-09
WEB-2370 - part 1: Refactored the Stripe Payment Elements component to isolate the Payment Elements part

nicfix:
## Motivation / Description
We need to add a custom Address Element and we can add the Email element as well using the work from @polmiro in a previous PR.

This PR prepares the ground cleaning up the `stripe-elements` component splitting the payment elements part in its dedicated component.

The next PRs will add the address and email components.

## Changes introduced
*  Refactored the Stripe Payment Elements component to isolate the Payment Elements part


src/ui/molecules/stripe-elements.svelte
```diff
@@ -194,120 +176,98 @@
...
+      paymentMethod: event.complete ? event.value.type : undefined,
+    });
+  };
 </script>
 
-<div id="payment-element"></div>
+<div
+  style="display: flex; flex-direction: column; align-items: stretch; gap: {spacing
+    .gapXLarge[viewport]};"
+>
```
vicfergar, line 261: Since we are not adding new components in this PR, this should not be needed for now, right?

nicfix: I will be adding the Address and the Email in the follow up ones, but you are right, is not strictly needed in this PR, removing it.

nicfix: Removed.

src/ui/molecules/stripe-elements.svelte
```diff
@@ -69,40 +70,11 @@
...
-    (async () => {
-      const elementsConfiguration = gatewayParams.elements_configuration;
-      if (!elementsConfiguration) return;
-
-      await StripeService.updateElementsConfiguration(
-        elements,
-        elementsConfiguration,
-      );
-    })();
-  }
```
vicfergar, line 82: It seems like this is a breaking change, we should call `updateElementsConfiguration` whenever `gatewayParams.elements_configuration` changes.

There is also no reason to call `updateElementsConfiguration` on mount.

src/ui/molecules/stripe-elements.svelte
```diff
@@ -141,13 +113,6 @@
 
   $: stripeLocale = getLocaleToUse(initialLocale);
 
-  $: {
-    // @ts-ignore
-    if (unsafeElements && unsafeElements._elements.length > 0) {
-      elements = unsafeElements;
-    }
-  }
```
vicfergar, line 149: This also looks like a behavioral change, we are no longer checking `unsafeElements._elements.length > 0` before initializing `elements`

src/ui/molecules/stripe-elements.svelte
```diff
@@ -194,120 +176,93 @@
...
-    const { countryCode, postalCode } =
-      getCountryAndPostalCodeFromConfirmationToken(confirmationToken);
+    const billingAddress =
+      confirmationToken.payment_method_preview?.billing_details?.address;
 
     if (
-      countryCode === lastTaxCustomerDetails?.countryCode &&
-      postalCode === lastTaxCustomerDetails?.postalCode
+      !billingAddress ||
+      JSON.stringify(billingAddress) === JSON.stringify(lastTaxCustomerDetails)
```
vicfergar, line 184: We are comparing a type returned by Stripe with a custom type. If any of these types change, it will break this behavior.

I’m a bit concerned about introducing behavioral changes in this PR, since we’re still missing integration tests for the tax calculation part and could unintentionally introduce breaking changes.

nicfix: I'll check the specific fields just to be sure

src/ui/molecules/stripe-elements.svelte
nicfix: Perfect, understood now why we had this.

Since the `elements` variable will get assigned only by the initStripe function now, we don't need to watch the elements variables per se, but we need to update the Elements configuration whenever the `gatewayParams.elements_configuration` changes, adding it.

src/ui/molecules/stripe-elements.svelte
nicfix: As discussed in a huddle I'll change this to be a more explicit check of the fact that the elements loaded.

I'll add that check to the confirm and submit functions.

nicfix: Added the paymentElementReadyForSubmission variable to explicitly handle this scenario:

https://github.com/RevenueCat/purchases-js/pull/432/files#diff-05306f59947d840db7ba303affe21d7bbe6b996f0188079c771ef86093498005R47

src/ui/molecules/stripe-elements.svelte
```diff
@@ -194,120 +179,113 @@
...
+      elements = elementsInstance;
 
+      // At this point, the gatewayParams might have been updated so we attempt to set them.
+      const elementsConfiguration = gatewayParams.elements_configuration;
+      if (elementsConfiguration) {
+        await StripeService.updateElementsConfiguration(
+          elements,
+          elementsConfiguration,
+        );
+      }
```
polmiro, line 234: I am not 100% sure but given that `initStripe` is called `onMount` and being awaited the comment does not seem right to me. Even more so since you already added the effect on `gatewayParams` changes this seems superfluous.

```suggestion
```

polmiro (approved):
Left a comment about something I think is not necessary. Otherwise looks good to me.

PR #433, opened 2025-04-09
[WEB-2398] Do not present taxes to customers in `not_collecting` jurisdictions (FF)

vicfergar:
## Motivation / Description

The backend discards tax breakdown items from not-collecting jurisdictions. The SDK should omit the tax breakdown if the list of tax items is empty.

## Changes introduced

29068e4 - chore: Add tax collection disabled stories for the pricing table
b4cb62a - chore: Increase test coverage for the helper method `checkoutCalculateTax`
3332c87 - fix: Do not present taxes if `tax_breakdown` is empty

## Linear ticket (if any)

[WEB-2398](https://linear.app/revenuecat/issue/WEB-2398/[sdk]-do-not-present-taxes-to-customers-in-not-collecting)

### DIFF
--- a/src/ui/molecules/pricing-table.svelte
+++ b/src/ui/molecules/pricing-table.svelte
@@ -23,11 +23,17 @@
   }
 
   const translator: Writable<Translator> = getContext(translatorContextKey);
+
+  let showTaxBreakdown = $derived(
+    priceBreakdown.taxCollectionEnabled &&
+      priceBreakdown.taxBreakdown &&
+      priceBreakdown.taxBreakdown.length > 0,
+  );
 </script>
 
 {#snippet pricingTable()}
   <div class="rcb-pricing-table">
-    {#if priceBreakdown.taxCollectionEnabled}
+    {#if showTaxBreakdown}
       <div class="rcb-pricing-table-row">
         <div class="rcb-pricing-table-header">
           {$translator.translate(LocalizationKeys.PricingTotalExcludingTax)}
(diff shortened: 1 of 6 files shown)


nicfix (approved):

src/ui/purchases-ui.svelte
```diff
@@ -235,21 +234,24 @@
...
+    priceBreakdown.totalAmountInMicros = data.total_amount_in_micros;
+    priceBreakdown.taxAmountInMicros = data.tax_amount_in_micros;
     priceBreakdown.totalExcludingTaxInMicros =
-      taxCalculation.data.total_excluding_tax_in_micros;
-    priceBreakdown.taxBreakdown =
-      taxCalculation.data.pricing_phases.base.tax_breakdown;
+      data.total_excluding_tax_in_micros;
+    priceBreakdown.taxBreakdown = data.pricing_phases.base.tax_breakdown;
     priceBreakdown.pendingReason = null;
+    priceBreakdown.taxCollectionEnabled = true;
```
nicfix, line 251: Don't we risk to override the setting coming from the server?
If the priceBreakdown.taxCollectionEnabled is false due to the Feature flag we might overwrite that here and also above.

nicfix: If it doesn't maybe we can add a test for that.

vicfergar: It doesn't, since we only trigger a calculation when the FF and the setting are enabled. We already have a test for that

polmiro (approved):
Neat

src/ui/molecules/pricing-table.svelte
```diff
@@ -23,11 +23,17 @@
   }
 
   const translator: Writable<Translator> = getContext(translatorContextKey);
+
+  let showTaxBreakdown = $derived(
```
nicfix, line 27: Much cleaner, thanks for making it more explicit!

nicfix (approved):

PR #436, opened 2025-04-10
WEB-2370: Added E2E tests for the tax calculation behaviour

nicfix:
## Motivation / Description
We added tax calculation but we never integrated some tests.
Tests will be skipped now since we are not building with VITE_ALLOW_TAX_CALCULATION_FF="true" but they can be used in development.

### DIFF
--- a/.env.test
+++ b/.env.test
@@ -1,4 +1,3 @@
 VITE_RC_ENDPOINT=http://localhost:8000
 VITE_RC_ANALYTICS_ENDPOINT=http://localhost:8000
 VITE_ASSETS_ENDPOINT=https://da08ctfrofx1b.cloudfront.net
-VITE_ALLOW_TAX_CALCULATION_FF=true
--- a/.env.test
+++ b/.env.test
@@ -1,3 +1,4 @@
 VITE_RC_ENDPOINT=http://localhost:8000
 VITE_RC_ANALYTICS_ENDPOINT=http://localhost:8000
 VITE_ASSETS_ENDPOINT=https://da08ctfrofx1b.cloudfront.net
+VITE_ALLOW_TAX_CALCULATION_FF=true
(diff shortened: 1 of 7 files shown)


.env.test
```diff
@@ -1,3 +1,4 @@
 VITE_RC_ENDPOINT=http://localhost:8000
 VITE_RC_ANALYTICS_ENDPOINT=http://localhost:8000
-VITE_ASSETS_ENDPOINT=https://da08ctfrofx1b.cloudfront.net
\ No newline at end of file
+VITE_ASSETS_ENDPOINT=https://da08ctfrofx1b.cloudfront.net
+VITE_ALLOW_TAX_CALCULATION_FF=true
```
vicfergar, line 4: I think we can add this FF env var to the circle CI pipeline so these tests are run also in the CI

examples/webbilling-demo/src/tests/tax-calculation.test.ts
```diff
@@ -0,0 +1,172 @@
...
+    const cards = await getAllElementsByLocator(page, CARD_SELECTOR);
+    const targetCard = cards[0];
+
+    // Perform a purchase action
+    await startPurchaseFlow(targetCard);
+
+    await expect(page.getByText("Total excluding tax")).not.toBeVisible();
+  });
+
+  test("should display the tax breakdown at the checkout step", async ({
```
vicfergar, line 43: This will depend on the test runner location since the tax pre-calculation is based on the client IP.

.env.test
nicfix: The problem is when we build the sdk more than the tests themselves

vicfergar (approved):
Left some questions, but LG overall!

examples/webbilling-demo/src/tests/tax-calculation.test.ts
nicfix: Yeah, this specific test can be flakey, I don't have a better idea on how to test the initial calculation though.

PR #445, opened 2025-04-16
Add `isAnonymous` API

tonidero:
## Motivation / Description
Quick new API used in the hybrid SDKs. Could be useful directly in this SDK, so adding the logic here.


ajpallares (approved):
Makes sense!! 👍

PR #449, opened 2025-04-16
[WEB-2430] Update font sizes

polmiro:
## Motivation / Description

Updates font sizes per Figma spec by using a new `Typography` component.


src/ui/atoms/typography.svelte
```diff
@@ -0,0 +1,124 @@
...
+<script lang="ts">
+  const { children, size = "body-base" }: Props = $props();
+</script>
+
+<span class="rcb-typography rcb-typography-{size}">
+  {@render children?.()}
+</span>
+
+<style>
+  /* svelte-ignore css-unused-selector */
```
polmiro, line 31: 1. Remove original variables in favor of `Typography` component.
2. Review font boldness: `font-weight` and `font-family`.

src/ui/atoms/typography.svelte
```diff
@@ -0,0 +1,132 @@
...
+  .rcb-typography-heading-2xl {
+    font: var(--rc-text-heading2xl-mobile);
+  }
```
alfondotnet, line 34: I feel like we should just use `style` directly instead of having to define all these css classes

alfondotnet (approved):

src/ui/atoms/typography.svelte
polmiro: sounds good 👍🏻

polmiro: The change got a little tricky having to convert the namings "-heading-2xl-" -> "headingXl" and also with media queries in the middle. Since it's a step forward and we will make bigger strides soon with the design system I will keep it as it is.

PR #453, opened 2025-04-22
[WEB-2480] Button to component lib

james-od:
## Motivation / Description

See: https://github.com/RevenueCat/purchases-ui-js/pull/69

Following the duplication of the `Button` component into the component library in `-ui`, we can now import and reference it in all usages within the SDK then delete the now unused component.

**Question**: What is the best way to properly test this? I'm particularly thinking about the click handler since the visual aspect all seems fine. I've added onclick props to the storybooks and verified console logs, is there a more end-to-end way to validate?


polmiro:
The integration tests already give good coverage for this interaction and chromatic did not pick up any visual change. So I believe this is good to go!

If you want to be extra sure perhaps a smoke test locally to try out the entire flow if you have not done it already.

polmiro (approved):

nicfix (approved):

PR #461, opened 2025-05-01
[Draft][WEB-2510] Pass brandingAppearance to Buttons

james-od:
## Motivation / Description

Undetermined if we'll actually proceed with this, but there is an issue with regards to fallback styles with our web component approach.

In the Svelte App (SDK), we don't currently pass brandingAppearance - styles are inherited through the theme set at the top level and the UI Library Button Component just consumes these.

In the React App, since we render in the shadow DOM, we need to pass the brandingAppearance to create the styles via the Theme class. But if brandingAppearance is empty, it will need to create default styles. However if we do that, then when it's empty for the Svelte App then default styles will always be applied.

Solution A: Add an "isSvelte/isReact" prop

I worry about this being a bit of a red flag, I don't want every shared component to have a bunch of conditionals if possible

Solution B: Enforce always passing brandingAppearance, even in the Svelte App

This is awkward in terms of prop drilling and overall migration effort - but maybe sets a clearer differentiation of the library and enforces the brandingAppearance API?

This PR is a proposal/spike for solution B.

### DIFF
--- a/src/ui/layout/message-layout.svelte
+++ b/src/ui/layout/message-layout.svelte
@@ -4,13 +4,23 @@
   import ModalSection from "./modal-section.svelte";
   import RowLayout from "./row-layout.svelte";
   import Typography from "../atoms/typography.svelte";
+  import { getContext } from "svelte";
+  import { brandingContextKey } from "../constants";
+  import type { BrandingAppearance } from "../../entities/branding";
+  import { type Writable } from "svelte/store";
 
-  export let onDismiss: () => void;
-  export let title: string | null = null;
-  export let type: string;
-  export let closeButtonTitle: string = "Go back to app";
-  export let icon: (() => any) | null = null;
-  export let message;
+  let {
+    onDismiss,
+    title = null,
+    type,
+    closeButtonTitle = "Go back to app",
+    icon = null,
+    message,
+  } = $props();
+
+  const brandingAppearanceStore =
+    getContext<Writable<BrandingAppearance>>(brandingContextKey);
+  const brandingAppearance = $derived($brandingAppearanceStore);
 
   function handleClick() {
     onDismiss();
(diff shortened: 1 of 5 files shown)


james-od:
On the visual changes - these match the new button designs to use 16px font for buttons, which is a slight increase from the previous 14px. Discussed with Iga and we agree its better.

Some of the visual changes seem to have failed to render and are only a single pixel - I think its flakiness of some sort. I don't really know what to do about that, have we seen this before?

src/ui/layout/message-layout.svelte
```diff
@@ -31,8 +39,8 @@
   </div>
   <div class="message-layout-footer">
     <ModalFooter>
-      <Button onclick={handleClick} type="submit"
-        ><Typography size="body-small">{closeButtonTitle}</Typography></Button
```
antoniobg, line 35: Is the `<Typography>` removal on purpose?

antoniobg (approved):

antoniobg:
Looking at the stories, is it possible that the button text is now slightly larger?

james-od:
Thanks for the review @antoniobg!
Yep, mentioned here https://github.com/RevenueCat/purchases-js/pull/461#issuecomment-2915676881. Discussed with Iga and we think this works, the shared component controls the font size and currently thats fixed at `body-base` (16px), and thats why I removed the Typography component. We can make this configurable if we need to in future.

antoniobg:
> Thanks for the review @antoniobg!
> 
> Yep, mentioned here https://github.com/RevenueCat/purchases-js/pull/461#issuecomment-2915676881. Discussed with Iga and we think this size looks better, the shared component controls the font size and currently thats fixed at `body-base` (16px), and thats why I removed the Typography component. We can make this configurable if we need to in future. 

Oops, sorry. I completely missed the comment

antoniobg (approved):

PR #463, opened 2025-05-02
Prioritize apple_pay in payment list if available

polmiro:
## Motivation / Description

This update prioritizes `apple_pay` and `google_pay` in the list of payment methods. This only has an effect if a wallet is is supported and displayed to the current viewer. Also updates the `defaultCollapsed` to `true` so that the `card` payment method does not get pre-selected and expanded when a wallet is available.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@revenuecat/purchases-js",
-  "version": "1.2.0",
+  "version": "1.2.1",
   "lockfileVersion": 3,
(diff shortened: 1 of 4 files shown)


antoniobg (approved):

vicfergar (approved):

polmiro:
In the end, I removed the "shortcut" given the feedback.

polmiro:
Closing in favor of Stripe Express Checkout Button

PR #470, opened 2025-05-05
Fix loading spinner positional glitch

polmiro:
## Motivation / Description

Fixes glitch with the loading spinner.

There are two different spinners being rendered:
1. on the initial load, waiting for the start of the checkout request to go through.
2. on the payment entry page while stripe is loading, or the payment is processing.

The second spinner was not well positioned. This is related to the stripe form elements, which are kept on the screen with "visibility: hidden" and not "display: none". If I recall well, it is to deal with the animations that are provided by Stripe smoothly.

This PR uses "visibility: hidden" for the ".rc-checkout-form-container" but uses "display:none" for the other parts that are not related to Stripe Elements. Relies on "absolute" positioning for the loading so that it will be centered on top of the "invsibile" Stripe Elements.


src/ui/layout/section-layout.svelte
```diff
@@ -76,7 +76,6 @@
 
   @container layout-query-container (width >= 768px) {
     .layout-wrapper {
-      min-height: 100vh;
```
polmiro, line 79: This is an unrelated problem I noticed in production. It is creating extra space in the desktop view that causes the scrollbars to appear unnecessarily.

src/ui/pages/payment-entry-page.svelte
```diff
@@ -48,6 +48,8 @@
   import StripeElementsComponent from "../molecules/stripe-elements.svelte";
   import PriceUpdateInfo from "../molecules/price-update-info.svelte";
 
+  type View = "loading" | "form" | "message";
```
alfondotnet, line 51:
```suggestion
  type View = "loading" | "form" | "error";
```

instead? but no big deal

polmiro: I like it, on it.

antoniobg (approved):

PR #472, opened 2025-05-06
Render wordmark when possible

alfondotnet:
## Motivation / Description

If the `/branding` endpoint comes with a wordmark, we want to render that first


src/ui/molecules/branding-header.svelte
```diff
@@ -41,11 +49,15 @@
 
   <div class="rcb-header">
     <div class="rcb-title">
-      {#if appIcon !== null && webpIcon !== null}
+      {#if wordmarkSrc !== null}
```
vicfergar, line 52:
```suggestion
      {#if wordmarkSrc !== null && wordmarkSrcWebp !== null}
```

PR #476, opened 2025-05-06
Support branded fonts in typography

elenaperezrioja:
## Motivation / Description
Typography accepts a new prop "branded" that will make the text use the font provided in brandFontURL in the /branding request.

## Changes introduced
- When brandFrontURL is non null, load the fontFace
- Define new css variables for the font family of branded texts. We need to return the non branded version by default if brand config is not available
- Typography uses that css variable when "branded" prop is passed

### DIFF
--- a/src/ui/atoms/typography.svelte
+++ b/src/ui/atoms/typography.svelte
@@ -16,14 +16,19 @@
   type Props = {
     size?: Size;
     children?: Snippet;
+    branded?: boolean;
   };
 </script>
 
 <script lang="ts">
-  const { children, size = "body-base" }: Props = $props();
+  const { children, size = "body-base", branded = false }: Props = $props();
 </script>
 
-<span class="rcb-typography rcb-typography-{size}">
+<span
+  class="rcb-typography rcb-typography-{size} {branded
+    ? 'rcb-typography-branded'
+    : ''}"
+>
   {@render children?.()}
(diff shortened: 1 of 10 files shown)


src/ui/atoms/typography.svelte
```diff
@@ -16,23 +16,39 @@
...
+    font-weight: normal;
+    font-style: normal;
+  }
+
   .rcb-typography {
     display: inline-block;
     line-height: normal;
   }
 
+  span.rcb-typography-branded {
```
elenaperezrioja, line 48: Using span. here to improve specificity, as "font" defines font-family too and we want to override that.

src/ui/molecules/product-header.svelte
```diff
@@ -23,7 +23,7 @@
     </div>
   {/if}
   <div class="rcb-product-title">
-    <Typography size="heading-lg">
+    <Typography size="heading-lg" branded>
```
elenaperezrioja, line 26: I talked to @FayeKeegan and ideally, it should be 2xl and font-weight:300. But that doesn't look great in the general case, so we might need to add more settings to branding in addition to the font.
Screenshot 2025-05-06 at 16 14 05
Screenshot 2025-05-06 at 16 12 57

polmiro (approved):

Issue #415, opened 2025-03-27
Body styles not restored correctly


jkk:
When closing the Purchases UI, several body styles are set to "auto", which is not necessarily what they were set to prior to opening the Purchase UI. The inline "auto" style overwrites styles defined via classes, and also clobbers any inline styles that were previous set.

To be correct, it should save the inline style values in onMount and then restore them after.

https://github.com/RevenueCat/purchases-js/blob/8c32ef98cb76932f8a8bede7068d2a2551f0e459/src/ui/purchases-ui.svelte#L102-L108

gezquinndesign:
Also seeing this on `1.2.1`

gezquinndesign:
This should probably be prioritised as it leads to a very bad experience on our react-native-web app.  We don't have any styles applied on any of those elements by default so when they are applied, our app is rendered unusable. I've submitted a PR.

alfondotnet:
Fixed by https://github.com/RevenueCat/purchases-js/commit/ec31cfe4ed1c0d38c2acd4ad317d33df0ac2b76f, it  will got out in the next release train. Thank you so much @gezquinndesign for your contribution!

PR #484, opened 2025-05-12
fix: save and apply original inline styles on close

gezquinndesign:
## Motivation / Description
When closing the Purchases UI, several body styles are set to "auto", which is not necessarily what they were set to prior to opening the Purchase UI. The inline "auto" style overwrites styles defined via classes, and also clobbers any inline styles that were previous set.

To be correct, it should save the inline style values in onMount and then restore them after.

## Changes introduced
Original html and body styles are saved on mount and restored on destroy

## Linear ticket (if any)
Issue #415


src/ui/purchases-ui.svelte
```diff
@@ -83,6 +83,11 @@
   let operationSessionId: string | null = $state(null);
   let gatewayParams: GatewayParams = $state({});
 
+  // For storing original styles
+  let originalHtmlHeight: string | null = $state(null);
+  let originalHtmlOverflow: string | null = $state(null);
+  let originalBodyHeight: string | null = $state(null);
```
alfondotnet, line 89:
```suggestion
  let originalHtmlHeight: string | null = $state("");
  let originalHtmlOverflow: string | null = $state("");
  let originalBodyHeight: string | null = $state("");
```

src/ui/purchases-ui.svelte
```diff
@@ -102,9 +113,10 @@
 
   onDestroy(() => {
     if (!isInElement) {
-      document.documentElement.style.height = "auto";
-      document.body.style.height = "auto";
-      document.documentElement.style.overflow = "auto";
+      // Restore original styles
+      document.documentElement.style.height = originalHtmlHeight || "";
+      document.body.style.height = originalBodyHeight || "";
+      document.documentElement.style.overflow = originalHtmlOverflow || "";
```
alfondotnet, line 119: This way you can probaby just do

```suggestion
      document.documentElement.style.height = originalHtmlHeight;
      document.body.style.height = originalBodyHeight;
      document.documentElement.style.overflow = originalHtmlOverflow;
```

?

alfondotnet:
Thanks for the PR @gezquinndesign! makes sense to me

gezquinndesign (commented):
Cool, applied your suggestions @alfondotnet, but then decided it might be better to remove the styles completely if they weren't originally defined to avoid setting the properties explicitly to empty strings.

alfondotnet (approved):

alfondotnet:
Thanks @gezquinndesign !!

@tonidero Since we have CI disabled for forks, builds don't run on external PRs, I changed the base to verify before merging to main, but maybe we have to talk about something here

PR #485, opened 2025-05-13
[DRAFT] Add paddle web paywall link config

rogersole:


nicfix (commented):
No issue on this one, I think this is the best approach since it makes it transparent to the SDK. Paddle becomes just another supported store.

I would just add some E2E tests to verify that the basic methods work (get offerings etc) also when the paddle api key is used.

PR #486, opened 2025-05-13
Remove unused stripe styles

polmiro:
## Motivation / Description

Noticed some rules are not used anymore. This should not trigger any Chromatic changes.

### DIFF
--- a/src/stripe/stripe-service.ts
+++ b/src/stripe/stripe-service.ts
@@ -155,36 +155,6 @@ export class StripeService {
             ".Input--invalid": {
               boxShadow: "none",
             },
-            ".TermsText": {
-              fontSize: textStyles.captionDefault[viewport].fontSize,
-              lineHeight: textStyles.captionDefault[viewport].lineHeight,
-            },
-            ".Tab": {
-              boxShadow: "none",
-              backgroundColor: "transparent",
-              color: customColors["grey-text-light"],
-              border: `1px solid ${customColors["grey-ui-dark"]}`,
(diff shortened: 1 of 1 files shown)


src/stripe/stripe-service.ts
```diff
@@ -155,36 +155,6 @@ export class StripeService {
             ".Input--invalid": {
               boxShadow: "none",
             },
-            ".TermsText": {
-              fontSize: textStyles.captionDefault[viewport].fontSize,
-              lineHeight: textStyles.captionDefault[viewport].lineHeight,
-            },
-            ".Tab": {
```
nicfix, line 162: Aren't Tabs used when we have more than one payment method?

polmiro: With the addition of the `ExpressCheckoutElement` we are relegated to only display the `card` method, at least as of now. But if we prefer to leave all these for a potential future I can just close the PR.

nicfix: Sorry I missed this bit, if we are 100% sure that other payment methods won't pop up it's fine (and actually better) to remove this unused CSS.

nicfix (approved):

src/stripe/stripe-service.ts
polmiro: Yes, we are only using "card" as payment method, and the wallets are deactivated implicitly in the `PaymentElement` given that we are using `ExpressCheckoutElement` in the same Stripe element group.

PR #489, opened 2025-05-14
feat: Tax support (private beta)

vicfergar:
## Motivation / Description

Remove the hardcoded feature flag that enables access to the private beta for taxes.


examples/webbilling-demo/src/tests/helpers/integration-test.ts
```diff
@@ -5,8 +5,8 @@ import { getEmailFromUserId } from "./test-helpers";
 export const ALLOW_PAYWALLS_TESTS =
   process.env.VITE_ALLOW_PAYWALLS_TESTS === "true";
 
-export const ALLOW_TAX_CALCULATION_FF =
-  process.env.VITE_ALLOW_TAX_CALCULATION_FF === "true";
+export const VITE_RC_TAX_E2E_API_KEY =
```
nicfix, line 8: The name of this variable is a bit confusing, do we expect to have the API key in this env var?
Or do we just expect its value to be "true"?

vicfergar: Fixed!

nicfix (approved):

PR #492, opened 2025-05-14
Fix flickering button when taxes are being refreshed

polmiro:
## Motivation / Description

Fixes a flickering issue on the button when taxes are being calculated.

### DIFF
--- a/examples/webbilling-demo/src/tests/tax-calculation.test.ts
+++ b/examples/webbilling-demo/src/tests/tax-calculation.test.ts
@@ -11,6 +11,7 @@ import {
   getStripePaymentFrame,
   confirmPayButtonDisabled,
   confirmPaymentError,
+  enterSecurityCode,
 } from "./helpers/test-helpers";
 import {
   integrationTest,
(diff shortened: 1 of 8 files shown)


examples/webbilling-demo/src/tests/tax-calculation.test.ts
```diff
@@ -198,54 +199,89 @@ integrationTest.describe("Tax calculation", () => {
...
+      const hasCompleted = await Promise.race([
+        italyTaxCalculationResponsePromise.then(() => true),
+        new Promise((resolve) => setTimeout(() => resolve(false), 0)),
+      ]);
+
+      if (hasCompleted) {
+        expect("Italy tax calculation should not have finished").toBeFalsy();
+      } else {
+        expect("Italy tax calculation finished").toBeTruthy();
+      }
```
vicfergar, line 284: This part seems hard to read for me, could we just fail if `firstRequestHasCompleted`?
```suggestion
      const firstRequestHasCompleted = await Promise.race([
        italyTaxCalculationResponsePromise.then(() => true),
        new Promise((resolve) => setTimeout(() => resolve(false), 0)),
      ]);
      
      if (firstRequestHasCompleted) {
        fail("Italy tax calculation should not have finished");
      }
```

polmiro: Looking into this with the new change altogether now.

src/ui/pages/payment-entry-page.svelte
```diff
@@ -292,13 +293,28 @@
     if (!isEmailComplete || !isPaymentInfoComplete) {
       return;
     }
+    if (processing) return;
+
+    if (taxCalculationStatus !== "loading") {
+      previousTaxCalculationStatus = taxCalculationStatus;
+      taxCalculationStatus = "loading";
```
polmiro, line 300: This one.

src/ui/pages/payment-entry-page.svelte
```diff
@@ -292,13 +293,28 @@
...
         .then(() => recalculateTaxes(signal))
         .then(() => signal.throwIfAborted())
-        .catch(handleErrors);
+        .catch((e) => {
+          if (!signal.aborted) {
+            handleErrors(e);
+          }
+        })
+        .finally(() => {
+          if (taxCalculationStatus === "loading" && !signal.aborted) {
```
polmiro, line 314: Plus this.

PR #493, opened 2025-05-14
chore: add asserts for configure

alfondotnet:
## Motivation / Description

If testing the configure with a JavaScript app, and not providing an appUserId currently fails downstream, will be nice to provide some helpful errors

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -214,6 +214,21 @@ export class Purchases {
           "Creating and returning new instance.",
       );
     }
+
+    if (!apiKey) {
+      throw new PurchasesError(
+        ErrorCode.ConfigurationError,
+        "API key is required. Please provide a valid RevenueCat API key.",
+      );
+    }
+
+    if (!appUserId) {
+      throw new PurchasesError(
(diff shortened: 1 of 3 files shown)


src/main.ts
```diff
@@ -214,6 +214,21 @@ export class Purchases {
...
+        "API key is required. Please provide a valid RevenueCat API key.",
+      );
+    }
+
+    if (!appUserId) {
+      throw new PurchasesError(
+        ErrorCode.ConfigurationError,
+        "App user ID is required. Please provide a valid user identifier.",
+      );
+    }
```
tonidero, line 230: Hmm shouldn’t the validators a couple lines below be catching these? https://github.com/RevenueCat/purchases-js/blob/main/src/helpers/configuration-validators.ts

Though looking at it, maybe an undefined value could go through the app user id validator… In that case maybe it’s worth moving this check there?

alfondotnet: sorry still not ready for review, should I have put it up as a draft, but makes sense

tonidero (approved):
I think this makes sense!

PR #498, opened 2025-05-19
Mock taxes endpoint when rate limited

polmiro:
## Motivation / Description

E2E will run by default with mocked taxes endpoint and real taxes endpoint. But an environment variable allows to skipp real tax endpoints.


examples/webbilling-demo/src/tests/helpers/fixtures.ts
```diff
@@ -40,3 +42,220 @@ export const ITALY_CUSTOMER_DETAILS = {
 export const SPAIN_CUSTOMER_DETAILS = {
   countryCode: "ES",
 };
+
+export const SPAIN_TAX_RESPONSE: RouteFulfillOptions = {
+  status: 200,
+  contentType: "application/json",
+  body: JSON.stringify({
+    mocked: true,
```
polmiro, line 50: This parameter is just to confidently be able to spot if the response in playwright was a mocked one or not.

vicfergar (approved):

PR #499, opened 2025-05-20
Fix tax calculation mistakenly launched

polmiro:
## Motivation / Description

Condition got lost at some point, probably when adding the check on loading to then revert it to the previous state if something when wrong. Tested manually, automated test will be added separately.


nicfix (approved):

james-od (approved):
Confirmed fixes the issue locally 👍

PR #504, opened 2025-05-22
release snapshots upon merge to main

alfondotnet:
## Motivation / Description

Creates a snapshot release upon merging to main


fastlane/Fastfile
```diff
@@ -78,6 +79,71 @@ lane :release do |options|
...
+  Dir.chdir(get_root_folder) do
+    sh("npm version #{snapshot_version} --no-git-tag-version")
+    sh("npm publish --tag snapshot")
+    sh("npm version #{stable_version} --no-git-tag-version")
+  end
+
+  add_git_tag(tag: snapshot_version, message: "Snapshot #{snapshot_version}")
+  push_to_git_remote(remote: "origin", tags: true)
+
+  create_github_release(
```
tonidero, line 98: Hmm I don't know if we want to create github releases for snapshots... It could get really noisy. Wdyt?

fastlane/Fastfile
```diff
@@ -78,6 +79,71 @@ lane :release do |options|
...
+
+  UI.message("Creating snapshot release #{snapshot_version}...")
+
+  Dir.chdir(get_root_folder) do
+    sh("npm version #{snapshot_version} --no-git-tag-version")
+    sh("npm publish --tag snapshot")
+    sh("npm version #{stable_version} --no-git-tag-version")
+  end
+
+  add_git_tag(tag: snapshot_version, message: "Snapshot #{snapshot_version}")
```
tonidero, line 95: Also, how about prefixing all snapshot versions with something? Maybe starting with `snapshot-X.Y.Z`? Mostly so they are grouped together in github and the stable releases stay at the top.

fastlane/Fastfile
```diff
@@ -78,6 +79,71 @@ lane :release do |options|
...
+      sh("npm", "install", "--save-exact", "@revenuecat/purchases-js@#{version_number}")
+
+      if system("git diff --quiet")
+        UI.message("No dependency update needed; skipping PR")
+        next
+      end
+
+      sh("git", "add", "package.json", "package-lock.json")
+      sh("git", "commit", "-m", "Update purchases-js to #{version_number}")
+      sh("git", "push", "--set-upstream", "origin", branch_name)
```
tonidero, line 130: This can work, but I was wondering, is it possible to trigger a workflow in the WPL repo from an API call? That way the automations for updating the version in that repo stay in that repo, and we just trigger that from here. If not, or it's more complex, I'm ok going this route for now.

fastlane/Fastfile
```diff
@@ -78,6 +79,71 @@ lane :release do |options|
...
+desc "Creates GitHub snapshot pre-release and publishes with the 'snapshot' tag"
+lane :snapshot_release do
+  stable_version   = current_version_number
+  snapshot_version = "#{stable_version}-SNAPSHOT.#{Time.now.strftime('%Y%m%d%H%M')}"
+
+  UI.message("Creating snapshot release #{snapshot_version}...")
+
+  Dir.chdir(get_root_folder) do
+    sh("npm version #{snapshot_version} --no-git-tag-version")
+    sh("npm publish --tag snapshot")
```
tonidero, line 91: I'm wondering if it would be better to change the library for these snapshots, so we keep the main library clean of all these versions, instead of changing the tag... Mostly so npm stays clean of having a ton of snapshots... But I think it's ok to go with this for now and we can always move to a separate repo in the future, so not a blocker.

tonidero (commented):
This looks good, but just left some comments. Lmk what you think!

fastlane/Fastfile
alfondotnet: We'd have to have some infra there for this, but I'll note it down, and if I test this and works, maybe we can get it out of the door

fastlane/Fastfile
```diff
@@ -78,6 +79,42 @@ lane :release do |options|
   github_release(version: version_number)
 end
 
+desc "Creates snapshot pre-release and publishes with the 'snapshot' tag"
+lane :snapshot_release do
+  stable_version   = current_version_number
```
tonidero, line 84: One thing we do in Android + iOS is to bump the version number in main to include it's a snapshot: https://github.com/RevenueCat/purchases-android/blob/178691123e9638616f5eff6f17bbb1cf46f2808e/fastlane/Fastfile#L165-L173

This could have some advantages here... Mainly, I believe with this code we would be creating snapshots of the last stable version, which doesn't make sense since snapshots usually refer to a yet "unreleased" version.

So I think we could potentially bump the version in `.version` to the next minor + `-SNAPSHOT` as we do in android/ios or at least calculate the next minor for the current version, instead of using the last released stable version.

tonidero (commented):
Looking great! Just have one comment about the version of snapshots. We should be using a version still unreleased I think, not the version that already exists as stable. Lmk what you think!

fastlane/Fastfile
alfondotnet: not sure about this, in practice you'll have multiple PRs merged between released versions, right?

Thinking here would be having some kind of a reference to the last version and then some suffix indicating that it's a different change ..

Agree with you that it could be confusing, but I think incrementing a minor in each PR merge can make things worse

PR #506, opened 2025-05-26
Add tracking for tax calculations

polmiro:
## Motivation / Description

Add tracking for Tax Calculations.


src/ui/pages/payment-entry-page.svelte
```diff
@@ -180,6 +181,12 @@
         signal,
       )
       .then((taxCalculation) => {
+        const event = createCheckoutPaymentTaxCalculationEvent({
```
vicfergar, line 184: Should we add a comment indicating that we intentionally track these events even if the calculation is aborted?

vicfergar (approved):
Minor suggestion. LG!

PR #490, opened 2025-05-14
[WEB-2475] Better errors for Stripe setup on sandbox

polmiro:
## Motivation / Description

Adds more specific error messages for errors setting up taxes with Stripe.

PR #507, opened 2025-05-26
Re-enable tax calculation e2e tests

polmiro:
## Motivation / Description

Re-enables tax e2e tests.


nicfix (changes requested):
I think we should leave this check for now or replace it with what you proposed here:
https://github.com/RevenueCat/purchases-js/pull/508/files

Having a way to skip tax tests is needed since we can fall into the quota exceeded easily.

polmiro:
Used `git bisect` pushing a CI build and it seems that this PR is the culprit of the tax integration tests failing in CI. Just leaving the note [here](https://github.com/RevenueCat/purchases-js/pull/490/files).

examples/webbilling-demo/src/tests/tax-calculation.test.ts
```diff
@@ -80,6 +83,15 @@ const mockTaxCalculationRequest = async (
           await page.route(TAX_ROUTE_PATH, async (route) => {
             route.abort();
           });
+        } else {
+          // Fail the test if the rate limit is reached
+          await page.route(TAX_ROUTE_PATH, async (route) => {
+            const response = await route.fetch();
+            const responseBody = await response.json();
+            if (responseBody["failed_reason"] === "rate_limit_exceeded") {
+              throw new Error("Stripe Tax Calculation API rate limit reached.");
```
vicfergar, line 92: ❤️

vicfergar (approved):

polmiro:
Done!

Current main:
- 69 skipped
- 110 passed
This PR: 
- 57 skipped
- 121 passed

examples/webbilling-demo/src/tests/tax-calculation.test.ts
```diff
@@ -74,12 +77,27 @@ const mockTaxCalculationRequest = async (
         To enable, set VITE_SKIP_TAX_REAL_TESTS_UNTIL=2025-02-21 in the environment variables.`,
       );
 
+      integrationTest.skip(
+        ({ browserName }) => !mockMode && browserName !== "chromium",
+        "Real tax calculation tests only run in Chromium",
```
nicfix, line 82: good one, doesn't make sense to repeat them for each browser

nicfix (approved):

PR #508, opened 2025-05-26
Skip running tax calculation tests based on date

polmiro:
## Motivation / Description

Re-enables tax e2e tests and adds a new env to allow deactivating them based on date to circumvent possible rate limit issues.

### DIFF
--- a/examples/webbilling-demo/src/tests/purchase-flow.test.ts
+++ b/examples/webbilling-demo/src/tests/purchase-flow.test.ts
@@ -133,17 +133,20 @@ test.describe("Purchase error paths", () => {
     await confirmStripeEmailError(page, "Your email address is invalid.");
   });
 
-  integrationTest("Email deliverability errors", async ({ page, userId }) => {
-    page = await navigateToLandingUrl(page, userId);
-    const email = `${userId}@revenueci.comm`;
+  integrationTest.fixme(
+    "Email deliverability errors",
+    async ({ page, userId }) => {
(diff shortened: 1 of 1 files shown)


examples/webbilling-demo/src/tests/helpers/integration-test.ts
```diff
@@ -5,7 +5,43 @@ import { getEmailFromUserId } from "./test-helpers";
...
+  if (!skipUntilDate) return false;
+  try {
+    // Validate the format is yyyy-mm-dd
+    const dateFormatRegex = /^\d{4}-\d{2}-\d{2}$/;
+    if (!dateFormatRegex.test(skipUntilDate)) {
+      console.error(
+        "Invalid date format for VITE_SKIP_TAXES_TESTS_UNTIL. Expected format is 'yyyy-mm-dd'",
+      );
+      return false;
+    }
```
nicfix, line 19: Nit: Maybe this check is not needed given that we are passing the string to the Date object and checking if we created a valid object.

But I might not see the edge case here.

nicfix (approved):

examples/webbilling-demo/src/tests/helpers/integration-test.ts
polmiro: It accepts a couple of different formats but I decided to make it explicit on "yyyy-mm-dd" to avoid confusion (epoch, iso timestamp, iso date, etc).

PR #514, opened 2025-05-28
WEB-2643: Do not show recurrent payment message for one-time purchases

nicfix:
## Motivation / Description
When purchasing a subscription we show a message explaining the user that following renewals will be charged to the card they insert in the payment step.

We should not show this message for one-time (consumable or not) purchases, this PR removes that message when a consumable is being purchased.

## Changes introduced
* Not showing the recurrent payment text when the purchase is not a subscription.
* Added E2E tests to verify that.

### DIFF
--- a/examples/webbilling-demo/src/tests/helpers/test-helpers.ts
+++ b/examples/webbilling-demo/src/tests/helpers/test-helpers.ts
@@ -22,10 +22,11 @@ export async function performPurchase(
   page: Page,
   card: Locator,
   email: string,
+  isSubscription: boolean = true,
 ) {
   await startPurchaseFlow(card);
   await enterEmail(page, email);
-  await enterCreditCardDetails(page, "4242 4242 4242 4242");
+  await enterCreditCardDetails(page, "4242 4242 4242 4242", {}, isSubscription);
   await clickPayButton(page);
   await confirmPaymentComplete(page);
 }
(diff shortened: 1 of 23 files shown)


examples/webbilling-demo/src/tests/helpers/test-helpers.ts
```diff
@@ -179,8 +181,17 @@ export async function enterCreditCardDetails(
...
   await expect(checkoutTitle).toBeVisible();
+
+  const allowingCardUsageText = page.getByText(
+    /By providing your card information you allow/,
+  );
+  if (isSubscription) {
+    await expect(allowingCardUsageText).toBeVisible();
+  } else {
+    await expect(allowingCardUsageText).not.toBeVisible();
+  }
```
vicfergar, line 193: Checking this in every single E2E test seems excessive to me. Since you’ve added a story for each product type, I think that should be sufficient.

If we still want to verify this in E2E tests, I’d suggest introducing helper methods like `confirmTermsVisible(page: Page)` and `confirmTermsNotVisible(page: Page)`, and only checking this in the following tests:
- Purchase a subscription product
- Purchase a consumable product
- Purchase a non-consumable product

nicfix: Removing the tests

vicfergar (approved):

PR #516, opened 2025-06-03
Add `paddle` store

fire-at-will:
## Changes introduced
Adds `paddle` to the list of supported store types and adds a test to test the parsing

I don't _think_ this is blocked on this PHC PR but am not 100% sure: https://github.com/RevenueCat/purchases-hybrid-common/pull/1150

### DIFF
--- a/src/tests/entities/customer-info.test.ts
+++ b/src/tests/entities/customer-info.test.ts
@@ -441,4 +441,121 @@ describe("customer info parsing", () => {
     const customerInfo = toCustomerInfo(subscriberResponse);
     expect(customerInfo).toEqual(expectedCustomerInfo);
   });
+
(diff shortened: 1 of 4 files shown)


src/tests/entities/customer-info.test.ts
```diff
@@ -441,4 +441,121 @@ describe("customer info parsing", () => {
     const customerInfo = toCustomerInfo(subscriberResponse);
     expect(customerInfo).toEqual(expectedCustomerInfo);
   });
+
+  test("subscriber info with paddle subscription purchase is parsed correctly", () => {
```
tonidero, line 445: TBH, not sure if we need a test for every store type... but I guess it doesn't hurt

tonidero (approved):
Looks good!

PR #530, opened 2025-06-17
[WEB-2744] Add object-based configuration for Purchases SDK initialization

antoniobg:
## Motivation / Description

- We added a new way to configure the Purchases SDK using an object. This makes it easier to pass configuration parameters, especially when dealing with many optional settings.

## Changes introduced

- Introduced a new `PurchasesConfig` interface for object-based configuration.
- Updated the `configure` method to accept either the new configuration object or the previous positional parameters.
- Made changes to the API documentation to reflect these updates.
- Added tests for the new object-based configuration to ensure it works as expected.

## Linear ticket (if any)

- https://linear.app/revenuecat/issue/WEB-2744/improve-sdk-configuration-by-using-an-object

### DIFF
--- a/api-report/purchases-js.api.md
+++ b/api-report/purchases-js.api.md
@@ -353,6 +353,7 @@ export class Purchases {
     changeUser(newAppUserId: string): Promise<CustomerInfo>;
     close(): void;
     static configure(config: PurchasesConfig): Purchases;
+    // @deprecated
     static configure(apiKey: string, appUserId: string, httpConfig?: HttpConfig, flags?: FlagsConfig): Purchases;
     static generateRevenueCatAnonymousAppUserId(): string;
     getAppUserId(): string;
--- a/api-report/purchases-js.api.md
+++ b/api-report/purchases-js.api.md
@@ -352,6 +352,7 @@ export interface PurchaseResult {
 export class Purchases {
     changeUser(newAppUserId: string): Promise<CustomerInfo>;
     close(): void;
+    static configure(config: PurchasesConfig): Purchases;
     static configure(apiKey: string, appUserId: string, httpConfig?: HttpConfig, flags?: FlagsConfig): Purchases;
     static generateRevenueCatAnonymousAppUserId(): string;
     getAppUserId(): string;
@@ -374,6 +375,14 @@ export class Purchases {
     }): Promise<void>;
     static setLogLevel(logLevel: LogLevel): void;
     static setPlatformInfo(platformInfo: PlatformInfo): void;
+}
+
+// @public
+export interface PurchasesConfig {
+    apiKey: string;
+    appUserId: string;
+    flags?: FlagsConfig;
+    httpConfig?: HttpConfig;
 }
 
 // @public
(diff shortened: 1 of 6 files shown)


src/main.ts
```diff
@@ -207,6 +259,13 @@ export class Purchases {
   static configure(
     apiKey: string,
     appUserId: string,
+    httpConfig?: HttpConfig,
+    flags?: FlagsConfig,
+  ): Purchases;
+
+  static configure(
```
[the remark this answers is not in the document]
antoniobg: Yes, if we want to keep the same name for both the old and the new implementation, we need to do it this way. At least, I haven't found any way to use the same function name with 2 completely different implementations.

api-report/purchases-js.api.md
```diff
@@ -376,6 +378,14 @@ export class Purchases {
     static setPlatformInfo(platformInfo: PlatformInfo): void;
 }
 
+// @public
+export interface PurchasesConfig {
+    apiKey: string;
+    appUserId: string;
+    flags?: FlagsConfig;
```
tonidero, line 385: I do wonder now whether we would like to flatten some of these properties into `PurchasesConfig`... But I think it's ok to keep it like this.

src/main.ts
```diff
@@ -115,6 +115,49 @@ export type { PlatformInfo } from "./entities/platform-info";
...
+ *   "your_api_key",
+ *   "user_123",
+ *   { additionalHeaders: { "Custom-Header": "value" } },
+ *   { autoCollectUTMAsMetadata: true }
+ * );
+ * ```
+ *
+ * @public
+ */
+export interface PurchasesConfig {
```
tonidero, line 142: Should we extract this to a separate file? Mostly so it doesn't grow the `main.ts` even more... If we do end up extracting it, we should make sure to export it from `main.ts` though, so it's available by consumers.

src/main.ts
```diff
@@ -216,11 +274,50 @@ export class Purchases {
...
-    validateAdditionalHeaders(httpConfig.additionalHeaders);
-    Purchases.instance = new Purchases(apiKey, appUserId, httpConfig, flags);
+
+    let finalApiKey: string;
+    let finalAppUserId: string;
+    let finalHttpConfig: HttpConfig;
+    let finalFlags: FlagsConfig;
+
+    // Check if first argument is a configuration object
+    if (typeof configOrApiKey === "object" && configOrApiKey !== null) {
```
tonidero, line 284: Hmm I'm wondering if we should convert the old parameters into a `PurchasesConfig` first, so the rest of the configuration process can be done just with a `PurchasesConfig`, and extract that into a `configureInternal` or something like that... Wdyt? Not a blocker though, since we probably also want to do some runtime validation as you're doing here now.

tonidero (approved):
I like it! Thanks for adding this!

api-report/purchases-js.api.md
tonidero: Yeah... it depens on how many settings we would need... Let's keep it as is for now.

tonidero (approved):
Looks great!

PR #531, opened 2025-06-18
Internally track embedded usages

nicfix:
## Motivation / Description
We want to hide the close button in some specific scenarios internally.
This PR extends the rcSource flag to distinguish between the WPB scenario and any other scenario.

## Changes introduced
* Added support for a new 'embedded' rcSource value
* Added a test for the new behaviour
* Added support in the example app

### DIFF
--- a/examples/webbilling-demo/src/util/PurchasesLoader.ts
+++ b/examples/webbilling-demo/src/util/PurchasesLoader.ts
@@ -2,6 +2,7 @@ import type { CustomerInfo, Offering } from "@revenuecat/purchases-js";
 import { LogLevel, Purchases } from "@revenuecat/purchases-js";
 import type { LoaderFunction } from "react-router-dom";
 import { redirect, useLoaderData } from "react-router-dom";
+import type { FlagsConfig } from "../../../../src/entities/flags-config.ts";
 
 declare global {
   interface Window {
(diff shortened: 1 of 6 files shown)


examples/webbilling-demo/src/util/PurchasesLoader.ts
```diff
@@ -2,6 +2,7 @@ import type { CustomerInfo, Offering } from "@revenuecat/purchases-js";
 import { LogLevel, Purchases } from "@revenuecat/purchases-js";
 import type { LoaderFunction } from "react-router-dom";
 import { redirect, useLoaderData } from "react-router-dom";
+import type { FlagsConfig } from "../../../../src/entities/flags-config.ts";
```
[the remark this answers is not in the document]
nicfix: Fixed

james-od (approved):
Went through the slack discussion and [prior PR](https://github.com/RevenueCat/purchases-js/pull/475/files), tests and main files all make sense to me.
I don't fully understand the changes in the billing demo though, why we don't extend the FlagsConfig type?

src/main.ts
```diff
@@ -622,23 +622,25 @@ export class Purchases {
...
-                unmount(component);
-              }
-
-              certainHTMLTarget.innerHTML = "";
-
-              Logger.debugLog("Purchase cancelled by user");
-              reject(new PurchasesError(ErrorCode.UserCancelledError));
-            };
+      const shouldPassOnCloseBehaviour =
+        this._flags.rcSource === "app" || this._flags.rcSource === "embedded";
```
alfondotnet, line 626:
```suggestion
      const shouldPassOnCloseBehaviour =
        ["app", "embedded"].includes(this._flags.rcSource);
```

And maybe put the array up as a more visible constant?

nicfix: Done

PR #542, opened 2025-07-07
preload stripe again

alfondotnet:
There has been some loading time regression of Stripe Elements if it doesn't preload assets, so for now we'll be including Stripe SDK in our main bundle and preloading the assets upon initialization, until we are forced to chunk again (at which point we might offer a configuration parameter)


antoniobg (approved):

RohovDmytro:
Can we consider disabling this preloading when using `react-native-purchases` on web?

psnet:
Is there are any updates?

PR #560, opened 2025-07-31
Support test store

tonidero:
## Motivation / Description
This PR adds support for a new test store that can be used to test the RC integration before actually having to setup the stores.
When a user tries to purchase a product obtained through a `test_` api key, instead of going through the existing flow, we will display a simple modal with some product info, and the developer can purchase it using the test store. This can be useful to test the RC integration without actually having to setup Web billing. Can be specially useful for vibe coding tools.

https://github.com/user-attachments/assets/7f08b583-6adc-433f-92ff-c4ddc311541d


src/main.ts
```diff
@@ -324,6 +325,9 @@ export class Purchases {
 
   /** @internal */
   private async fetchAndCacheBrandingInfo(): Promise<void> {
+    if (isRCTestStoreApiKey(this._API_KEY)) {
+      throw Error("Branding info is not available for RC Test Store API keys.");
```
vicfergar, line 329: This will be expected when using a Test Store API Key. Should we log a **warning** in this case?

tonidero: Yeah good point, I think a warning makes more sense 👍

src/tests/helpers/test-store-purchase-helper.test.ts
```diff
@@ -0,0 +1,253 @@
...
+    const props = mountCall[1].props;
+
+    props?.onCancel();
+
+    try {
+      await promise;
+      expect(false).toBe(true); // Should not reach here
+    } catch {
+      // Expected error
+    }
```
vicfergar, line 249: nit: Using the `expectPromiseToError` helper function will make this a bit cleaner:

```suggestion
    await expectPromiseToError(
      promise,
      new PurchasesError(ErrorCode.UserCancelledError)
    );
```

PR #563, opened 2025-08-06
Rename `TestStore` to `SimulatedStore`

tonidero:
## Motivation / Description
As title says, a rename to reflect the latest name for the new store

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -15,7 +15,10 @@ import {
 import { type ProductResponse } from "./networking/responses/products-response";
 import { RC_ENDPOINT } from "./helpers/constants";
 import { Backend } from "./networking/backend";
-import { isRCTestStoreApiKey, isSandboxApiKey } from "./helpers/api-key-helper";
+import {
+  isSimulatedStoreApiKey,
+  isSandboxApiKey,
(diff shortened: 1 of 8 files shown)


src/main.ts
```diff
@@ -358,7 +361,7 @@ export class Purchases {
       Logger.debugLog(
         "Initializing Purchases SDK with Web billing sandbox API Key",
       );
-    } else if (isRCTestStoreApiKey(apiKey)) {
+    } else if (isSimulatedStoreApiKey(apiKey)) {
       Logger.debugLog("Initializing Purchases SDK with RC Test store API Key.");
```
vicfergar, line 365:
```suggestion
      Logger.debugLog("Initializing Purchases SDK with RC Simulated store API Key.");
```

vicfergar (approved):
Left a suggestion, but LGTM!

src/main.ts
tonidero: Actually, we're planning on leaving anything user facing, like logs with the `Test store` name. `Simulated store` will be the internal name for this store. This is to avoid ambiguity within our codebase, which is tricky with the "test" naming.

PR #566, opened 2025-08-07
chore: Paddle Support Limitation and API Key Restriction

vicfergar:
## Motivation / Description

While this SDK technically supports Paddle API keys, Paddle integration is still in development with limited feature availability. To prevent potential issues and confusion for developers, we're implementing stricter controls around Paddle API key usage.

## Changes introduced

- **Updated Error Message:** The invalid API key error message now only recommends using "Web Billing API key" instead of "Web Billing or Paddle API key," removing the suggestion to use Paddle.
- **Code Refactoring:** Improved API key validation by extracting detection logic into dedicated helper functions (`isPaddleApiKey`, `isWebBillingApiKey`, etc.).
- Updated API key validation to reject Paddle API keys by default

### DIFF
--- a/src/helpers/configuration-validators.ts
+++ b/src/helpers/configuration-validators.ts
@@ -1,3 +1,4 @@
+import type { FlagsConfig } from "../main";
 import { ErrorCode, PurchasesError } from "../entities/errors";
 import { SDK_HEADERS } from "../networking/http-client";
 import {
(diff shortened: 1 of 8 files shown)


src/helpers/configuration-validators.ts
```diff
@@ -1,18 +1,20 @@
...
-    !paddle_api_key_regex.test(apiKey) &&
-    !rc_test_store_api_key_regex.test(apiKey)
+    !isWebBillingApiKey(apiKey) &&
+    !isPaddleApiKey(apiKey) &&
+    !isRCTestStoreApiKey(apiKey)
   ) {
     throw new PurchasesError(
       ErrorCode.InvalidCredentialsError,
-      "Invalid API key. Use your Web Billing or Paddle API key.",
+      "Invalid API key. Use your Web Billing API key.",
```
vicfergar, line 17: @tonidero should we indicate that **SimulatedStore** API keys are supported, or is it still WIP?

tonidero: Still WIP, so better not to add it here for now. Thanks for checking!

tonidero (approved):
Looks great!

src/main.ts
```diff
@@ -884,7 +884,8 @@ export class Purchases {
    */
   public isSandbox(): boolean {
     return (
-      isSandboxApiKey(this._API_KEY) || isSimulatedStoreApiKey(this._API_KEY)
+      isWebBillingSandboxApiKey(this._API_KEY) ||
+      isSimulatedStoreApiKey(this._API_KEY)
```
tonidero, line 888: I've been thinking, it might be good to move this logic to the `api-key-helper` and use it in the `X-Is-Sandbox` header and possibly the `Backend` class calculation, which seems to only be used in tests...

Not directly related to this PR though, so can be done in a separate PR, or I can do it later.

tonidero (approved):
Just a possible improvement, but LGTM!

PR #569, opened 2025-08-08
Add getVirtualCurrencies backend network request

fire-at-will:
### Context
We're working on adding the ability to fetch virtual currency balances to the `purchases-js` SDK. To keep the size of the PRs down, we're breaking this functionality down into multiple PRs, and we'll expose the functionality the developers through the SDK's API when it's all ready.

### Description
This PR adds the ability for the SDK to call the backend's `v1/subscribers/{APP_USER_ID}/virtual_currencies` endpoint to fetch the virtual currencies for a given app user ID.

Specifically, this PR:
- Creates a new `GetVirtualCurrenciesEndpoint` object to describe the endpoint
- Adds a new function to the backend class to support calling this endpoint:
```typescript
async getVirtualCurrencies(
  appUserId: string,
): Promise<VirtualCurrenciesResponse>
```
- Adds unit tests for both the `GetVirtualCurrenciesEndpoint` object and the `backend.getVirtualCurrencies` function

No new public facing APIs or functionality is added by this PR.

### Next Steps
Next up, we'll:
- Create the public `VirtualCurrency` / `VirtualCurrencies` classes and build mappers to them from the `VirtualCurrenciesResponse` object. These objects will mirror those found in the mobile SDKs.
- Investigate caching the responses from the `backend.getVirtualCurrencies` calls like we do in the mobile SDKs

### DIFF
--- a/src/tests/test-responses.ts
+++ b/src/tests/test-responses.ts
@@ -398,6 +398,32 @@ export const checkoutCompleteResponse: CheckoutCompleteResponse = {
   },
 };
 
+export const getVirtualCurrenciesResponseWith3Currencies = {
+  virtual_currencies: {
+    GLD: {
+      balance: 100,
+      code: "GLD",
+      description: "It's gold",
+      name: "Gold",
+    },
+    SLV: {
+      balance: 100,
+      code: "SLV",
+      name: "Silver",
+    },
+    BRNZ: {
+      balance: -1,
+      code: "BRNZ",
+      description: "It's bronze",
+      name: "Bronze",
+    },
+  },
+};
+
+export const getVirtualCurrenciesResponseWithNoCurrencies = {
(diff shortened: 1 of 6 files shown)


src/tests/test-responses.ts
```diff
@@ -398,6 +398,32 @@ export const checkoutCompleteResponse: CheckoutCompleteResponse = {
...
+    GLD: {
+      balance: 100,
+      code: "GLD",
+      description: "It's gold",
+      name: "Gold",
+    },
+    SLV: {
+      balance: 100,
+      code: "SLV",
+      name: "Silver",
```
fire-at-will, line 412: Intentionally omitting `description` here since it is nullable and may not be included in responses from the backend

tonidero (approved):
Looks great!!

PR #570, opened 2025-08-11
[WEB-2778] Add tax breakdown support to checkout tax responses and examples

antoniobg:
## Motivation / Description

We added a new `tax_breakdown` property in our tax calculation response to better represent the tax details for each item. This change helps us display tax breakdowns more clearly in our application.

## Changes introduced

- Updated the `CheckoutCalculateTaxResponse` interface to include a `tax_breakdown` property.
- Adjusted the existing tax response fixtures to match the new structure.
- Modified the way tax data is accessed in the payment entry page.
- Made sure that tests and other parts of our codebase utilize the new `tax_breakdown`.

## Linear ticket (if any)

N/A


src/networking/responses/checkout-calculate-tax-response.ts
```diff
@@ -22,11 +22,7 @@ export interface CheckoutCalculateTaxResponse {
   tax_amount_in_micros: number;
   total_excluding_tax_in_micros: number;
   tax_inclusive: boolean;
-  pricing_phases: {
-    base: {
-      tax_breakdown: TaxBreakdown[];
-    };
-  };
```
antoniobg, line 29: @tonidero any concerns about this? This property is being replaced by `tax_breakdown`. The old `pricing_phases` is still returned by the API, but it's been marked as deprecated. Since this is a private interface, I think we can directly remove it.

tonidero (commented):
This LGTM, but since I don't have much context, will leave to someone else in the team to approve

PR #571, opened 2025-08-11
Expose Public Virtual Currency Models

fire-at-will:
## Motivation / Description
We're working on adding the ability to fetch virtual currency balances to the purchases-js SDK. To keep the size of the PRs down, we're breaking this functionality down into multiple PRs.

## Changes Introduced
This PR introduces the `VirtualCurrency` and `VirtualCurrencies` interfaces, but does not expose them publicly yet. These are the interfaces of the objects that the `getVirtualCurrencies()` and `getCachedVirtualCurrencies()` functions will return when they are implemented in a future PR. Their properties match those found in the mobile SDKs. 

We will expose them publicly in a future PR, when we add the public functions to the SDK.

```typescript
export interface VirtualCurrency {
  readonly balance: number;
  readonly name: string;
  readonly code: string;
}

export interface VirtualCurrencies {
  readonly all: { [key: string]: VirtualCurrency };
}
```

The PR also introduces functions to convert the backend response models to the public models, along with unit tests to ensure that the conversions are performed correctly

### Next Steps
Investigate caching the responses returned from the backend


src/main.ts
```diff
@@ -123,6 +123,8 @@ export type { PurchaseResult } from "./entities/purchase-result";
 export type { BrandingAppearance } from "./entities/branding";
 export type { PlatformInfo } from "./entities/platform-info";
 export type { PurchasesConfig } from "./entities/purchases-config";
+export type { VirtualCurrency } from "./entities/virtual-currency";
```
fire-at-will, line 126: This exposes the `VirtualCurrency` and `VirtualCurrencies` types publicly 👍

tonidero (approved):
Looks great!!

fire-at-will:
> we might also want to keep this PR on hold until the rest of the implementation is ready, so we can merge everything in a short time

That works for me - I'll merge this into a new `virtual-currency-dev` branch for now, and base future PRs off of that branch. 👍

PR #574, opened 2025-08-13
Implement getVirtualCurrencies

fire-at-will:
## Motivation / Description
We're working on adding the ability to fetch virtual currency balances to the purchases-js SDK. To keep the size of the PRs down, we're breaking this functionality down into multiple PRs.

## Changes Introduced
This PR introduces the `VirtualCurrency` and `VirtualCurrencies` interfaces, and exposes them through the `getVirtualCurrencies()` function.

```typescript
public async getVirtualCurrencies(): Promise<VirtualCurrencies> {
    return await this._getVirtualCurrenciesForUserId(this._appUserId);
}

export interface VirtualCurrency {
  readonly balance: number;
  readonly name: string;
  readonly code: string;
}

export interface VirtualCurrencies {
  readonly all: { [key: string]: VirtualCurrency };
}
```

### New getVirtualCurrencies() Function
The PR introduces the `getVirtualCurrencies()` function to the SDK, which fetches the user's virtual currencies from the network and caches them in memory for up to five minutes.

### Testing
- Added unit tests for the `getVirtualCurrencies()` function
- Added unit tests for the new `InMemoryCache` class

## Next Steps
This PR will be merged into the `virtual-currency-dev` branch. Next, we'll add the `getCachedVirtualCurrencies()` and `invalidateVirtualCurrenciesCache()` functions in a followup PR into `virtual-currency-dev`.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -968,7 +968,7 @@ export class Purchases {
     appUserId: string,
   ): Promise<VirtualCurrencies> {
     const cachedVirtualCurrencies =
-      await this.inMemoryCache.getCachedVirtualCurrencies(appUserId);
+      this.inMemoryCache.getCachedVirtualCurrencies(appUserId);
 
     if (cachedVirtualCurrencies) {
       Logger.debugLog("Vending VirtualCurrencies from cache.");


src/entities/virtual-currency.ts
```diff
@@ -35,6 +35,6 @@ export function toVirtualCurrency(
     balance: virtualCurrencyResponse.balance,
     name: virtualCurrencyResponse.name,
     code: virtualCurrencyResponse.code,
-    serverDescription: virtualCurrencyResponse.description,
+    serverDescription: virtualCurrencyResponse.description ?? null,
```
fire-at-will, line 38: Fixes a bug where if the `description` field was omitted in the response (our backend does this), `serverDescription` would be set to `undefined` instead of `null`

src/helpers/in-memory-cache.ts
```diff
@@ -0,0 +1,48 @@
+import type { VirtualCurrencies } from "../entities/virtual-currencies";
+
+interface CacheEntry<T> {
+  data: T;
+  timestamp: number;
+}
+
+export class InMemoryCache {
```
tonidero, line 8: I think this is fine since we don't use localStorage but pinging @nicfix to confirm. We've been trying to avoid caching most things in web until now.

src/helpers/in-memory-cache.ts
```diff
@@ -0,0 +1,48 @@
+import type { VirtualCurrencies } from "../entities/virtual-currencies";
+
+interface CacheEntry<T> {
+  data: T;
+  timestamp: number;
```
tonidero, line 5: Maybe rename to `lastUpdatedAt`?

src/helpers/in-memory-cache.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+export class InMemoryCache {
+  private virtualCurrenciesCache = new Map<
```
tonidero, line 9: Hmm would it make sense to make this a truly generic class instead of having maps for each piece of data we might want to cache? That way we would have different caches for `VirtualCurrencies` and other things.

src/helpers/in-memory-cache.ts
```diff
@@ -0,0 +1,48 @@
...
+      timestamp: Date.now(),
+    });
+  }
+
+  getCachedVirtualCurrencies(appUserID: string): VirtualCurrencies | null {
+    const entry = this.virtualCurrenciesCache.get(appUserID);
+    return this.getCachedData(entry ?? null);
+  }
+
+  invalidateVirtualCurrenciesCache(appUserID: string): void {
```
tonidero, line 30: I guess it shouldn't be an issue by having the map keyed by user ID... But I wonder if we should clear the cache when changing users... More of a cleanup note. We also clear caches when changing users on mobile and seems like a good practice...

tonidero (commented):
Great job on this! Only doubt is whether we want to start caching VirtualCurrencies, since we don't cache anything else as of now... I understand that this will make it easier to support the hybrid SDK APIs, where we have these caching APIs exposed... 

If we wanted to keep avoiding caching, in the hybrids, we could potentially:
- Have the invalidate cache API no-op on web
- Move the caching to the purchases-js-hybrid-mappings layer for the getCachedVirtualCurrencies.

Having said that, I do think it might be ok to cache some of these in memory, however, I think we might want to deal with the cleanup of the cache... Don't want to be leaking data across user changes I think.

cc @nicfix would be great to get your input here as well!

src/helpers/in-memory-cache.ts
fire-at-will: I considered this! There were two reasons I went with having one cache that could be expanded to include caching other items in the future:
- That's the pattern used by our DeviceCache in `purchases-ios`/`purchases-android`
- If we had one `InMemoryCache` instance per item that we're caching, that could lead to us storing lots of `InMemoryCache` objects in the `Purchases` object. That's not necessarily bad, but could lead to bloating the Purchases class a bit with a lot of cache objects, instead of just having one cache object.

tonidero:
> do you know if there's a reason we avoid caching most items in purchases-js today, other than maybe us just not getting around to implementing it yet?

I believe we basically wanted to avoid specially using localStorage since I believe it might have some issues and be less secure... As for in-memory cache, I believe we wanted to limit the complexity of having a cache (with the corresponding issues) while it wasn't needed, since it's pretty common on web to just perform a request for all resources on a refresh for example. I do think we should add some kind of caching at some point, just not sure if we've reached the point we have to 😅 . cc @web in case you have any other thoughts.

src/helpers/in-memory-cache.ts
fire-at-will: Good idea! If we decide to go forward with the caching, I'll update the PR to clear the cache out when `changeUser()` is called 👍

PR #579, opened 2025-08-26
Invalidate cache after simulated purchase

fire-at-will:
## Changes introduced
Previously, we were invalidating the in-memory cache after a regular purchase completed, but not after a simulated purchase completed. This PR changes that so that we now invalidate the in-memory cache after a simulated purchase has completed.


tonidero (approved):
Ohh good catch! this makes a lot of sense, thanks!

PR #584, opened 2025-09-01
Add `currency` property to simulated store post receipt

tonidero:
## Motivation / Description
This adds a `currency` property in the POST /receipt performed by simulated store purchases. This could potentially be useful in the future to correctly attribute the purchased currency.


ajpallares (approved):
Makes sense!

PR #585, opened 2025-09-04
Fix Expo Go compatibility

tonidero:
## Motivation / Description
We were running into issues in Expo Go in react-native-purchases because it wasn't able to find the `crypto` library, and when importing the `uuid` library it was saying that it couldn't find the crypto library.

This refactors the code to handle missing `crypto` definition + use our custom code to generate UUIDs, instead of using the `UUID` library directly, which fails because it internally uses `crypto` when importing it.

Note that this means that in Expo Go, we will only have pseudo-random UUID generation.

### DIFF
--- a/src/behavioural-events/event.ts
+++ b/src/behavioural-events/event.ts
@@ -1,5 +1,5 @@
 import { camelToUnderscore } from "../helpers/camel-to-underscore";
-import { v4 as uuidv4 } from "uuid";
+import { generateUUID } from "../helpers/uuid-helper";
 
 export type EventData = {
   eventName: string;
@@ -49,7 +49,7 @@ export class Event {
   public readonly data: EventData;
 
   public constructor(data: EventData) {
-    this.id = uuidv4();
+    this.id = generateUUID();
     this.timestampMs = Date.now();
     this.data = data;
   }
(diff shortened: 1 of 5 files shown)


src/behavioural-events/event.ts
```diff
@@ -1,5 +1,5 @@
 import { camelToUnderscore } from "../helpers/camel-to-underscore";
-import { v4 as uuidv4 } from "uuid";
+import { generateUUID } from "../helpers/uuid-helper";
```
tonidero, line 2: Just importing `uuid` was causing issues in Expo Go... so I changed to use our `generateUUID` instead.

src/helpers/uuid-helper.ts
```diff
@@ -1,20 +1,39 @@
...
+    return cryptoObj.randomUUID();
   }
 
   const bytes = new Uint8Array(16);
-  if (crypto && crypto.getRandomValues) {
-    crypto.getRandomValues(bytes);
-  } else {
-    for (let i = 0; i < 16; i++) {
-      bytes[i] = Math.floor(Math.random() * 256);
+  try {
```
tonidero, line 26: This might not be needed, but I wanted to be extra safe...

antoniobg (approved):
Amazing. Thank you @tonidero I've been running into this issue since yesterday

PR #587, opened 2025-09-05
Fail posting simulated receipt if not using test api key

tonidero:
## Motivation / Description
This PR has 2 changes:
- It immediately fails if trying to post a simulated store receipt with a non-test API key.
- It exposes an internal method to detect whether the SDK was configured with a test API key


ajpallares (approved):
🚀

Issue #578, opened 2025-08-21
Unnecessary Loading of Stripe.js


taszilvo:
Hey,

I still face this issue... [https://github.com/RevenueCat/purchases-js/issues/311](https://github.com/RevenueCat/purchases-js/issues/311)

We use your SDK with Paddle integration. We only initialize and use RevenueCat's purchases SDK for fetching the offerings and products, for checkout flow we only use Paddle checkout directly. That means we never use Stripe, however I still see Stripe's iframe loaded and a lot of Stripe background network calls.

Thank you for your response!

pedroimpulcetto:
hum, ok 
PR #542 basically reverts the Stripe stuff
Stripe SDK is still preloaded in the main bundle

divebuddycom:
I am also using @revenuecat/purchases-js (v1.20.0) and noticing several Network POST API calls to Stripe (https://m.stripe.com/6). These calls occur on every screen navigation or page visit. I am not using the Stripe SDK (@stripe/stripe-react-native) and have no other Stripe dependencies in my project. Why is this happening?

pedroimpulcetto:
Hey @divebuddycom I have PR #597 fixing this issue, can you please help me testing if that works for your use case?

divebuddycom:
PR #597 did not fix the problem. I am still seeing several calls to stripe. All of the "6" calls you see in the screenshot occurred within the same couple of seconds, after only navigating to two random pages on my site.

<img width="1147" height="393" alt="Image" src="https://github.com/user-attachments/assets/f73929bd-ddd4-4ae9-92d9-93d33b18f11f" />

divebuddycom:
Any update to this issue?

psnet:
Will this be fixed?

nicfix:
Hi all, we recently switched to `stripe-pure` which loads the stripe js only when it's really needed.
This will prevent any call to Stripe to happen before the purchase method is called.

Issue #311, opened 2025-02-11
Unnecessary Loading of Stripe.js


ArianHamdi:
In our case, we only use the `getCustomerInfo` method. Do you still think it is necessary to load Stripe.js?

<img width="1093" alt="Image" src="https://github.com/user-attachments/assets/de637003-b7fb-4f7e-b07f-6cccd933c7fe" />

therealstein:
Yes please stop it. Its annoying !

alfondotnet:
Hi @ArianHamdi .

Makes sense to me that we don't prematurely download the bundle if you don't call `.purchase`. Will get back to you on this quick, sorry for the long time to respond.

PR #597, opened 2025-09-15
fix: Importing loadStripe without side effects

pedroimpulcetto:
## Motivation / Description
Do not prematurely load the Stripe module before it is called 
Fixing #578 and  #311 

## Changes introduced
Using [`stripe-js/pure`](https://github.com/stripe/stripe-js?tab=readme-ov-file#importing-loadstripe-without-side-effects) to load stripe module without side effects 

> If you would like to use loadStripe in your application, but defer loading the Stripe.js script until loadStripe is first called, use the alternative @stripe/stripe-js/pure import module:


nicfix:
Closing since we recently migrated to stripe-pure and this issue is no more there, thanks for your PR!

PR #606, opened 2025-10-03
Connecting the new data needed for the paywalls component

nicfix:
## Motivation / Description
We are working on adding Web Paywalls.
This PR aligns the SDK with the latest changes to the purchases-ui-js to do so.

### DIFF
--- a/src/entities/offerings.ts
+++ b/src/entities/offerings.ts
@@ -412,14 +412,14 @@ export interface Offering {
    * The paywall components configured in the RevenueCat dashboard, if available.
    * @internal
    */
-  readonly paywall_components: PaywallData | null;
+  readonly paywallComponents: PaywallData | null;
 
   /**
    * The UI configuration needed to render the paywall components, usually set
    * together
    * @internal
    */
-  readonly ui_config?: UIConfig;
+  readonly uiConfig?: UIConfig;
 }
 
 /**
@@ -782,8 +782,8 @@ export const toOffering = (
     twoMonth: packagesById[PackageType.TwoMonth] ?? null,
     monthly: packagesById[PackageType.Monthly] ?? null,
     weekly: packagesById[PackageType.Weekly] ?? null,
-    paywall_components: offeringsData.paywall_components,
-    ui_config: uiConfig,
+    paywallComponents: offeringsData.paywall_components,
+    uiConfig: uiConfig,
   };
 };
 
--- a/src/main.ts
+++ b/src/main.ts
@@ -444,11 +444,11 @@ export class Purchases {
     certainHTMLTarget.innerHTML = "";
 
     const offering = paywallParams.offering;
-    if (!offering.paywall_components) {
+    if (!offering.paywallComponents) {
       throw new Error("You cannot use paywalls yet, they are coming soon!");
     }
 
-    if (!offering.ui_config) {
+    if (!offering.uiConfig) {
       throw new Error(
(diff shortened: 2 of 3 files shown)


tonidero (commented):
Some comments about the public API

api-report/purchases-js.api.md
```diff
@@ -538,6 +567,11 @@ export interface VirtualCurrency {
     readonly serverDescription: string | null;
 }
 
+// Warnings were encountered during analysis:
+//
+// dist/Purchases.es.d.ts:14:5 - (ae-forgotten-export) The symbol "AppFontFamilyConfig" needs to be exported by the entry point Purchases.es.d.ts
+// dist/Purchases.es.d.ts:1501:20 - (ae-forgotten-export) The symbol "AppFontsConfig" needs to be exported by the entry point Purchases.es.d.ts
```
tonidero, line 573: Looks like we're not exporting these types?

package-lock.json
```diff
@@ -60,6 +60,53 @@
         "vitest": "^3.2.4"
       }
     },
+    "../purchases-ui-js": {
```
tonidero, line 63: Hmm this is weird... Why did it change to use the npm dependency to use a local dependency?

examples/webbilling-demo/src/tests/localization.test.ts
```diff
@@ -11,21 +11,21 @@ import { integrationTest } from "./helpers/integration-test";
...
-  ["de", "Sicherer Checkout über RevenueCat"],
+  ["es", "Suscribirse a", "Pago seguro mediante RevenueCat"],
+  ["it", "Abbonati a", "Pagamento sicuro tramite RevenueCat"],
+  ["en", "Subscribe to", "Secure checkout by RevenueCat"],
+  ["fr", "S'abonner à", "Paiement sécurisé par RevenueCat"],
+  ["de", "Abonnieren", "Sicherer Checkout über RevenueCat"],
 ];
 
-TEST_CASES.forEach(([lang, title]) => {
+TEST_CASES.forEach(([lang, subscribeTo, safePayment]) => {
```
tonidero, line 21: Hmm `safePayment` seems unused now right? Should we just remove it?

src/main.ts
```diff
@@ -448,6 +448,12 @@ export class Purchases {
       throw new Error("You cannot use paywalls yet, they are coming soon!");
```
tonidero, line 448: Would it be worth changing this error to something like
```suggestion
      throw new Error("No paywall found for this offering. Please make sure you've added a paywall in the RevenueCat dashboard.");
```

Not directly related to this PR though.

src/entities/offerings.ts
```diff
@@ -408,7 +408,18 @@ export interface Offering {
...
+   * @internal
+   */
   readonly paywall_components: PaywallData | null;
+
+  /**
+   * The UI configuration needed to render the paywall components, usually set
+   * together
+   * @internal
+   */
+  readonly ui_config?: UIConfig;
```
tonidero, line 422: Would it be worth changing these to camelCase now that they are internal? Can also happen in future PRs though.

examples/webbilling-demo/src/tests/localization.test.ts
tonidero: Ahh true, this is ok then 👍

tonidero (approved):

PR #631, opened 2025-10-24
Simplified the creation of fixtures for paywall variables tests.

nicfix:
Refactored the test and the variables fixtures to make it easier to add the test case I need.


package.json
```diff
@@ -45,7 +45,8 @@
     "lint": "prettier --check .",
     "format": "prettier --write .",
     "pack-build": "npm run build && npm pack",
-    "test": "vitest",
+    "test": "vitest run",
```
nicfix: Split the 2 so that Junie won't get stuck waiting for the watch

PR #632, opened 2025-10-29
Add period and price variables

marek-dabek:
## Motivation / Description

https://linear.app/revenuecat/issue/WEB-3373/several-variables-are-never-populated-in-the-purchases-js-code

## Changes introduced

* Moved period and price to separate files
* More conversion constants
* Add `price_per_*` variables
* Add `period_in_*` variables

### DIFF
--- a/src/helpers/paywall-price-helpers.ts
+++ b/src/helpers/paywall-price-helpers.ts
@@ -94,13 +94,7 @@ function getPricePerMonth(
   period: Period | null,
   translator: Translator,
 ) {
-  // Check if price is valid
-  if (!price || price.amountMicros === 0 || !price.currency) {
-    return "N/A";
-  }
-
-  // If no period info, assume it's a monthly price
-  if (!period || !period.number || period.number <= 0) {
+  if (!period) {
     return translator.formatPrice(price.amountMicros, price.currency);
   }
 


src/helpers/paywall-price-helpers.ts
```diff
@@ -0,0 +1,200 @@
...
+  return fallback;
+}
+
+function getPricePerMonth(
+  price: Price,
+  period: Period | null,
+  translator: Translator,
+) {
+  // Check if price is valid
+  if (!price || price.amountMicros === 0 || !price.currency) {
```
marek-dabek: Can those be zero or missing? If so, do we need to add this to other periods as well?

src/helpers/paywall-price-helpers.ts
```diff
@@ -0,0 +1,200 @@
...
+) {
+  const fallback = translator.formatPrice(price.amountMicros, price.currency);
+
+  if (!period) {
+    return fallback;
+  }
+
+  if (period.unit === "year") {
+    return translator.formatPrice(
+      price.amountMicros / MONTHS_PER_YEAR / WEEKS_PER_MONTH,
```
marek-dabek: Is this correct? No period.number division.

Also wouldn't it be better to use `WEEKS_PER_YEAR`?

src/helpers/paywall-period-helpers.ts
```diff
@@ -0,0 +1,76 @@
...
+      return period.number * DAYS_PER_WEEK;
+    case PeriodUnit.Day:
+      return period.number;
+  }
+}
+
+function getDurationInWeeks(period: Period): number {
+  switch (period.unit) {
+    case PeriodUnit.Year:
+      return period.number * WEEKS_PER_YEAR;
```
marek-dabek: Price calculation uses `MONTHS_PER_YEAR / WEEKS_PER_MONTH`. Which is **more** correct?

src/helpers/paywall-price-helpers.ts
nicfix: Yeah, WEEKS_PER_YEAR is definitely a better approximation.

src/helpers/paywall-price-helpers.ts
marek-dabek: Background: months were added first and we supported `months` and `everything else`. Weeks were added later, now days and years.

Sounds like it's safe to assume we don't need this check anymore.

PR #637, opened 2025-11-03
WEB-3390: presentPaywall method.

nicfix:
## Motivation / Description
Renamed the internal method with the same name we use in the other platforms.
Made the method public.

## Changes introduced
* Renamed the internal method with the same name we use in the other platforms.
* Made the method public.
* Updated the demo app to use the new name
* Removed the ts-ignore needed to use an @internal method.
* Updated the public API docs.

### DIFF
--- a/api-report/purchases-js.api.md
+++ b/api-report/purchases-js.api.md
@@ -281,6 +281,25 @@ export interface PresentedOfferingContext {
 }
 
 // @public
+interface PresentPaywallParams {
+    // (undocumented)
+    readonly customerEmail?: string;
+    // (undocumented)
+    readonly htmlTarget?: HTMLElement;
+    readonly offering: Offering;
+    // (undocumented)
+    readonly onBack?: () => void;
+    // (undocumented)
+    readonly onNavigateToUrl?: (url: string) => void;
+    // (undocumented)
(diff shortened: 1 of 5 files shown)


francocorreasosa (approved):

src/main.ts
```diff
@@ -410,17 +410,12 @@ export class Purchases {
 
   /**
    * Renders an RC Paywall and allows the user to purchase from it using Web Billing.
-   * @experimental
-   * @internal
```
tonidero: Just to check that we're ready then for this to be used by customers on the next release? Just confirming :)

nicfix: Yes, this is already being used in Web Purchase Links!

api-report/purchases-js.api.md
```diff
@@ -280,6 +280,25 @@ export interface PresentedOfferingContext {
...
 }
 
+// @public
+interface PresentPaywallParams {
+    // (undocumented)
+    readonly customerEmail?: string;
+    // (undocumented)
+    readonly htmlTarget?: HTMLElement;
+    readonly offering: Offering;
+    // (undocumented)
```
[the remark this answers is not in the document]
nicfix: Agree, will do in a follow up.

PR #648, opened 2025-11-13
WEB-3408: Add Overlay background when renderPaywall is called with no targetElement.

nicfix:
## Changes introduced
Changed the sdk so that the paywall will appear as an overlay when presentPaywall is called with no target element


tonidero (approved):
Changes make sense to me!

PR #651, opened 2025-11-14
Add workflowIdentifier to PresentedOfferingContext interface

elenaperezrioja:
## Motivation / Description
Pass workflow_identifier when making a purchase. This will be added in the backend to the purchase session and will let us identify that a puchase came from navigating a workflow.

## Changes introduced
Using the workflow is present in the purchases context to send it in the checkout request

### DIFF
--- a/src/networking/backend.ts
+++ b/src/networking/backend.ts
@@ -25,6 +25,7 @@ import type {
   PurchaseMetadata,
   PurchaseOption,
 } from "../entities/offerings";
+import type { PurchasesContext } from "../entities/purchases-config";
 import type { CheckoutCompleteResponse } from "./responses/checkout-complete-response";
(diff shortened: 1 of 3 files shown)


src/networking/backend.ts
```diff
@@ -169,6 +177,11 @@ export class Backend {
         presentedOfferingContext.placementIdentifier;
     }
 
+    if (this.purchasesContext?.workflowIdentifier) {
+      requestBody.presented_workflow_id =
+        this.purchasesContext.workflowIdentifier;
```
tonidero: We might end up adding an intermediate `PresentedWorkflowContext` in the base PR, so this might change slightly, but can wait until that's done :)

tonidero (approved):
I think these changes make sense! Only thing is that if we add the `PresentedWorkflowContext` inside `PurchasesContext`, it might change the implementation slightly, but we can fix it if/when we do that.

PR #654, opened 2025-11-16
Fix `identifyUser` wasCreated

tonidero:
## Motivation / Description
We were incorrectly obtaining the was_created field from the response. We should have used the response code instead. This fixes that and should deal with https://github.com/RevenueCat/purchases-flutter/issues/1539


ajpallares (approved):
Makes sense! Thanks for the fix!

PR #665, opened 2025-11-25
WEB-3461: Add presentExpressPurchaseButton internal method

nicfix:
## Motivation / Description
We want to give developers the possibility to execute a purchase by using express checkout buttons without showing our checkout form.

This PR does that. This is the first of other additional PRs so for now the method is marked as `@internal` and `@experimental`.

## Changes introduced
* Adds the new `presentExpressPurchaseButton` as an `@internal` and `@experimental` method.
* Adds the new `ExpressPurchaseButton` component which implements the entire checkout flow specialized for express checkout elements
* Passes the pre-rendered `ExpressPurchaseButton` down the paywall's Purchase button, only if the `useExpressPurchaseButtons` option is passed to the `PresentPaywallParams`.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -14,8 +14,8 @@ importers:
         specifier: ^7.48.0
         version: 7.52.13(@types/node@22.10.5)
       "@revenuecat/purchases-ui-js":
-        specifier: 2.1.1
-        version: 2.1.1(svelte@5.17.0)
+        specifier: 2.2.1
+        version: 2.2.1(svelte@5.17.0)
       "@storybook/addon-essentials":
         specifier: ^8.5.0
         version: 8.6.4(@types/react@19.0.10)(storybook@8.6.4(prettier@3.4.2))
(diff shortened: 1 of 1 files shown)


src/ui/express-purchase-button.svelte
```diff
@@ -0,0 +1,299 @@
...
+    if (!newClientSecret) {
+      return false;
+    }
+
+    const newConfirmationToken = await updateTaxCalculationBasedOnWalletDetails(
+      elements,
+      stripe,
+    );
+
+    // We'll charge the customer with the ip based tax.
```
nicfix: Note to self:
This is not accurate, I'll remove this comment

src/ui/molecules/stripe-express-checkout-element.svelte
```diff
@@ -22,36 +22,51 @@
...
     ) => void | Promise<void>;
+    onClick?: (event: StripeExpressCheckoutElementClickEvent) => void;
+    onShippingAddressChange?: (
+      event: StripeExpressCheckoutElementShippingAddressChangeEvent,
+    ) => void;
     elements: StripeElements;
     billingAddressRequired: boolean;
     forceEnableWalletMethods: boolean;
     expressCheckoutOptions?: StripeExpressCheckoutConfiguration;
+    hideOtherOptions?: boolean;
```
vicfergar: It's not very clear that this will hide the other options separator, could we rename to:
```suggestion
    hideCheckoutSeparator?: boolean;
```

Another option is extracting the separator form this molecule

vicfergar (commented):
There is some code duplication between `src/ui/express-purchase-button.svelte` and `src/ui/molecules/stripe-elements.svelte`. Could we refactor to reuse as much code as possible?

I’m also concerned about the following:
- Losing behavioral event tracking
- ~Tax recalculation logic might not work properly in tax-exclusive scenarios~ We've verified it's working as expected

vicfergar (approved):
I think there’s still room to improve code reusability, but LGTM so far.

PR #682, opened 2025-12-03
Add stepId parameter to purchase flow tracking

elenaperezrioja:
❗ Don't merge before https://github.com/RevenueCat/khepri/pull/16308 is merged
## Motivation / Description
Add support for tracking workflow step IDs in the purchase flow to track where purchases are initiated within workflows. This is because a workflow might have multiple steps that trigger a purchase.

## Changes introduced
- Added `stepId` parameter to the `PurchaseParams` interface
- Updated the purchase operation helper to pass the step ID through the purchase flow
- Modified the backend API call to include the step ID in the checkout start request as `presented_step_id` (PR pending)
- Added tests to verify the step ID is properly included in or omitted from requests
- Updated the Purchases UI component to accept and pass the step ID parameter


src/entities/purchase-params.ts
```diff
@@ -25,6 +25,12 @@ export interface PurchaseParams {
    */
   customerEmail?: string;
 
+  /**
+   * The step ID from the workflow where the purchase is being initiated.
+   * @internal
+   */
+  stepId?: string;
```
tonidero: Just to confirm, this is not known at SDK configuration time, so we can't include this info in the `WorkflowContext` there right? 

I'm not in love with splitting the information given to the SDK between configuration time and purchase time like this, specially when it means modifying the API like this. I'm not totally familiar with the architecture on how workflows will work though... And it's internal so I guess it's ok for now...

Maybe we could consider passing something like an object `WorkflowPurchaseContext` however, to group this information together, in case we want to add more data in the future, to not have too many generic fields like this?

Another option could be to use the `metadata` field below, but I'm not sure if that could be problematic in the backend.

tonidero:
> I'm not sure how to fix the errors... I'm not getting anything in local

For the `danger` test, you can just rerun it in CircleCI (you will need to log in with your github account). For the test-e2e, I'm not sure... maybe rerunning, in case it's flaky tests?

PR #685, opened 2025-12-04
Web 3459: Connect the loader from purchases-ui-js

nicfix:
## Motivation / Description

  - Wire the wallet/express checkout “ready” callback through our paywall flow so the updater is provided only after the Stripe express button
    is actually ready, matching the updated purchases-ui behavior and avoiding premature updates.

  ## Changes introduced

  - walletButtonRender now takes an { selectedPackageId, onReady } bag and invokes onReady after storing the button updater (src/main.ts).
  - Thread an optional onReady prop through the express purchase button types, wrapper, and Svelte component, calling onButtonReady when the
    underlying ExpressCheckoutElement signals readiness (src/ui/express-purchase-button/*).
  - Point @revenuecat/purchases-ui-js to the local checkout and refresh the lockfile to reference file:../purchases-ui-js (package.json, pnpm-
    lock.yaml).

  ## Linear ticket (if any)

  - WEB-3459

  ## Additional comments

  - Dependency currently targets a local @revenuecat/purchases-ui-js; swap back to a published version before releasing.


package.json
```diff
@@ -60,7 +60,7 @@
     "@eslint/js": "^9.16.0",
     "@microsoft/api-extractor": "^7.48.0",
     "@paddle/paddle-js": "^1.5.1",
-    "@revenuecat/purchases-ui-js": "2.2.1",
+    "@revenuecat/purchases-ui-js": "file://../purchases-ui-js",
```
nicfix: Temporarily until I publish the 2.2.2 version of purchases-js

PR #690, opened 2025-12-13
[WEB-3429] Make close buttons conditional

james-od:
## Motivation / Description

https://linear.app/revenuecat/issue/WEB-3429/close-buttons-not-rendering-on-react-native-web

This is the second of three PRs. The aim is to re-eenable close buttons, but allow the passing of an override to hide them. We want to hide them only in Web Purchase Links (rc-billing-checkout).

In this PR we pass `hideBackButtons` from `paywallParams` where we mount `Paywall` in `presentPaywall`. Additionally we gate `onBackClicked` on `hideBackButtons`.
I've updated the demo project to accept a `hideBackButtons` query param in the `rc_paywall` route. 
And then added two integration tests checking for the close button being visible/hidden.
I'm just using a CSS locator, to avoid needing to add a test id for close buttons.

See also:
- https://github.com/RevenueCat/purchases-ui-js/pull/177 (Once this has been released, we'll need to bump the purchases-ui-js version and confirm the e2e tests then pass).
- https://github.com/RevenueCat/rc-billing-checkout/pull/310

### DIFF
--- a/api-report/purchases-js.api.json
+++ b/api-report/purchases-js.api.json
@@ -3918,6 +3918,33 @@
             },
             {
               "kind": "PropertySignature",
+              "canonicalReference": "@revenuecat/purchases-js!PresentPaywallParams#hideBackButtons:member",
+              "docComment": "/**\n * Whether to hide back buttons in the paywall. Defaults to false.\n */\n",
+              "excerptTokens": [
+                {
+                  "kind": "Content",
+                  "text": "readonly hideBackButtons?: "
+                },
+                {
+                  "kind": "Content",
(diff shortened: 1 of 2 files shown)


src/main.ts
```diff
@@ -649,6 +649,9 @@ export class Purchases {
           onVisitCustomerCenterClicked: onVisitCustomerCenterClicked,
           uiConfig: offering.uiConfig!,
           onBackClicked: () => {
+            if (paywallParams.hideBackButtons) {
```
nicfix: Maybe we don't need to prevent the callback from being called.

PR #722, opened 2026-01-16
Web 3521: Using GET /checkout/credentials instead of POST /checkout/start to render the Express Purchase Button

nicfix:
## Motivation / Description
When I first implemented the Express Checkout Button I used the `/checkout/start` endpoint to retrieve the stripe credentials to render the express checkout elements. However the `/checkout/start` endpoint is throttled since it actually starts a purchase creating an operation session etc.

This caused issues when multiple ECB were rendered at the same time, and or they were rendered quickly one after the other (as example in a useEffect in dev due to React.StrictMode). The endpoint was returning 429 causing a failure in rendering the button.

The new /credentials endpoint doesn't need throttling since it just fetches the credentials for the related payment gateway.

This PR uses the new endpoint and starts a checkout only when the button is actually clicked.

## Changes introduced
* Using the /credentials endpoint to get the stripe credentials before rendering the express purchase button
* Added the express_purchase_buttons page to the demo app to test it with multiple express purchase buttons

Sadly we cannot run (yet) the E2E tests on this page since we need https to be able to render these buttons, I have a dedicated ticket to try and replicate the fake-https environment I create locally in the E2E tests container.

### DIFF
--- a/src/tests/helpers/purchase-operation-helper.test.ts
+++ b/src/tests/helpers/purchase-operation-helper.test.ts
@@ -79,7 +79,7 @@ describe("PurchaseOperationHelper", () => {
 
   function setGetClientCredentialsResponse(httpResponse: HttpResponse) {
     server.use(
-      http.post(
+      http.get(
         "http://localhost:8000/rcbilling/v1/checkout/credentials",
         () => {
           return httpResponse;
--- a/src/tests/networking/backend.test.ts
+++ b/src/tests/networking/backend.test.ts
@@ -512,7 +512,7 @@ describe("getProducts request", () => {
 describe("getClientCredentials request", () => {
   function setClientCredentialsResponse(httpResponse: HttpResponse) {
     server.use(
-      http.post(
+      http.get(
         "http://localhost:8000/rcbilling/v1/checkout/credentials",
         () => {
           return httpResponse;
--- a/src/tests/networking/endpoints.test.ts
+++ b/src/tests/networking/endpoints.test.ts
@@ -121,7 +121,7 @@ describe("checkoutClientCredentials endpoint", () => {
   const endpoint = new CheckoutClientCredentials();
 
   test("uses correct method", () => {
-    expect(endpoint.method).toBe("POST");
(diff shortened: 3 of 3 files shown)


src/ui/molecules/stripe-express-checkout-element.svelte
```diff
@@ -72,10 +74,19 @@
   const onClickCallback = async (
     event: StripeExpressCheckoutElementClickEvent,
   ) => {
-    const options = {
+    let options = {
       ...(expressCheckoutOptions ? expressCheckoutOptions : {}),
     } as ClickResolveDetails;
-    onClick && onClick(event);
+    if (onClick) {
```
nicfix: Note to self: Not sure I like this change, I can use the event to resolve the options as it was happening before.

nicfix:
Ok I discovered that there are still some bits missing, I'll work on them and then reopen the PR.

src/ui/express-purchase-button/express-purchase-button.svelte
```diff
@@ -179,35 +187,26 @@
...
+          await initStripe(clientCredentialsResult, rcPackage);
         stripe = stripeInstance;
         elements = elementsInstance;
         expressCheckoutOptions = expOptions;
-      } else {
-        await updateStripe(checkoutStartResult, elements);
       }
     } catch (e) {
       handleError(e as PurchaseFlowError);
+      return;
```
nicfix: Note to self: This would leave the button in "loading"

src/ui/express-purchase-button/express-purchase-button.svelte
```diff
@@ -179,35 +187,26 @@
...
     } catch (e) {
+      handleError(e as PurchaseFlowError);
       return;
     }
 
     try {
       if (!stripe || !elements) {
         const { stripeInstance, elementsInstance, expOptions } =
-          await initStripe(checkoutStartResult, rcPackage);
+          await initStripe(clientCredentialsResult, rcPackage);
```
nicfix: Note to self: Add the purchaseOption here

PR #723, opened 2026-01-16
chore: Draft Discount Fields in SDK Request/Response Schemas

vicfergar:


src/networking/responses/products-response.ts
```diff
@@ -12,6 +14,9 @@ export interface PricingPhaseResponse {
 export interface PurchaseOptionResponse {
   id: string;
   price_id: string;
+  discount_identifier: string | null;
```
vicfergar: I guess we'll eventually need to include this `discount_identifier` in the `checkout/start` request body, so the backend knows which discount was selected/applied at checkout

src/networking/responses/products-response.ts
```diff
@@ -12,6 +14,9 @@ export interface PricingPhaseResponse {
...
+  discount_name: string | null;
+  discount_percentage: number | null;
```
vicfergar: This was added for representation purposes, though it might not be needed at this stage.

src/networking/responses/products-response.ts
```diff
@@ -1,6 +1,8 @@
 export interface PriceResponse {
   amount_micros: number;
   currency: string;
+  discount_amount_micros: number | null;
```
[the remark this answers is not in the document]
james-od: Dumb question, how specifically would `discount_amount_micros` and `discounted_amount_micros` differ? Former is [amount being discounted], latter is [amount_micros] - [amount being discounted]?

PR #804, opened 2026-03-17
feat: Add PaywallListener for purchase lifecycle callbacks

tonidero:
## Summary
- Adds a `PaywallListener` interface with `onPurchaseStarted`, `onPurchaseError`, and `onPurchaseCancelled` optional callbacks for observing purchase lifecycle events within `presentPaywall`
- Wires listener into both the regular purchase flow and the express purchase button flow, with try/catch safety around all listener invocations
- Deprecates `PresentPaywallParams.onPurchaseError` in favor of `listener.onPurchaseError` (both still called when both are provided)
- Adds toast notifications to the example app paywall launcher page to demonstrate listener callbacks

## Test plan
- [x] New unit tests in `paywall-listener.test.ts` (6 tests) pass
- [x] Full test suite passes (520 tests)
- [x] API report updated via `api-extractor`
- [x] Manual verification: launch paywall in demo app, confirm toasts appear for purchase started, error, and cancellation events


src/main.ts
```diff
@@ -670,6 +709,7 @@ export class Purchases {
               return {};
             }
             let buttonUpdater: ExpressPurchaseButtonUpdater | null = null;
+            notifyPurchaseStarted(pkg);
```
nicfix: This is not really the right spot for the ExpressPurchaseButton.
Basically it will be triggered by any rendering of the button in this place.

We should probably expose an explicit parameter to the presentExpressPurchaseButton and pass it down to the component that handles it.

src/main.ts
```diff
@@ -687,9 +727,7 @@ export class Purchases {
                 Logger.errorLog(
                   `Error presenting express purchase button: ${err}`,
                 );
-                if (paywallParams.onPurchaseError) {
-                  paywallParams.onPurchaseError(err);
-                }
+                notifyPurchaseError(err);
```
nicfix: We could pass this to the presentExpressPurchaseButton method as parameter so that it's handled by it instead of here.

nicfix (commented):
Overall the approach makes sense, I just think the callbacks for the express purchase button should be passed to the presentExpressPurchaseButton method to make sure they are then passed to the right callbacks inside the component that renders it.

As they are now:
1. The start purchase will be called every time a Express purchase button is rendered, not clicked
2. The purchase cancelled works, but if we move it inside the presentExpressPurchaseButton we make it available to that method as well

src/main.ts
tonidero: Ohhh my bad... I should have tested that flow 🥴. I will address this next week. Thanks for the feedback!!

nicfix: No worries, Express Purchase Button it's a bit tricky!

nicfix (approved):

PR #818, opened 2026-03-24
Add presented offering context to paywall events

rickvdl:
## Summary
- Include `placement_identifier`, `targeting_revision`, and `targeting_rule_id` in paywall events under a `presented_offering_context` nested object
- Field is omitted when no placement or targeting data exists
- Consistent with iOS and Android implementations

## Related PRs
- iOS: RevenueCat/purchases-ios#6476
- Android: RevenueCat/purchases-android#3253

## Test plan
- [x] Unit tests for all context permutations (full, placement-only, targeting-only, none, empty string)
- [x] JSON round-trip serialization tests
- [x] Integration test verifying context flows from offering to event payload
- [x] Manual testing with targeting rules and placements in demo app

### DIFF
--- a/src/behavioural-events/paywall-event.ts
+++ b/src/behavioural-events/paywall-event.ts
@@ -1,3 +1,4 @@
+import type { PresentedOfferingContext } from "../entities/offerings";
 import { generateUUID } from "../helpers/uuid-helper";
 
 export type PaywallEventType =
@@ -44,6 +45,7 @@ interface BasePaywallEventData {
   offeringId: string;
   paywallRevision: number;
   paywallRcPublicId: string | null;
+  presentedOfferingContext?: PresentedOfferingContext;
 }
 
 interface PaywallDisplayData {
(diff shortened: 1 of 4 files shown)


src/behavioural-events/paywall-event.ts
```diff
@@ -191,6 +200,50 @@ const toComponentInteractionPayload = (
   return payload as PaywallComponentInteractionPayload;
 };
 
+function toPresentedOfferingContextPayload(
+  context: PresentedOfferingContext | undefined,
+): PresentedOfferingContextPayload | undefined {
```
tonidero: NABD, but I've usually preferred returning null to undefined...

tonidero (approved):
Just a small comment... but looking good in any case. Thanks for adding this!

src/behavioural-events/paywall-event.ts
rickvdl: Ah right, fixed :)

PR #825, opened 2026-04-01
Refactor checkoutStart/postCheckoutStart to use params objects

vkartaviy:
## Summary

- Replace positional parameters in `checkoutStart` and `postCheckoutStart` with single typed params objects
- Rename `email` → `customerEmail`, `stepId` → `presentedStepId` for clarity
- Group params by domain: purchase identity, presentation context, customer data

Motivated by [WEB-4004](https://linear.app/revenuecat/issue/WEB-4004) (locale passthrough) — adding another optional positional param made call sites unreadable. This cleanup lands first so the locale PR is a clean one-field addition.

Discussed and aligned with the team in [Slack](https://revenuecat.slack.com/archives/C03UKCQLW1Z/p1775051572428789).

## What changed

**`purchase-operation-helper.ts`** — `checkoutStart(params: CheckoutStartParams)`
**`backend.ts`** — `postCheckoutStart(params: CheckoutStartRequestParams)`

All callers (4 Svelte components, PaddleService) and test assertions updated to match.

## Not in scope

- `purchaseOption` / `presentedOfferingContext` param order is inconsistent between `checkoutStart` and `postCheckoutStart` — noted for future cleanup
- Presentation context fields (`presentedStepId`, `paywallId`) could move into `PresentedOfferingContext` for better domain modeling — separate PR

## Breaking changes

None. Both methods are internal — never called directly by SDK consumers. The public API (`purchases.purchase(params)`) is not affected.

## Test plan

- [x] All 547 existing tests pass
- [x] Lint, prettier, eslint, typecheck, svelte-check all pass

### DIFF
--- a/src/helpers/purchase-operation-helper.ts
+++ b/src/helpers/purchase-operation-helper.ts
@@ -112,9 +112,9 @@ interface CheckoutStartParams {
   appUserId: string;
   productId: string;
   purchaseOption: PurchaseOption;
-  presentedOfferingContext: PresentedOfferingContext;
 
   // Presentation context
+  presentedOfferingContext: PresentedOfferingContext;
   workflowPurchaseContext?: WorkflowPurchaseContext;
   paywallId?: string;
 
--- a/src/helpers/purchase-operation-helper.ts
+++ b/src/helpers/purchase-operation-helper.ts
@@ -107,6 +107,22 @@ export class PurchaseFlowError extends Error {
   }
 }
 
+interface CheckoutStartParams {
+  // Purchase identity
+  appUserId: string;
+  productId: string;
+  purchaseOption: PurchaseOption;
+  presentedOfferingContext: PresentedOfferingContext;
+
+  // Presentation context
+  workflowPurchaseContext?: WorkflowPurchaseContext;
+  paywallId?: string;
+
+  // Customer data
+  customerEmail?: string;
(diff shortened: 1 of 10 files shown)


src/helpers/purchase-operation-helper.ts
```diff
@@ -107,6 +107,22 @@ export class PurchaseFlowError extends Error {
   }
 }
 
+interface CheckoutStartParams {
+  // Purchase identity
+  appUserId: string;
+  productId: string;
+  purchaseOption: PurchaseOption;
+  presentedOfferingContext: PresentedOfferingContext;
```
tonidero: I wonder if this should be part of the `Presentation context` group below?

tonidero (approved):
Looks great to me! Nice refactor!

src/helpers/purchase-operation-helper.ts
vkartaviy: Good call 👍🏼 - moved `presentedOfferingContext` to the "Presentation context" group in both interfaces. It's semantically context about how the offering was presented, not purchase identity. Fixed in 301e893.                                                                                  
                                                                     
As a future follow-up (out of scope for this PR): `presentedStepId` and `paywallId` could potentially be folded into `PresentedOfferingContext` itself, since they're all about "how the user got to checkout." That would clean up the params further but touches a shared type used across many call sites.

PR #828, opened 2026-04-03
WEB-3903: Stripe Checkout branding support

nicfix:
## Motivation / Description
Stripe Checkout branding and layout handling in `purchases-js` had drifted from the backend contract and introduced redundant/unused data paths.  
This PR simplifies the Stripe flow so UI branding comes from the shared branding context (backend source of truth), improves fullscreen/mobile behavior, and removes the temporary sandbox-only Stripe key restriction.

## Changes introduced
- Enabled Stripe for all valid `strp_` API keys (not only `strp_sb_` sandbox keys).
- Updated API-key validation/routing tests to reflect live Stripe key support.
- Removed unused Stripe checkout response fields from client typings:
  - `stripe_billing_params.branding_settings`
  - `stripe_billing_params.appearance`
- Simplified Stripe branding handling in UI:
  - stop merging/overriding branding from checkout-start response fields that are no longer used
  - rely on the shared branding context for appearance theming.
- Refactored Stripe fullscreen layout behavior:
  - removed unused fullscreen header rendering logic
  - kept fullscreen content centered and responsive
  - improved wrapper sizing behavior (`min-height` flow).
- Added mobile horizontal padding to message layout content/footer to avoid full-bleed CTA/controls on narrow screens.
- Introduced explicit `"purchasing"` page state and adjusted loading display logic accordingly.
- Minor checkout container cleanup (removed unnecessary `flex: 1` in Stripe checkout page container).
## Linear ticket (if any)
WEB-3903

### DIFF
--- a/src/helpers/api-key-helper.ts
+++ b/src/helpers/api-key-helper.ts
@@ -16,11 +16,7 @@ export function isPaddleApiKey(apiKey: string): boolean {
 }
 
 export function isStripeApiKey(apiKey: string): boolean {
-  const isStripeApiKey = apiKey ? stripe_api_key_regex.test(apiKey) : false;
-  const isSandboxStripeApiKey = apiKey ? apiKey.startsWith("strp_sb_") : false;
-  // Supporting only sandbox api keys for now, we'll add support for live api keys
-  // when we make the feature generally available
-  return isStripeApiKey && isSandboxStripeApiKey;
+  return apiKey ? stripe_api_key_regex.test(apiKey) : false;
 }
 
 export function isSimulatedStoreApiKey(apiKey: string): boolean {
--- a/src/networking/responses/checkout-start-response.ts
+++ b/src/networking/responses/checkout-start-response.ts
@@ -1,21 +1,10 @@
 import type { GatewayParams } from "./stripe-elements";
-import type { BrandingAppearance } from "../../entities/branding";
 
 export interface StripeBillingParams {
   client_secret: string;
   environment: string;
   publishable_api_key: string;
   stripe_account_id: string;
-  branding_settings?: {
-    background_color?: string;
-    border_style?: string;
-    button_color?: string;
-    display_name?: string;
-    font_family?: string;
-    icon?: { file?: string; type?: string };
-    logo?: { file?: string; type?: string };
-  } | null;
-  appearance?: Partial<BrandingAppearance> | null;
 }
 
 export interface WebBillingCheckoutStartResponse {
--- a/src/ui/stripe-checkout-purchases-ui-inner.svelte
+++ b/src/ui/stripe-checkout-purchases-ui-inner.svelte
@@ -16,6 +16,7 @@
   import ColLayout from "./layout/col-layout.svelte";
   import type { StripeBillingParams } from "../networking/responses/checkout-start-response";
   import type { BrandingAppearance } from "../entities/branding";
+  import { Theme } from "./theme/theme";
 
   interface Props {
     currentPage: "loading" | "stripe-checkout" | "success" | "error";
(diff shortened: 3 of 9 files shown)


src/ui/stripe-checkout-purchases-ui-inner.svelte
```diff
@@ -15,12 +15,17 @@
...
+  import { brandingContextKey } from "./constants";
 
   interface Props {
-    currentPage: "loading" | "stripe-checkout" | "success" | "error";
+    currentPage:
+      | "loading"
+      | "stripe-checkout"
+      | "success"
+      | "error"
+      | "purchasing";
```
nicfix, line 27: Added to distinguish between the initial loading and the loading while waiting for purchase confirmation

src/ui/stripe-checkout-purchases-ui.svelte
```diff
@@ -227,7 +216,6 @@
 <StripeCheckoutPurchasesUiInner
   {currentPage}
   {brandingInfo}
-  brandingAppearance={mergedBrandingAppearance}
```
nicfix, line 230: Relying on the one set up in the context instead of passing it down through props

src/networking/responses/checkout-start-response.ts
```diff
@@ -1,21 +1,10 @@
...
-  branding_settings?: {
-    background_color?: string;
-    border_style?: string;
-    button_color?: string;
-    display_name?: string;
-    font_family?: string;
-    icon?: { file?: string; type?: string };
-    logo?: { file?: string; type?: string };
-  } | null;
-  appearance?: Partial<BrandingAppearance> | null;
```
nicfix, line 18: These fields are not needed (and won't be returned by the backend anymore) since they are used only in the backend to set-up the appearance of the checkout session.

Not in the SDK.

src/helpers/api-key-helper.ts
```diff
@@ -16,11 +16,7 @@ export function isPaddleApiKey(apiKey: string): boolean {
 }
 
 export function isStripeApiKey(apiKey: string): boolean {
-  const isStripeApiKey = apiKey ? stripe_api_key_regex.test(apiKey) : false;
-  const isSandboxStripeApiKey = apiKey ? apiKey.startsWith("strp_sb_") : false;
-  // Supporting only sandbox api keys for now, we'll add support for live api keys
-  // when we make the feature generally available
-  return isStripeApiKey && isSandboxStripeApiKey;
+  return apiKey ? stripe_api_key_regex.test(apiKey) : false;
```
nicfix, line 19: Opening Stripe to anyone on SDK side.

james-od (approved):
Code LGTM, on the denied snapshots - assume we don't want to narrow the view there?

src/ui/stripe-checkout-purchases-ui.svelte
```diff
@@ -74,33 +74,22 @@
   setContext(translatorContextKey, translatorStore);
   setContext(eventsTrackerContextKey, eventsTracker);
 
+  const brandingAppearanceStore = writable<BrandingAppearance | null>(
```
lburdock, line 77: We switched brandingContext to a writable store when Stripe Checkout was returning appearance info from the checkoutStart call. Now that we've switched back to brandingInfo.appearance from the normal branding fetch, do we still need this to be a store or should we convert back to plain context?
https://github.com/RevenueCat/purchases-js/pull/768/changes#diff-3bc74c66a768131a2da4acfe39d2bd99f94418bc8c931f2b461f37a063b52916R29

(This is not urgent for this PR, it can be done in a future PR if needed!)

nicfix: Yeah I was confused as well tbh.
I think we can clean it up in a follow-up but we should either go for context everywhere or not.
The branding gets loaded also from the calls in the decorator functions so I think that using the store and listening to changes might be the best approach.

nicfix:
Yeah, I denied them since I thought my changes would affect only paddle and Stripe but then I noticed that they affect also rcbilling and I denied them, I'll fix it by adding the padding to the container rather than the content.

PR #837, opened 2026-04-10
Track paywall component interaction events

JZDesign:
## Summary
- extend the paywall event model with `paywall_component_interacted` and serialize the full interaction payload
- bridge `onComponentInteraction` from `presentPaywall()` into paywall analytics with lifecycle gating and package-to-product enrichment
- add serialization and `presentPaywall()` coverage for interaction tracking, enrichment, and before-impression / after-close suppression

## Companion work
- companion `purchases-ui-js` PR: https://github.com/RevenueCat/purchases-ui-js/pull/249
- this PR adds the analytics model and bridge, but production behavior still depends on consuming a `purchases-ui-js` release that includes those callback emitters

## Decisions made
- interactions are only tracked after `paywall_impression` and are dropped after `paywall_close`
- interaction events reuse the paywall session/display metadata owned by `purchases-js`
- package identifiers are enriched into product identifiers from the current offering when the UI payload does not already provide product ids
- `purchase_button.component_value` remains `\"unknown\"` until the UI schema exposes purchase-method metadata
- `package_selection_sheet` remains unimplemented for web until the UI package can mark those sheets explicitly

## Remaining choices / follow-ups
- after the companion UI PR lands, `purchases-js` still needs to consume a released/bumped `@revenuecat/purchases-ui-js` version before the new events fire end-to-end in shipped builds
- when web adds explicit sheet analytics roles, this bridge is ready to accept `package_selection_sheet` payloads
- if the UI starts providing purchase-button method values or checkout URLs, this event model can carry them without another wire-format change

## Test plan
- [x] `pnpm run typecheck`
- [x] `pnpm run test:typecheck`
- [x] `pnpm exec vitest run src/tests/behavioral-events/paywall-event.test.ts src/tests/present-paywall.events.test.ts`


MonikaMateska (approved):
Thanks for the implementation, the code looks good to me!
I’m approving from a general readability & structure perspective, but it would be better to have someone who knows this repo and the paywall analytics pipeline more deeply give a pass as well.

src/behavioural-events/paywall-event.ts
```diff
@@ -3,20 +3,74 @@ import { generateUUID } from "../helpers/uuid-helper";
...
+}
+
+interface PaywallCloseOrCancelEventData extends BasePaywallEventData {
+  type: "paywall_close" | "paywall_cancel";
+}
+
+export interface PaywallComponentInteractionEventData
+  extends BasePaywallEventData,
+    ComponentInteractionData {
+  type: "paywall_component_interacted";
```
RosieWatson, line 68: I assume this might already be set in stone but would `paywall_component_interaction` not make more sense?

src/behavioural-events/paywall-event.ts
```diff
@@ -56,12 +129,69 @@ export class PaywallEvent {
       paywall_rc_public_id: this.data.paywallRcPublicId,
     };
 
-    if (this.data.displayMode !== undefined) {
+    if ("displayMode" in this.data && this.data.displayMode !== undefined) {
```
RosieWatson: This extra check seems a bit redundant to me, unless you specifically need to distinguish between missing property vs present but undefined - which I don't think is the case


```suggestion
    if (this.data.displayMode !== undefined) {
```

src/behavioural-events/paywall-event.ts
```diff
@@ -56,12 +129,69 @@ export class PaywallEvent {
...
+      if (this.data.destinationIndex !== undefined) {
+        payload.destination_index = this.data.destinationIndex;
+      }
+      if (this.data.originContextName !== undefined) {
+        payload.origin_context_name = this.data.originContextName;
+      }
+      if (this.data.destinationContextName !== undefined) {
+        payload.destination_context_name = this.data.destinationContextName;
+      }
+      if (this.data.defaultIndex !== undefined) {
```
RosieWatson: This whole if statement makes me a bit cautious, as I think it's super easy to accidentally change something with little consequence - could we make an object that we can map over? Would add some type safety for the future and potentially make it easier to parse. E.g.:

```
const INTERACTION_FIELD_MAP = {
  componentType: "component_type",
  componentName: "component_name",
  componentValue: "component_value",
  componentURL: "component_url",
  originIndex: "origin_index",
  destinationIndex: "destination_index",
  originContextName: "origin_context_name",
  destinationContextName: "destination_context_name",
  defaultIndex: "default_index",
  originPackageIdentifier: "origin_package_id",
  destinationPackageIdentifier: "destination_package_id",
  defaultPackageIdentifier: "default_package_id",
  originProductIdentifier: "origin_product_id",
  destinationProductIdentifier: "destination_product_id",
  defaultProductIdentifier: "default_product_id",
  currentPackageIdentifier: "current_package_id",
  resultingPackageIdentifier: "resulting_package_id",
  currentProductIdentifier: "current_product_id",
  resultingProductIdentifier: "resulting_product_id",
} as const satisfies Record<
  keyof ComponentInteractionData,
  keyof PaywallEventPayload
>;
```

and then a function like

```
if (this.data.type === "paywall_component_interacted") {
  for (const [srcKey, destKey] of Object.entries(INTERACTION_FIELD_MAP)) {
    const value = this.data[srcKey as keyof ComponentInteractionData];
    if (value !== undefined) {
      (payload[destKey] as typeof value) = value;
    }
  }
}
```

src/behavioural-events/paywall-event.ts
```diff
@@ -30,6 +84,25 @@ type PaywallEventPayload = {
...
+  origin_package_id?: string;
+  destination_package_id?: string;
+  default_package_id?: string;
+  origin_product_id?: string;
+  destination_product_id?: string;
+  default_product_id?: string;
+  current_package_id?: string;
+  resulting_package_id?: string;
+  current_product_id?: string;
+  resulting_product_id?: string;
```
RosieWatson: Are these changes relevant to every `PaywallEventPayload` or could we maybe have a type that is more specific for this event type so they don't need to be applied as optional to all the existing ones?

src/main.ts
```diff
@@ -592,8 +595,27 @@ export class Purchases {
...
-    const trackPaywallEvent = (type: PaywallEventType) => {
+    const getProductIdentifierForPackageId = (packageId?: string) => {
+      if (packageId === undefined) {
+        return undefined;
+      }
+
+      return offering.availablePackages.find(
+        (pkg) => pkg.identifier === packageId,
+      )?.webBillingProduct.identifier;
+    };
```
RosieWatson, line 647: Could we reduce how much we have to run this function by creating a map that we can then just reference instead? E.g.

```
const productIdsByPackage = new Map(
  offering.availablePackages.map((p) => [p.identifier, p.webBillingProduct.identifier]),
);
```

src/main.ts
```diff
@@ -798,11 +849,15 @@ export class Purchases {
           hideBackButtons: paywallParams.hideBackButtons,
           walletButtonRender,
           customVariables: paywallParams.customVariables,
+          onComponentInteraction: (data: ComponentInteractionData) => {
+            trackComponentInteraction(data);
+          },
```
RosieWatson:
```suggestion
          onComponentInteraction: trackComponentInteraction;
          },
```

src/main.ts
```diff
@@ -692,14 +710,47 @@ export class Purchases {
...
+
+        this.eventsTracker.trackPaywallEvent({
+          type: "paywall_component_interacted",
+          appUserId: this._appUserId,
+          sessionId: paywallSessionId,
+          offeringId: offering.identifier,
+          paywallRevision: 0,
+          paywallRcPublicId: offering.paywallComponents?.id ?? null,
+          ...paywallDisplayData,
+          ...data,
```
RosieWatson: With this - we are just silently passing everything through that might be added in `purchases-ui-js` which is maybe fine (honestly not strongly opinionated on this) but could also be safer to have a formed function that makes sure we define things more - like:

```
const toInteractionEvent = (
  data: ComponentInteractionData,
): PaywallComponentInteractionEventData => ({
  type: "paywall_component_interacted",
  ...base,
  ...paywallDisplayData,
  componentType: data.componentType,
  componentName: data.componentName,
  componentValue: data.componentValue,
  componentURL: data.componentURL,
  originIndex: data.originIndex,
  // ...etc
  originProductIdentifier:
    data.originProductIdentifier ?? productIdFor(data.originPackageIdentifier),
  // ...etc
});
```

src/behavioural-events/paywall-event.ts
JZDesign: There is a whole thread in slack where we started with something more like that, and ended up here.

I see both sides. But at this point, it's set.

src/behavioural-events/paywall-event.ts
JZDesign: Refactored this into `toDisplayPayload()`, so the redundant `in` check is gone as part of the broader serializer cleanup.

src/behavioural-events/paywall-event.ts
JZDesign: Applied this pattern with an `INTERACTION_FIELD_MAP` plus a helper that maps the interaction fields into the serialized payload.

src/behavioural-events/paywall-event.ts
JZDesign: Split the payload typing so the interaction-only fields now live on a dedicated interaction payload type instead of the shared paywall payload shape.

src/main.ts
JZDesign: Done. I now precompute `productIdsByPackage` once and use that map for the package-to-product enrichment.

src/main.ts
JZDesign: Kept a small wrapper here because we still need to capture the latest navigation interaction before tracking, for the text-link compatibility handling. The tracking itself stays isolated in `trackComponentInteraction()`.

src/main.ts
JZDesign: Done. I added `toInteractionEvent()` so we explicitly build the tracked event instead of forwarding `...data` through.

src/behavioural-events/paywall-event.ts
```diff
@@ -101,16 +101,16 @@ type PaywallComponentInteractionPayload = {
...
-  origin_package_id?: string;
-  destination_package_id?: string;
-  default_package_id?: string;
-  origin_product_id?: string;
-  destination_product_id?: string;
-  default_product_id?: string;
-  current_package_id?: string;
-  resulting_package_id?: string;
-  current_product_id?: string;
-  resulting_product_id?: string;
```
JZDesign: Great catch!

RosieWatson (approved):
nothing but perfection

PR #838, opened 2026-04-13
FUN-2156: Update purchases-js to support new onCompleteWorkflowNavigate param

francocorreasosa:
### TL;DR

Upgraded `@revenuecat/purchases-ui-js` to version 3.10.0 and added support for complete workflow navigation handling in paywalls.

### What changed?

- Upgraded `@revenuecat/purchases-ui-js` from version 3.9.2 to 3.10.0
- Added `onCompleteWorkflowNavigate` callback parameter to `PresentPaywallParams` interface for handling complete workflow button navigation with configurable methods (deep_link, external_browser, in_app_browser)
- Implemented default navigation behavior in the paywall presentation logic that opens external URLs in new tabs for external browser method and navigates in the same tab for other methods
- Added `appUserId` to the paywall component props

### How to test?

1. Present a paywall that includes a complete workflow button (exit URL)
2. Test the default behavior without providing `onCompleteWorkflowNavigate` callback
3. Test custom navigation behavior by providing the `onCompleteWorkflowNavigate` callback
4. Verify different navigation methods work correctly (external_browser opens new tab, others navigate in same tab)

### Why make this change?

This change provides developers with more control over how users are navigated when completing workflows in paywalls, allowing for custom handling of exit URLs while maintaining sensible defaults for different navigation methods.

### DIFF
--- a/src/entities/present-paywall-params.ts
+++ b/src/entities/present-paywall-params.ts
@@ -1,5 +1,10 @@
 import type { Offering } from "./offerings";
-import type { CustomVariables } from "@revenuecat/purchases-ui-js";
+import type {
+  CompleteWorkflowNavigateArgs,
+  CustomVariables,
+} from "@revenuecat/purchases-ui-js";
+
+export type { CompleteWorkflowNavigateArgs };
 
 /**
  * Parameters for the {@link Purchases.presentPaywall} method.
@@ -34,6 +39,16 @@ export interface PresentPaywallParams {
    * Callback to be called when the paywall tries to navigate to an external URL.
    */
   readonly onNavigateToUrl?: (url: string) => void;
+
+  /**
+   * Called when the paywall uses a complete_workflow button (exit URL).
+   * If omitted, the SDK opens the URL: new tab for `external_browser`, same tab for
+   * `in_app_browser` / `deep_link`.
+   * @internal
+   */
+  readonly onCompleteWorkflowNavigate?: (
+    args: CompleteWorkflowNavigateArgs,
+  ) => void | Promise<void>;
 
   /**
    * Callback to be called when the paywall tries to navigate back.
--- a/api-report/purchases-js.api.json
+++ b/api-report/purchases-js.api.json
@@ -4007,7 +4007,7 @@
             {
               "kind": "PropertySignature",
               "canonicalReference": "@revenuecat/purchases-js!PresentPaywallParams#customVariables:member",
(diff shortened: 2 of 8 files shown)


francocorreasosa:
* **#838** <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-js/838?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/RevenueCat/purchases-js/838?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

nicfix (approved):

src/entities/present-paywall-params.ts
```diff
@@ -1,6 +1,24 @@
 import type { Offering } from "./offerings";
 import type { CustomVariables } from "@revenuecat/purchases-ui-js";
 
+/**
+ * How the paywall should navigate when a complete workflow (exit) URL is used.
+ * @public
```
tonidero: Does this need to be public? Or could it be `@internal`? Same for the other new types below

src/entities/present-paywall-params.ts
```diff
@@ -35,6 +53,15 @@ export interface PresentPaywallParams {
    */
   readonly onNavigateToUrl?: (url: string) => void;
 
+  /**
+   * Called when the paywall uses a complete_workflow button (exit URL).
+   * If omitted, the SDK opens the URL: new tab for `external_browser`, same tab for
+   * `in_app_browser` / `deep_link`.
+   */
+  readonly onCompleteWorkflowNavigate?: (
+    args: CompleteWorkflowNavigateArgs,
```
tonidero, line 50: Just to check, but will this always exist? As in, we will always have an exit url? I believe we should... but asking just in case, since we might want to remain in a "Successful" page at some point, and these arguments should be optional there I guess?

src/entities/present-paywall-params.ts
```diff
@@ -35,6 +53,15 @@ export interface PresentPaywallParams {
    */
   readonly onNavigateToUrl?: (url: string) => void;
 
+  /**
+   * Called when the paywall uses a complete_workflow button (exit URL).
+   * If omitted, the SDK opens the URL: new tab for `external_browser`, same tab for
+   * `in_app_browser` / `deep_link`.
+   */
+  readonly onCompleteWorkflowNavigate?: (
```
tonidero, line 49: TBH, my main question to make this public is because I don't know if we want to expose this as part of the `presentPaywall` API, or whether we should create a specific `presentWorkflow` API in the future as part of the SDK (I know right now that's separate) that handles all this... Until then, I think we might want to consider all this internal, since I don't think it will be useful for developers using the SDK directly until then.

tonidero (commented):
Mostly some questions related to the API 🙇

src/entities/present-paywall-params.ts
francocorreasosa: yep you are totally right this should be internal only

PR #844, opened 2026-04-16
Only preload Stripe module for Stripe-based API keys

tonidero:
## Summary
- Skip Stripe module preloading during configuration for API keys that don't use Stripe (Paddle, simulated store)
- Only preload when the API key is a web billing (`rcb_`) or Stripe billing (`strp_`) key and `storeLoadTime` is `"configuration"`
- Added tests verifying preload behavior for each API key type


james-od (approved):
Smart!

PR #846, opened 2026-04-16
[WEB-3965] Fix offer variables in paywalls

james-od:
## Motivation / Description

Variables definition for reference: https://www.revenuecat.com/docs/tools/paywalls/creating-paywalls/variables

While developing the Discounts project, which allows applying a fixed percentage discount to a web product for a given period of time, we realised that the variables in paywalls weren't working correctly.

I then noticed that intro offers have the exact same problems.

Specifically this solves the tickets in this bucket: https://linear.app/revenuecat/project/flexible-discounts-282eb1531b65/issues

The changes are:
- Account for offer.cycleCount when determining duration based variables
- Forever discounts now no longer populate offer_period* based variables
- Floor() the durationInWeeks when WEEKS_PER_MONTH is used to avoid decimals when multiplying by 4.33. Opting to leave the 4.33 as-is to not affect existing pricing logic. Opting not to Floor() anything else because assume the `period.number` is whole. I think this matches the dashboard preview equivalent.
- Add tests

Two main questions for the paywalls team are:
- Does this generally match the equivalents displayed in the editor previews?
- `offer_period_abbreviated` shows `/mo` for both a `6 month discount period with a monthly cycle`, and for a `6month period paid up front`. The latter case may be a bit misleading but this seems to follow the var definition. I guess you would construct a string based on this + the `offer_period_in_months`.

I do worry a little bit about duplicating ~complex logic across here, the paywall previews and the backend meaning bugs like e.g this https://github.com/RevenueCat/revenuecat-app/pull/7698. But thats a broader change.

## Testing

I've used this paywall https://app.revenuecat.com/projects/e53bf105/paywalls/pw7013f7d754a84855/builder which logs the values of all relevant variables, with this offering https://app.revenuecat.com/projects/e53bf105/product-catalog/offerings/ofrnga9154231f1, with a set of different products. The variables now all seem to behave as I'd expect.
Unit tests are I think reasonably extensive.

### DIFF
--- a/src/helpers/paywall-offer-helpers.ts
+++ b/src/helpers/paywall-offer-helpers.ts
@@ -7,10 +7,7 @@ import type {
 } from "../entities/offerings";
 import { type Translator } from "../ui/localization/translator";
 import { getNextRenewalDate, type Period } from "./duration-helper";
-import {
-  getDurationInDays,
-  getPeriodVariables,
-} from "./paywall-period-helpers";
+import { getPeriodVariables } from "./paywall-period-helpers";
 import { getPriceVariables } from "./paywall-price-helpers";
 
 type OfferPhase = PricingPhase | DiscountPhase;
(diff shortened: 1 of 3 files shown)


james-od:
@cursor review pls

RosieWatson (approved):
I _think_ this should match well with what we show in the dashboard, I think my knowledge of discounts is a bit weak, but from my playing around it looks good

james-od:
Thanks @RosieWatson, much appreciated!

PR #847, opened 2026-04-17
Guard against accidental version bumps outside release branches

tonidero:
## Summary

Adds two complementary guardrails so contributors can't accidentally modify the version manually, and so drift between the three version sources (`.version`, `package.json`, `src/helpers/constants.ts`) is caught in CI:

- **Danger rule** (local `Dangerfile`) — fails PRs that touch the `version` line in `package.json`, `.version`, or `src/helpers/constants.ts` when the head branch is **not** `release/*`. Error message points contributors to `bundle exec fastlane bump`.
- **CI consistency check** — new step in the existing `test` job (`pnpm run check-version-consistency`) runs `scripts/check-version-consistency.mjs`, which reads the three files and fails if they disagree. Runs regardless of branch, so it also catches drift on release branches.

The Fastlane `bump` lane already updates all three files atomically via the `files_with_version_number` map in `fastlane/Fastfile`, so no changes are needed on the release side.

### Why

Contributors have occasionally edited `package.json`'s `version` directly in feature PRs (historical examples around PRs #824 / #835), thinking that's how releases are cut. That causes drift from `.version`, which is Fastlane's source of truth.

## Test plan

- [ ] CI: `danger` job passes on this PR (no version lines touched).
- [ ] CI: `test` job's new "Check version consistency" step passes on this PR.
- [ ] Manual sanity: push a scratch commit that bumps only `package.json` version → both Danger and the CI step should fail with the instructive messages. Revert before merging.
- [ ] On a `release/*` branch, the Danger rule should skip (allow the bump) while the CI consistency step still passes as long as the three files match.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -142,6 +142,9 @@ jobs:
       - revenuecat/setup-git-credentials
       - install-dependencies
       - run:
+          name: Check version consistency
+          command: pnpm run check-version-consistency
+      - run:
           name: Install dependencies webbilling-demo
           working_directory: examples/webbilling-demo
           command: pnpm install
--- a/Dangerfile
+++ b/Dangerfile
@@ -22,8 +22,9 @@ unless head_branch.to_s.start_with?('release/')
 
     fail(
       "`#{path}` contains a version change, but this PR is not targeting a `release/*` branch. " \
-      "Versions are managed by Fastlane — please revert the change and run `bundle exec fastlane bump` " \
-      "to cut a release. See `fastlane/Fastfile` for the release flow."
+      "Versions are managed by Fastlane — please revert the change and either run `bundle exec fastlane bump` " \
+      "locally, or trigger the `manual-trigger-bump` pipeline in CircleCI with the `action` parameter set to " \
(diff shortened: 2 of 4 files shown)


rickvdl (commented):
Good idea, thanks for adding! 

Two questions:
- Danger isn't a required step, so we could probably still pass the PR (accidentally?). Or do you think this extra notice will alert us more than (only the) diff did?
- If we ever need to make manual changes ourselves we'd just do it on the release branch then?

tonidero:
> Danger isn't a required step, so we could probably still pass the PR (accidentally?). Or do you think this extra notice will alert us more than (only the) diff did?

Good point... I actually just made Danger required after your comment. I think it should pass, no reason to merge it if it's failing I think

> If we ever need to make manual changes ourselves we'd just do it on the release branch then?

Pretty much... I don't think there is any case where we need to change the version numbers other than during releases? If we need to modify the format or something like that on the versions, this could be done in a "normal" branch and we would need to bypass it, but I think that should be quite uncommon and something we can handle when we get there IMO.

ajpallares (approved):
Very useful to prevent future issues. Thank you! 🙌

PR #849, opened 2026-04-21
[WEB-4171] Discount input

james-od:
## Motivation / Description

Note: Some translations are missing due to rate limits with the existing script. I'm opting to proceed with merging this regardless in order to unblock, with a follow up raised to resolve the translation generation and apply to the missing strings: https://linear.app/revenuecat/issue/WEB-4205/sdk-fix-translations

Figma: https://www.figma.com/design/RYB9cjbSZCf1mroI0bkBbK/WEB-Billing?node-id=15700-17102&m=dev

Adds the option to show a Discount Code input field, ultimately passed from purchases.purchase() or purchases.presentPaywall().

https://github.com/user-attachments/assets/68503b34-b5d9-40d3-b30b-cbae86726c37

The trickiest part here is ensuring that the products, taxes, etc are all up to date given the discount.
So, when a code is added or removed we now:
- Refetch /products
- Call checkout/start again

We expose `showDiscountCodeField`, `discountCode` and `onDiscountCodeChanged` to allow callers control over how the discount code is provided (e.g via URL), updates when it changes (e.g updating the URL) and control over defaults, and whether or not to display the input at all.

## Changes introduced

- New warning icon, plus added as an `Icon`
- Stories to cover various visual cases for the discount input
- Unit tests
- Following the designs, the "Hide details" control button defined in PricingDropdown now says "Add promo code" instead of "Show details" if a discount code input is available. This required generating additional translations.
- The discount input is shown within the pricing table, as per the spec. In both chip and input forms.

## Additional comments

- Tested against Google Pay Wallet button, Apple Pay I had a bit of a nightmare with locally - can double check after merging


src/ui/purchases-ui-inner.svelte
```diff
@@ -74,6 +94,22 @@
...
+
+    priceBreakdown = defaultPriceBreakdown ?? {
+      currency: updatedInitialPrice.currency,
+      totalAmountInMicros: updatedInitialPrice.amountMicros,
+      totalExcludingTaxInMicros: updatedInitialPrice.amountMicros,
+      taxCalculationStatus: "unavailable",
+      taxAmountInMicros: null,
+      taxBreakdown: null,
+    };
+  });
```
james-od, line 111: Resets priceBreakdown whenever productDetails or purchaseOptionToUse changes, so the summary no longer shows pricing from the previous checkout state. This rebuilds a fallback breakdown from the newly selected product/option, with tax fields cleared to an initial “unavailable” state. That gives the UI a clean baseline until PaymentEntryPage recalculates tax and pushes an updated breakdown back up.

src/main.ts
```diff
@@ -884,6 +889,49 @@ export class Purchases {
     return toOfferings(offeringsResponse, productsResponse);
   }
 
+  public async _getProductWithDiscountCode(
```
james-od, line 1114: We need a way to be able to fetch a fresh product payload after a discount has been applied.

src/ui/purchases-ui.svelte
```diff
@@ -146,66 +172,188 @@
       : false,
   );
 
-  onMount(async () => {
-    if (productId === null) {
-      handleError(
+  const startCheckout = (
```
james-od, line 175: Most of this is split out from the existing onMount behaviour, so that we can now also (re)start the checkout after a discount code has been applied.

src/entities/present-paywall-params.ts
```diff
@@ -35,11 +35,26 @@ export interface PresentPaywallParams {
...
+   * outside of the paywall UI, for example in the hosting page's URL.
+   */
+  readonly discountCode?: string;
+
+  /**
+   * Called when the applied discount code changes in the checkout shown from
+   * the paywall. This can be used to sync host state such as URL parameters.
+   */
+  readonly onDiscountCodeChanged?: (discountCode: string | null) => void;
+
```
vicfergar, line 59: Since discounts are still in private beta, should we mark these new fields as `@experimental`?
```suggestion
  /**
   * If set to true, the Web Billing checkout shown from the paywall
   * will display a discount input code field.
   */
  @experimental
  readonly showDiscountCodeField?: boolean;

  /**
   * Initial discount code to apply to the checkout when one already exists
   * outside of the paywall UI, for example in the hosting page's URL.
   */
  @experimental
  readonly discountCode?: string;

  /**
   * Called when the applied discount code changes in the checkout shown from
   * the paywall. This can be used to sync host state such as URL parameters.
   */
  @experimental
  readonly onDiscountCodeChanged?: (discountCode: string | null) => void;

```

src/entities/purchase-params.ts
```diff
@@ -73,6 +73,25 @@ export interface PurchaseParams {
...
+   * for example from a URL parameter.
+   */
+  discountCode?: string;
+
+  /**
+   * Called when the applied discount code changes in the Web Billing checkout.
+   * This can be used by host applications to keep external state, such as the URL,
+   * in sync with the checkout.
+   */
+  onDiscountCodeChanged?: (discountCode: string | null) => void;
```
vicfergar, line 96: Same. Should we mark them as `@experimental`?

```suggestion
  /**
   * If set to true, the Web Billing checkout will show a discount input code field.
   */
  @experimental
  showDiscountCodeField?: boolean;

  /**
   * Initial discount code to display as applied in the Web Billing checkout.
   * This is useful when the code originated outside of the checkout UI,
   * for example from a URL parameter.
   */
  @experimental
  discountCode?: string;

  /**
   * Called when the applied discount code changes in the Web Billing checkout.
   * This can be used by host applications to keep external state, such as the URL,
   * in sync with the checkout.
   */
  @experimental
  onDiscountCodeChanged?: (discountCode: string | null) => void;
```

src/main.ts
vicfergar: I think we can include this info as part of the `PATH /checkout/<operation_session_id>` response, wdyt?

src/ui/atoms/icon.svelte
```diff
@@ -23,7 +24,8 @@
     | "chevron-down"
     | "info"
     | "back"
-    | "close";
+    | "close"
+    | "warning";
```
vicfergar, line 28: Could we add a story for this new icon?

vicfergar (approved):
LGTM! One thing I noticed is that some scenarios for the variant `PaymentEntryPage` are missing when `showDiscountCodeField` is enabled.

src/entities/present-paywall-params.ts
james-od: 100% https://github.com/RevenueCat/purchases-js/pull/849/commits/511350e187b7782491e0dfba3f93be31b7dad042

src/entities/purchase-params.ts
james-od: 100% https://github.com/RevenueCat/purchases-js/pull/849/commits/511350e187b7782491e0dfba3f93be31b7dad042

src/ui/atoms/icon.svelte
james-od: Done! https://github.com/RevenueCat/purchases-js/pull/849/commits/511350e187b7782491e0dfba3f93be31b7dad042

src/main.ts
james-od: Yeah I think that performing this in the backend somewhere makes sense for sure, but will roll that into a followup to unblock other dependent work.

PR #852, opened 2026-04-23
[WEB-4179] Add public paywall availability to offerings

nicfix:
## Summary
This change adds a public hasPaywall boolean to Offering.

Today, web integrators can pass an Offering into presentPaywall(...), but there is no supported way to tell in advance whether that offering actually has a paywall attached. The SDK already knows this internally through paywallComponents, but that field is intentionally excluded from the public type surface. In practice, that leaves developers with two bad options: either call presentPaywall(...) and handle the runtime error, or reach into internal, unsupported fields.

hasPaywall closes that gap by exposing the one piece of information applications actually need, without exposing the internal paywall payload itself.

## Changes
- add a public `hasPaywall` boolean to `Offering`
- derive it from the internal paywall payload without exposing internal fields


nicfix:
@codex review

james-od (approved):

PR #861, opened 2026-04-27
Add express checkout theme support

marek-dabek:
## Motivation / Description

Add a way to pass express checkout button theme to button renderer.

Frontend PR: https://github.com/RevenueCat/revenuecat-app/pull/9434
purchases-ui-js PR: https://github.com/RevenueCat/purchases-ui-js/pull/271

## Changes introduced

* Add an optional parameter to express checkout rendering function
* Pass the style parameter to Stripe

## Linear ticket (if any)

Related to: https://linear.app/revenuecat/issue/FUN-2144/express-checkout-button-theme-support

### DIFF
--- a/src/entities/present-express-purchase-button-params.ts
+++ b/src/entities/present-express-purchase-button-params.ts
@@ -1,7 +1,7 @@
-import type { Package, PurchaseMetadata, PurchaseOption } from "./offerings";
+import type { WalletButtonStyleType } from "@revenuecat/purchases-ui-js";
 import type { CustomTranslations } from "../ui/localization/translator";
+import type { Package, PurchaseMetadata, PurchaseOption } from "./offerings";
 import type { PaywallListener } from "./paywall-listener";
-
 /**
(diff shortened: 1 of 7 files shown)


src/entities/present-express-purchase-button-params.ts
```diff
@@ -70,4 +70,10 @@ export interface PresentExpressPurchaseButtonParams {
    * Optional listener for purchase lifecycle events.
    */
   listener?: PaywallListener;
+  /*
+   * @internal
```
jholwill: question: Is the purchase button internal only? 

If not, will be worth removing this @internal marker so the docs can correctly update

jholwill (approved):

src/entities/present-express-purchase-button-params.ts
marek-dabek: I marked it as internal for now, as we want to use it in funnels. I'd be good to ask the web team if they want to expose it.

PR #863, opened 2026-04-29
[WEB-3962] Show discount percentage and duration at checkout

james-od:
## Motivation / Description

Collection of minor tweaks to discounts side of checkout:

1. We now suffix the discount line item with the % discount, and possibly the discount length. The length is shown only if it exceeds the base billing period (otherwise it will only apply on the first transaction). The brackets are only shown if the discount input/chip is not displayed and we're instead showing the discount name.
```
- Discounts enabled / input enabled (chip)
    - One time = 25% off
    - Discount Period <= Billing Period = 25% off
    - Discount Period >= Billing Period = 25% off for 6 months
    - Forever = 25% off
- Discounts enabled / input not enabled
    - One time = (25% off)
    - Discount Period <= Billing Period = (25% off)
    - Discount Period >= Billing Period = (25% off for 6 months)
    - Forever =  (25% off)
```
2. Collection of tweaks to the input field:
- Increase horizontal padding of chip to 12px to match the "hide details" button padding
- Use `--rc-color-grey-ui-dark` bg colour for "Apply" button bg so it can never blend into the page bg
- Use `--rc-color-grey-text-light` for the focussed border colour of the input, instead of the existing blue
- Match the font of the "Apply" button to the "Pay now" checkout button
- Drop the darkness of the chip border

Note some translations are still missing - this will be addressed soon.


src/ui/molecules/pricing-table.svelte
```diff
@@ -86,13 +86,13 @@
...
 
-  const discountDurationSuffix = $derived.by(() => {
+  const discountSuffix = $derived.by(() => {
     if (
       !promotionalPricePhase ||
       !("durationMode" in promotionalPricePhase) ||
       promotionalPricePhase.durationMode !== "time_window"
     ) {
-      return "";
+      return `${appliedDiscountPercentage}% off`;
```
lburdock, line 97: Do we need any checks to make sure that appliedDiscountPercentage isn't null here?

lburdock (approved):

src/ui/molecules/pricing-table.svelte
james-od: Yep! https://github.com/RevenueCat/purchases-js/pull/863/commits/3258f7b6bbce80bbfe8453bb042a5bf007ebbe94

PR #865, opened 2026-05-01
[WEB-4224] Simplify checkout refresh after applying code

james-od:
## Motivation / Description

Linked with Linked with https://github.com/RevenueCat/khepri/pull/19948

The current approach to discounts in the SDK checkout essentially relies on restarting the checkout. This has a few downsides in UX and tracking.
The idea here is to instead create and call a new endpoint that allows us to mutate the existing session instead of creating a new one.
This directly joins into the existing flow for tax recalculation, and generally shifts more logic off the SDK and into Khepri.

The main change is that pricing is no longer just a tax calculation step. The backend now treats pricing refresh as a stateful checkout operation that can:
- Recalculate tax,
- Resolve and apply/remove a discount code,
- Rersist the selected purchase option onto the checkout session,
- Reuse prior tax location details when the client refreshes pricing without sending new tax inputs,
- Return richer pricing metadata such as original_amount_in_micros, applied_discounts, and interrupt_checkout.

- Adds an in-session checkout repricing flow in khepri via `PATCH /checkout/<operation_session_id>`
There is already a PATCH endpoint here but it isn't actually used at all any more. So I've fully replace the existing functionality there, which was related to Paddle in WPLs.

We also now return original_amount_in_micros, applied_discounts, and interrupt_checkout in order to avoid needing to recalculate pricing semantics from product data plus local assumptions in the client.

The logic we do around tax recalculations with ip address and country code and postal code explicitly passed was a little awkward when expanding the endpoint to also cover discounts. But I think I've landed on something that behaves identically.

## Release Notes

* **New Features**
  * Applied discounts now display detailed information including time-window offers (e.g., "10% off for 3 months").
  * Discount codes can be applied and removed without restarting the checkout flow.

* **Improvements**
  * Enhanced pricing refresh workflow for more reliable calculations and state management.
  * Improved error handling and messaging during checkout operations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/RevenueCat/purchases-js/pull/865)


examples/webbilling-demo/src/tests/helpers/fixtures.ts
vicfergar: nit: Not related to this PR, but since this field is deprecated, it should be safe to remove it from the fixtures.

examples/webbilling-demo/src/tests/tax-calculation.test.ts
```diff
@@ -63,11 +66,15 @@ const mockTaxCalculationRequest = async (
...
+      return;
+    }
     if (!completed) {
       await route.fulfill(fulfillment);
       completed = true;
     } else {
-      route.fallback();
+      await route.fallback();
     }
   });
```
vicfergar: nit: We could extract the request-matching logic into a helper method:

```
async function routePricingRefreshRequest(
  page: Page,
  handler: (route: Route, request: Request) => Promise<object>,
) {
  await page.route(TAX_ROUTE_PATH, async (route) => {
    const request = route.request();

    if (!isPricingRefreshRequest(request)) {
      await route.fallback();
      return;
    }

    await handler(route, request);
  });
}
```

examples/webbilling-demo/src/tests/tax-calculation.test.ts
```diff
@@ -540,6 +565,10 @@ integrationTest.describe("Tax calculation setup errors", () => {
     "Invalid tax origin address",
     async ({ page, userId }) => {
       await page.route(TAX_ROUTE_PATH, async (route) => {
+        if (!isPricingRefreshRequest(route.request())) {
+          await route.fallback();
+          return;
+        }
         await route.fulfill(INVALID_TAX_ORIGIN_RESPONSE);
```
vicfergar, line 562: I think this is outdated now, since we changed the backend response for this case, right?

examples/webbilling-demo/src/tests/helpers/fixtures.ts
james-od: https://github.com/RevenueCat/purchases-js/pull/865/commits/b71ef0904ae11107ede9329fe3089eefdaaf0f67

examples/webbilling-demo/src/tests/tax-calculation.test.ts
james-od: https://github.com/RevenueCat/purchases-js/pull/865/commits/8901d4be8206815ca9a79a4d3340bc46e2d50fcf

examples/webbilling-demo/src/tests/tax-calculation.test.ts
james-od: I think you're right yep!
https://github.com/RevenueCat/purchases-js/pull/865/commits/130ec959dd1911092dbbd56fedefbe9a296b28b1

examples/webbilling-demo/src/tests/tax-calculation.test.ts
```diff
@@ -42,7 +42,26 @@ import {
...
+const TAX_ROUTE_PATH = "**/checkout/*";
+
+const isPricingRefreshRequest = (request: Request) =>
+  request.method() === "PATCH";
+
+async function routePricingRefreshRequest(
+  page: Page,
+  handler: (route: Route, request: Request) => Promise<void>,
+) {
+  await page.route(TAX_ROUTE_PATH, async (route) => {
```
vicfergar: nit
```suggestion
const REFRESH_PRICING_PATH = "**/checkout/*";

const isPricingRefreshRequest = (request: Request) =>
  request.method() === "PATCH";

async function routePricingRefreshRequest(
  page: Page,
  handler: (route: Route, request: Request) => Promise<void>,
) {
  await page.route(REFRESH_PRICING_PATH, async (route) => {
```

src/networking/responses/checkout-pricing-response.ts
```diff
@@ -5,17 +5,29 @@ export interface TaxBreakdown {
...
+  time_window?: string | null;
+}
+
+export enum CheckoutPricingFailedReason {
   tax_collection_disabled = "tax_collection_disabled",
   invalid_tax_location = "invalid_tax_location",
   rate_limit_exceeded = "rate_limit_exceeded",
   missing_required_permission = "missing_required_permission",
   invalid_origin_address = "invalid_origin_address",
+  invalid_head_office_address = "invalid_head_office_address",
```
nicfix, line 25: That's an interesting new error !?

src/tests/helpers/purchase-operation-helper.test.ts
```diff
@@ -409,13 +449,57 @@ describe("PurchaseOperationHelper", () => {
...
+  test("checkoutRefreshPricing interrupts checkout for discount sandbox setup errors in payload", async () => {
+    setCheckoutStartResponse(
+      HttpResponse.json(checkoutStartResponse, {
+        status: StatusCodes.OK,
+      }),
+    );
+    setCheckoutRefreshPricingResponse(
+      HttpResponse.json(
+        {
+          failed_reason: "taxes_not_active",
```
nicfix, line 461: Are we sure we want to interrupt the checkout in this scenario?

What happens if taxes are not active but we just want to apply the discount?

Until now every time we were not able to calculate tax we were allowing the purchase anyway. 

Would this behavior change that?

src/tests/ui/purchases-ui.test.ts
```diff
@@ -305,45 +333,55 @@ describe("PurchasesUI", () => {
...
-        rcPackage.webBillingProduct.price.currency,
-        "SAVE10",
-      );
+      expect(checkoutRefreshPricingSpy).toHaveBeenCalledWith({
+        discountCode: "SAVE10",
+      });
+      expect(onDiscountCodeChanged).toHaveBeenCalledWith("SAVE10");
+      expect(
+        screen.getByRole("button", { name: "Remove promo code SAVE10" }),
+      ).toBeTruthy();
```
vicfergar, line 343: Should we also assert that the pricing summary reflects the discounted price?

src/ui/pages/payment-entry-page.svelte
```diff
@@ -218,32 +250,51 @@
...
-            taxCalculationStatus = "pending";
+            nextTaxCalculationStatus = "pending";
           } else {
-            taxCalculationStatus = "disabled";
+            nextTaxCalculationStatus = "disabled";
           }
         } else {
-          taxCalculationStatus = "calculated";
+          nextTaxCalculationStatus = "calculated";
         }
```
vicfergar, line 267: This code seems duplicated. Could we use the helper `getTaxCalculationStatusForPricingResponse`?

vicfergar (approved):

src/ui/price-breakdown-utils.ts
```diff
@@ -0,0 +1,31 @@
+import type { CheckoutPricingResponse } from "../networking/responses/checkout-pricing-response";
```
nicfix: Nit: Maybe this util should be on networking/data side rather than ui ?

Probably worth having it in the checkout-pricing-response.ts file?

src/networking/responses/checkout-pricing-response.ts
james-od: Was added here https://github.com/RevenueCat/khepri/pull/19065 and is still passed in the refactor https://github.com/RevenueCat/khepri/pull/19948/changes#diff-014d0ef8901204c4126831f8568b042b15ac6559e5db904f6226a0fa00fdb03eR823-R829

src/tests/helpers/purchase-operation-helper.test.ts
james-od: This is only going to affect sandbox, but I think the behaviour is still unchanged.
Old tax calc endpoint behaviour is:
- If production, don't throw at all - allow the purchase
- If sandbox, BE returns a 422 
New endpoint behaviour is:
- If production, don't throw at all - allow the purchase
- If sandbox, BE returns a 200 with `interrupt_checkout: true` and `failed_reason: "taxes_not_active"`

src/ui/pages/payment-entry-page.svelte
james-od: Hmm, I don't really see sufficient overlap in the two methods?

src/ui/price-breakdown-utils.ts
james-od: Added in https://github.com/RevenueCat/purchases-js/pull/865/commits/f56e16d3ae3450eed7eafc537ee3140cc92e8472

src/tests/ui/purchases-ui.test.ts
james-od: Yeah good point, added in https://github.com/RevenueCat/purchases-js/pull/865/commits/f56e16d3ae3450eed7eafc537ee3140cc92e8472

examples/webbilling-demo/src/tests/tax-calculation.test.ts
james-od: Added in https://github.com/RevenueCat/purchases-js/pull/865/commits/f56e16d3ae3450eed7eafc537ee3140cc92e8472

nicfix (approved):

src/tests/helpers/purchase-operation-helper.test.ts
nicfix: Approved then!

PR #868, opened 2026-05-06
Add displayName and price to SubscriptionInfo

andsolleiro:
## Motivation / Description

Standardize the `SubscriptionInfo` object exposed by `purchases-js` with the iOS and Android SDKs. Both mobile SDKs already expose `displayName` and `price` on `SubscriptionInfo`, but the JS SDK does not — the data is in the backend response, just not surfaced.

References:
- iOS: [`SubscriptionInfo.swift`](https://github.com/RevenueCat/purchases-ios/blob/main/Sources/Identity/SubscriptionInfo.swift) (`displayName: String?`, `price: ProductPaidPrice?`)
- Android: [`SubscriptionInfo.kt`](https://github.com/RevenueCat/purchases-android/blob/main/purchases/src/main/kotlin/com/revenuecat/purchases/SubscriptionInfo.kt) (`displayName: String?`, `price: Price?`)

Note on ``period``: neither mobile SDK exposes a separate `period` field on `SubscriptionInfo`. The closest field is `periodType` (the enum `normal | intro | trial | prepaid`), which the JS SDK already exposes — so no change is needed there.

## Changes introduced

- Added `display_name` and `price` to `SubscriberSubscriptionResponse` (the wire response type).
- Added two new public fields to `SubscriptionInfo`:
  - `displayName: string | null`
  - `price: Price | null` — reuses the existing public `Price` type from `offerings.ts` (with `amount`, `amountMicros`, `currency`, `formattedPrice`) so the SDK has a single, consistent price representation. The wire shape is `{ amount: number, currency: string }` (where ``amount`` is in major units, e.g. ``3.59``); we convert it the same way the offerings parser does, using the existing ``formatPrice`` helper for the locale-formatted string.
- Updated 3 existing fixtures in `customer-info.test.ts` and 2 in `main.test.ts` to include the new fields with `null` defaults.
- Added 2 new tests covering: (a) parsing a subscription response that contains `display_name` and `price`, and (b) defaulting both to `null` when the response omits them.
- Regenerated `api-report/purchases-js.api.md` and `purchases-js.api.json` via `pnpm extract-api`.
- Added a changelog entry under `### ✨ New Features`.

## Out of scope

- `autoResumeDate` and `productPlanIdentifier` exist on Android's `SubscriptionInfo` but not iOS; deferring those to a follow-up to keep this PR focused on true cross-platform parity.

## Linear ticket (if any)

N/A

## Additional comments

- `pnpm test` → 576/576 tests pass
- `pnpm typecheck` → clean
- `pnpm lint` (prettier + eslint) → clean

## Summary by CodeRabbit

* **New Features**
  * Subscription data now includes display names for enhanced clarity and improved user experience.
  * Added comprehensive pricing information to subscriptions, including numeric amounts, currency codes, micro-amount precision, and formatted price representation.
  * Updated subscription information structure to provide complete pricing and display data with proper null handling for missing values.


CHANGELOG.latest.md
```diff
@@ -1,2 +1,6 @@
-### 🔄 Other Changes
-* Handle background safe area colors via purchases-ui-js@4.1.0 (#860) via Jamie Holwill (@jholwill)
+## RevenueCat SDK
+### ✨ New Features
+* [WEB-4179] Add public paywall availability to offerings (#852) via Nicola Sacco (@nicfix)
+* Add `displayName` and `price` to `SubscriptionInfo` to align with iOS and Android SDKs
```
james-od: This is automatically generated during releases so shouldn't need manual additions

src/entities/customer-info.ts
```diff
@@ -435,6 +441,21 @@ function toDateIfNotNull(value: string | undefined | null): Date | null {
   return new Date(value);
 }
 
+function toSubscriptionPrice(
+  price: SubscriberSubscriptionPriceResponse | undefined | null,
+): Price | null {
+  if (price == null) return null;
+  // Round to avoid IEEE-754 noise (e.g. 19.99 * 100 = 1998.9999...)
```
james-od: I'm thinking we should add the amount_micros to the response from Khepri instead in order to avoid needing to do this and risk inconsistencies 🤔

src/entities/customer-info.ts
```diff
@@ -435,6 +441,21 @@ function toDateIfNotNull(value: string | undefined | null): Date | null {
   return new Date(value);
 }
 
+function toSubscriptionPrice(
+  price: SubscriberSubscriptionPriceResponse | undefined | null,
+): Price | null {
+  if (price == null) return null;
```
tonidero: We probably also want to return null if it's undefined, maybe this could be 
```suggestion
  if (!price) return null;
```

CHANGELOG.latest.md
tonidero: Agreed, we shouldn't add this here (doesn't actually hurt because it gets overriden on release but still)

src/entities/customer-info.ts
tonidero: And we might want to add a case for this scenario too.

src/entities/customer-info.ts
tonidero: Agreed... but I wouldn't consider it a blocker for now.

tonidero (commented):
Just some small comments, but looks good overall. Thank you!!

src/entities/customer-info.ts
andsolleiro: done

src/entities/customer-info.ts
andsolleiro: done. I reverted that change at SDK level. I understand is something minor anyway.

CHANGELOG.latest.md
andsolleiro: done

james-od (approved):
Nice!

PR #876, opened 2026-05-12
Add /pure entrypoint that resolves @stripe/stripe-js to its /pure variant

tonidero:
## Summary

- Adds a secondary entrypoint, `@revenuecat/purchases-js/pure`, that exposes the same public API as the default but resolves `@stripe/stripe-js` to the [`/pure`](https://docs.stripe.com/sdks/stripejs-versioning#stripejs-pure-loader) variant at bundle time. The pure variant does not auto-inject the Stripe.js `<script>` tag on import — it's inserted only when a purchase begins.
- Motivation: when a downstream library bundles `purchases-js` with Rollup's `output.inlineDynamicImports: true` (required by some hybrid SDK distributions), our internal dynamic `import("@stripe/stripe-js")` is flattened to a static one, which causes Stripe.js to load at SDK initialization. The `/pure` entrypoint sidesteps that.
- Implemented via a build-time alias gated by `BUILD_VARIANT=pure`; no source changes in `src/`. Types are identical and shared between both entrypoints.
- Adds a `/pure` test page and "Try /pure entrypoint" button to the example app for manual verification.

## Usage

```ts
import { Purchases } from "@revenuecat/purchases-js/pure";
```

Same public API as the default entrypoint. Recommended only for consumers affected by `inlineDynamicImports`-style bundling.

## Test plan

- [x] `pnpm build` produces both `dist/Purchases.{es,umd}.js` and `dist/Purchases-pure.{es,umd}.js`
- [x] Default Stripe chunk contains the auto-load init (`Promise.resolve().then(E())`); pure chunk does not, and contains the pure-only `setLoadParameters` API
- [x] `pnpm typecheck`, `pnpm test` (591 tests), `prettier --check .`, `eslint src/` all pass
- [x] Example app builds and includes both Stripe loader chunks
- [ ] Manual smoke test in example app: `cd examples/webbilling-demo && pnpm dev`, visit `/pure/<id>`, confirm `js.stripe.com` is not requested until clicking a package
- [ ] Manual smoke test in a downstream Rollup project with `output.inlineDynamicImports: true`: import from `/pure`, confirm no Stripe.js request on page load

### DIFF
--- a/CHANGELOG.latest.md
+++ b/CHANGELOG.latest.md
@@ -1,3 +1,7 @@
+## RevenueCat SDK
+### ✨ New Features
+* Add `@revenuecat/purchases-js/pure` entrypoint that resolves `@stripe/stripe-js` to its `/pure` variant, so Stripe.js is not auto-loaded at SDK initialization. Intended for consumers bundling with Rollup's `output.inlineDynamicImports: true`.
+
 ## RevenueCatUI SDK
 ### 🐞 Bugfixes
 * Paywalls | Update purchases-ui-js to support visibility from custom variables w/tests (#870) via Rosie Watson (@RosieWatson)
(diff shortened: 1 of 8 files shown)


tonidero:
Closing this in favor of https://github.com/RevenueCat/purchases-js/pull/882

PR #892, opened 2026-05-29
WFL-217 | Add support for workflows

RosieWatson:
## Motivation / Description

Add support for workflow endpoints to be used

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -77,6 +77,7 @@ import {
   workflowDataToNavData,
   type WorkflowData,
   type UIConfig,
+  mergeCustomVariables,
 } from "@revenuecat/purchases-ui-js";
 import { PaywallDefaultContainerZIndex } from "./ui/theme/constants";
 import {
@@ -1033,8 +1034,19 @@ export class Purchases {
               closePaywall();
             },
             onCompleteWorkflowNavigate,
+            onNavigateToUrlClicked: navigateToUrl,
+            onRestorePurchasesClicked,
+            onVisitCustomerCenterClicked,
             onComponentInteraction,
-            customVariables: paywallParams.customVariables,
+            globalVariables: paywallParams.customVariables
+              ? mergeCustomVariables(
+                  paywallParams.customVariables,
+                  workflowDataResponse.ui_config as unknown as UIConfig,
+                )
+              : undefined,
+            maxContentWidth: workflowDataResponse.content_max_width
+              ? String(workflowDataResponse.content_max_width)
+              : undefined,
           },
         });
 
--- a/src/main.ts
+++ b/src/main.ts
@@ -70,7 +70,14 @@ import type {
   ComponentInteractionData as UIComponentInteractionData,
   WalletButtonRender,
 } from "@revenuecat/purchases-ui-js";
-import { Paywall, type PaywallData } from "@revenuecat/purchases-ui-js";
+import {
+  Paywall,
+  type PaywallData,
+  Workflow,
+  workflowDataToNavData,
+  type WorkflowData,
+  type UIConfig,
+} from "@revenuecat/purchases-ui-js";
 import { PaywallDefaultContainerZIndex } from "./ui/theme/constants";
 import {
   buildVariablesPerPackage,
@@ -556,11 +563,20 @@ export class Purchases {
     if (!offering) {
       throw new Error("No offering found.");
     }
-    if (!offering.paywallComponents) {
+
+    // Check for a workflow associated with this offering.
+    const workflowsResponse = await this.backend
+      .getWorkflows(this._appUserId)
+      .catch(() => null);
+    const matchedWorkflowSummary = workflowsResponse?.workflows.find(
+      (w) => w.offering_id === offering.identifier,
+    );
+
+    if (!matchedWorkflowSummary && !offering.paywallComponents) {
       throw new Error("This offering doesn't have a paywall attached.");
     }
 
-    if (!offering.uiConfig) {
+    if (!matchedWorkflowSummary && !offering.uiConfig) {
       throw new Error(
         "No ui_config found for this offering, please contact support!",
       );
@@ -606,15 +622,16 @@ export class Purchases {
       ? paywallParams.selectedLocale
       : navigator.language;
 
-    const finalLocale = calculateLocale(
-      offering.paywallComponents,
-      selectedLocale,
-    );
+    // finalLocale and translator are only needed for the paywall path — the
+    // Workflow component handles locale resolution internally.
+    const finalLocale = offering.paywallComponents
+      ? calculateLocale(offering.paywallComponents, selectedLocale)
+      : selectedLocale;
 
     const translator = new Translator(
       {},
       finalLocale,
-      offering.paywallComponents.default_locale,
+      offering.paywallComponents?.default_locale ?? englishLocale,
     );
 
     const paywallSessionId = generateUUID();
@@ -722,7 +739,7 @@ export class Purchases {
         onDiscountCodeChanged: paywallParams.onDiscountCodeChanged,
         selectedLocale: finalLocale,
         defaultLocale:
-          offering.paywallComponents?.default_locale || englishLocale,
+          offering.paywallComponents?.default_locale ?? englishLocale,
         paywallId: offering.paywallComponents?.id,
       });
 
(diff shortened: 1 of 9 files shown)


src/networking/endpoints.ts
```diff
@@ -200,6 +200,84 @@ export class GetVirtualCurrenciesEndpoint implements Endpoint {
...
+  method: HttpMethodType = "GET";
+  name: string = "getWorkflows";
+  private readonly appUserId: string;
+
+  constructor(appUserId: string) {
+    this.appUserId = appUserId;
+  }
+
+  urlPath(): string {
+    return `${SUBSCRIBERS_PATH}/${encodeURIComponent(this.appUserId)}/workflows`;
```
vegaro: are we missing the type parameter here?

RosieWatson: yeeah, good spot

src/tests/test-responses.ts
```diff
@@ -1044,5 +1044,11 @@ export function getRequestHandlers(): RequestHandler[] {
     }),
   );
 
+  requestHandlers.push(
+    http.get("http://localhost:8000/v1/subscribers/:userId/workflows", () =>
```
vegaro: I am surprised this works without the type parameter?

src/networking/endpoints.ts
```diff
@@ -200,13 +200,100 @@ export class GetVirtualCurrenciesEndpoint implements Endpoint {
...
+
+/**
+ * Step 2: fetch the full workflow payload from the subscriber endpoint.
+ * The path is taken verbatim from the workflow_url returned by step 1
+ * (after substituting {app_user_id}), and is authenticated with the
+ * api_key returned by step 1.
+ */
+export class GetWorkflowDataEndpoint implements Endpoint {
+  method: HttpMethodType = "GET";
+  name: string = "getWorkflowData";
```
vegaro: is it problematic that `getWorkflowData` is also used in `GetWorkflowDataByIdEndpoint`?

src/networking/responses/workflow-response.ts
```diff
@@ -0,0 +1,94 @@
+/**
+ * Response from GET /workflows/v1/workflow/{workflowLinkId}.
+ * Returns the public API key and a workflow_url template for the subscriber
+ * fetch (step 2). The workflow_url contains `{app_user_id}` as a placeholder.
+ */
+export interface WorkflowMetadataResponse {
+  api_key: string;
+  workflow_url: string;
+  workflow_id: string;
+  allow_anonymous_purchases: boolean | undefined;
```
vegaro: oh first time I see this property, interesting

src/networking/endpoints.ts
```diff
@@ -200,13 +200,100 @@ export class GetVirtualCurrenciesEndpoint implements Endpoint {
...
   | GetOfferingsEndpoint
   | PurchaseEndpoint
   | GetProductsEndpoint
   | GetCustomerInfoEndpoint
+  | IdentifyEndpoint
   | GetBrandingInfoEndpoint
+  | CheckoutPrepareEndpoint
+  | CheckoutStartEndpoint
+  | CheckoutRefreshPricingEndpoint
+  | CheckoutCompleteEndpoint
```
vegaro: these are unrelated to the PR, just want to make sure it's intended

src/main.ts
```diff
@@ -841,6 +858,197 @@ export class Purchases {
...
+            onExitBack: () => {
+              if (paywallParams.onBack) {
+                paywallParams.onBack(closePaywall);
+                return;
+              }
+              closePaywall();
+            },
+            onCompleteWorkflowNavigate,
+            onComponentInteraction,
+            customVariables: paywallParams.customVariables,
```
vegaro: I was checking https://github.com/RevenueCat/purchases-ui-js/blob/main/src/lib/components/workflows/Workflow.svelte

and it looks like `customVariables` is not a prop?

vegaro: `maxContentWidth` is one that we are not passing here and I believe it's part of the response.

vegaro: Another question... it looks like `Workflow.svelte` doesn't have `onRestorePurchasesClicked`, `onVisitCustomerCenterClicked`, `onNavigateToUrlClicked`, etc

Is that something broken, coming in a future PR? If I understand correctly it means that clicks in links like "Terms" or in restore buttons won't do anything?

vegaro (commented):
some comments which I am not 100% sure about, so holding approval until that's answered

src/tests/test-responses.ts
RosieWatson: msw isn't super bothered about the params, so it does a kind of catch all, but this should be updated

src/networking/endpoints.ts
RosieWatson: I think it's only used in error logs, so would make debugging harder, will rename

src/networking/endpoints.ts
RosieWatson: Yeah good spot, they probably should be there, but will ping the funnels guys to see if they want them added and will do separately

src/networking/responses/workflow-response.ts
RosieWatson: I think this snuck in when I was looking at what it would take to move the logic from rc-workflows into purchases-js so that we didn't need to duplicate it too much, but ended up being a much larger project that I need to scope out more, will clean this up!

src/main.ts
RosieWatson: Making an update to purchases-ui-js to fix part of this

src/networking/responses/workflow-response.ts
```diff
@@ -0,0 +1,93 @@
...
+ */
+export interface WorkflowStep {
+  id: string;
+  screen_id: string;
+  type: string;
+  param_values: Record<string, unknown>;
+  trigger_actions: Record<string, unknown>;
+  triggers: Record<string, unknown>;
+  outputs: Record<string, unknown>;
+  metadata: Record<string, unknown> | null;
```
jholwill, line 23: Would there be any value in typing these more concretely?

Though, looking again, we already defined some of these interfaces in the `ui-js` package (with the exact same shape - i should have noticed the unknowns previously), we should be able to re-use all these types from there as we import the package? Could avoid duplication / maintenance burden?

src/networking/backend.ts
```diff
@@ -480,4 +491,121 @@ export class Backend {
...
   }
+
+  /**
+   * Fetch the list of published workflows for an app user.
+   * Each entry includes an optional `offering_id` that can be matched against
+   * an offering's identifier to determine whether to show a workflow instead
+   * of a standalone paywall.
+   *
+   * GET /v1/subscribers/{appUserId}/workflows
+   */
```
jholwill, line 502: nit: none of the other functions in this file include docstrings. Do you think they are adding a lot of extra context worth keeping?

src/main.ts
```diff
@@ -841,6 +859,208 @@ export class Purchases {
...
+        }
+        if (paywallParams.onPurchaseError) {
+          paywallParams.onPurchaseError(err);
+        }
+      }
+    };
+
+    // ── Workflow path ────────────────────────────────────────────────────────
+    // If a workflow is associated with this offering, fetch its full data and
+    // mount the Workflow component instead of Paywall.
```
jholwill, line 902: Instead of a comment to highlight this, it might be worth pulling out both the paths into functions which could be clearer?

Comments like this are easy to miss / get separated from the code they relate to / hard to miss where the code block ends.

Was it going to require too many params to be passed in pulling it out of inline?

jholwill: In fact, these two paths are almost identical - other than the workflowDataResponse and some formatting - the only difference really lies in the returned promise?

Maybe could move the if statement further down and might not need the code comment or to pull it out into a function

src/networking/endpoints.ts
```diff
@@ -200,6 +200,84 @@ export class GetVirtualCurrenciesEndpoint implements Endpoint {
   }
 }
 
+/**
+ * GET /v1/subscribers/{appUserId}/workflows — returns the list of published
+ * workflows for an app, each with an optional offering_id for matching.
+ */
```
jholwill, line 206: nit: this file also doesn't include any other docstrings.

The step 1 / 2 below i think are more suited to where they are used rather than the edpoint definition?

jholwill (changes requested):

jholwill (changes requested):
Think we need to just de-duplicate some of the code in `src/main.ts`

RosieWatson:
Comments addressed and replaced with [this PR](https://github.com/RevenueCat/purchases-js/pull/913) as @jholwill is on holiday for 2 weeks and I don't have permissions to remove the blocking review

PR #894, opened 2026-06-02
Show correctly discounted price in footer

lburdock:
## Motivation / Description

When a discount code is applied to a subscription with a free trial, the footer terms still show the undiscounted base price. This PR fixes that by using the updated purchase_options provided by the operation session endpoint.

## Changes Introduced

* Uses the correct discounted price in the footer.
* Updates the example app to allow us to enable `showDiscountCodeField` via a query param.

## Before
Screenshot 2026-06-02 at 11 35 28 AM

#### Time window discount

Screenshot 2026-06-10 at 12 15 40 PM
Screenshot 2026-06-10 at 12 15 22 PM

#### One time discount
Screenshot 2026-06-10 at 12 15 08 PM
Screenshot 2026-06-10 at 12 14 55 PM

#### Forever discount
Screenshot 2026-06-10 at 12 14 43 PM
Screenshot 2026-06-10 at 12 14 28 PM

## Additional comments
Completes WEB-4344

---

> [!NOTE]
> **Medium Risk**
> Changes how purchase options and pricing copy are sourced during checkout and discount flows; incorrect fallback could misstate renewal prices or trial terms, though behavior is covered by new unit/UI tests.
> 
> **Overview**
> Fixes checkout footer and related copy showing **undiscounted** subscription terms (e.g. post-trial price) after a promo code is applied.
> 
> Checkout pricing responses now include **`selected_purchase_option`**. The UI resolves the **active purchase option** via `getActiveCheckoutPurchaseOption` (session option when present, otherwise the `/products` option), and **`purchases-ui`** keeps pricing, Stripe elements config, and applied discount code in sync with the latest session refresh instead of stale product catalog data.
> 
> **`payment-entry-page`** derives subscription phases from the passed `purchaseOption` (not `productDetails.subscriptionOptions`) and can bubble session pricing updates to the parent so trial/intro/discount labels update when codes are applied or removed.
> 
> The web billing demo exposes **`showDiscountCodeField`** through a query param on paywall routes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ca10ed693a7a96083fedfae791d70e20f8e070c6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/examples/webbilling-demo/src/pages/paywall/index.tsx
+++ b/examples/webbilling-demo/src/pages/paywall/index.tsx
@@ -184,6 +184,8 @@ const PaywallPage: React.FC = () => {
   const displayName = searchParams.get("$displayName");
   const nickname = searchParams.get("nickname");
   const skipSuccessPage = searchParams.get("skipSuccessPage") === "true";
+  const showDiscountCodeField =
+    searchParams.get("showDiscountCodeField") === "true";
   const attributesSetRef = useRef(false);
 
   useEffect(() => {
@@ -247,6 +249,7 @@ const PaywallPage: React.FC = () => {
         await purchases.purchase({
           rcPackage: pkg,
           purchaseOption: option,
+          showDiscountCodeField,
           selectedLocale: lang || navigator.language,
           customerEmail: email || undefined,
           skipSuccessPage: skipSuccessPage,
--- a/examples/webbilling-demo/src/pages/rc_paywall/index.tsx
+++ b/examples/webbilling-demo/src/pages/rc_paywall/index.tsx
@@ -13,6 +13,8 @@ const RCPaywallPage: React.FC = () => {
   const lang = searchParams.get("lang");
   const email = searchParams.get("email");
   const hideBackButtons = searchParams.get("hideBackButtons") === "true";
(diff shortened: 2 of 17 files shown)


james-od (commented):
I'm not sure I follow this - discounts aren't intended to stack with free trials or intro offers, the behaviour should be that if a discount is applied then the trial/intro offer is ignored?
So I think the bug is actually making sure we enforce that? 
If I start on a checkout here, with a product that has a free trial and also has a "forver" discount applicable:
Image
^I can see the free trial is being applied, the final price is correct and the footer text is correct.

If I then type in the discount code I see this:

Image

^ The trial should not have stacked with the discount. This is why we see the incorrect footer text.

If I refresh (i.e get the code from the URL instead of the input) I see this

Image

^ This is correct, the discount is being applied instead of the trial

If I then remove the discount code we get this:
Image
^ This is probably not correct, we'd expect the free trial to apply.

Unless I've missed something, wdyt?

james-od (approved):
Nice, code looks good and I ran a few manual scenarios - all seems to work well!

PR #896, opened 2026-06-03
feat(paddle): inline checkout support at the service layer (WST-567 PR 1)

rogersole:
## Summary

First PR in the stack to switch RevenueCat's Paddle integration from the **overlay** (hosted modal popup) checkout to an **inline** checkout (Paddle's checkout embedded inside a container we control). Tracked under **WST-567**.

This PR lays the **service-layer foundation** and **bumps the Paddle SDK**. It is intentionally a no-op at runtime — overlay remains the default and no caller opts into inline yet.

## What changed

- **`src/paddle/paddle-service.ts`**
  - Extracted a pure, exported `buildPaddleCheckoutOptions({ transactionId, locale, customerEmail, displayMode })` helper that builds the options for `Paddle.Checkout.open()`.
  - Added a `PaddleCheckoutDisplayMode = "overlay" | "inline"` type and threaded an optional `displayMode` (default `"overlay"`) through `PaddleService.purchase()`.
  - When `displayMode: "inline"`, the helper additionally sets `frameTarget` / `frameInitialHeight` / `frameStyle` (exported `PADDLE_INLINE_FRAME_TARGET` for the UI to reuse).
- **`@paddle/paddle-js` `1.5.1 → 1.6.4`** — no breaking changes; `1.6.2+` adds the Upsell / Saved-Payment-Method / `updateCheckout` APIs needed by later work.
- **Tests** — added coverage for overlay (default) vs. inline option building, email inclusion/omission, locale passthrough, and that `purchase()` opens overlay by default and inline when requested.

## Why this is safe to merge on its own

`displayMode` defaults to `"overlay"` and **no caller passes anything else yet**, so the built options are equivalent to the previous inline object literal. The inline capability is dormant until the follow-up UI PR wires it up.

## Test plan

- [x] `tsc --noEmit` clean
- [x] `vitest run` full suite green (paddle-service: 35 passed)
- [x] `prettier` + `eslint` clean on changed files

## Stack

1. **PR 1 (this)** — service-layer inline support + SDK bump
2. PR 2 — inline UI page, gated off by default
3. PR 3 — inline state machine, lifecycle & `closed` semantics
4. PR 4 — theming, localization & responsive frame
5. PR 5 — cutover (flip default to inline)

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     "@babel/runtime": "^7.26.10",
     "@eslint/js": "^9.16.0",
     "@microsoft/api-extractor": "^7.48.0",
-    "@paddle/paddle-js": "^1.5.1",
+    "@paddle/paddle-js": "^1.6.4",
     "@revenuecat/purchases-ui-js": "^4.7.2",
     "@storybook/addon-essentials": "^8.6.15",
     "@storybook/addon-interactions": "^8.6.15",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,73 +1,74 @@
-lockfileVersion: "9.0"
+lockfileVersion: '9.0'
 
 settings:
   autoInstallPeers: true
   excludeLinksFromLockfile: false
 
 importers:
+
   .:
     devDependencies:
-      "@babel/runtime":
+      '@babel/runtime':
         specifier: ^7.26.10
         version: 7.28.6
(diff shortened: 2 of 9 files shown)


dalshamaa (approved):
This looks good to me but I'm not the most familiar with UI stuff

vkartaviy (approved):

PR #897, opened 2026-06-03
feat(paddle): inline checkout UI page, gated off by default (WST-567 PR 2)

rogersole:
## Summary

Second PR in the stack switching Paddle from overlay to **inline** checkout (**WST-567**). Builds on the service-layer foundation from #896.

Introduces the inline checkout **UI page and an internal feature flag**, both **gated off by default** — production still uses the overlay. No public API change.

> ⚠️ Stacked on **#896** — review/merge that first. This PR's diff is against `paddle-inline/01-service-inline-support`.

## What changed

- **New `src/ui/paddle-inline-checkout-page.svelte`** — renders RevenueCat's branded `FullscreenTemplate` containing the `.paddle-checkout-container` div that Paddle injects its inline iframe into (uses the exported `PADDLE_INLINE_FRAME_TARGET` from #896 as the single source of truth for the class name).
- **`src/ui/paddle-purchases-ui.svelte`** — adds an internal `useInlineCheckout` prop (default `false`). When `true`:
  - renders `PaddleInlineCheckoutPage` (so the container exists in the DOM before `purchase()` runs), and
  - passes `displayMode: "inline"` through to `PaddleService.purchase()`.
  When `false`, the overlay path is **byte-for-byte unchanged** — no `displayMode` is passed and no container is rendered.

## Why this is safe to merge on its own

`useInlineCheckout` defaults to `false` and is **not wired to any public configure() option yet**, so nothing reaches the new code path in production. The flag-off branch renders exactly as before.

## Out of scope (next PR)

The inline **state machine**, `checkout.closed` semantics (no popup to dismiss), iframe mount/unmount lifecycle, and simplifying the full-page body-style overrides land in PR 3.

## Test plan

- [x] `tsc --noEmit` clean
- [x] `svelte-check` 0 errors
- [x] `vitest run` full suite green; Paddle UI suite 19 passed (16 existing unchanged + 3 new: container renders when enabled, `displayMode: "inline"` passed when enabled, neither happens by default)

---

**Update:** also adds the **cancel affordance** for inline checkout. Inline has no Paddle-provided dismiss (unlike the overlay modal), so `PaddleInlineCheckoutPage` now renders an in-page close button (wired to `onClose`, shown when not embedded in a host element), and the Paddle flow registers a `popstate → onClose` listener so browser-back also cancels — matching the Web Billing / Stripe flows. Covered by tests (close button renders + invokes `onClose`; hidden when `isInElement`).

---

**Update 2 (review feedback):** the inline page now renders through the branded `FullscreenTemplate` with a `BrandingHeader` (back/close affordance) and the `ProductInfo` panel — restoring the product/offering details and replacing the ad-hoc floating close button. Cancelling the inline checkout calls **`Paddle.Checkout.close()`** (`PaddleService.closeCheckout()`) to tear down the embedded iframe before unmounting, per Paddle's branded inline checkout guidance. (`FullscreenTemplate` is used rather than the two-column `Template` because the latter left an orphaned container node when swapping to the processing/status pages.)

---

**Update 3 — inline order summary (new component, by necessity):** the inline checkout now renders a branded two-column order summary (amount card + Subtotal/Tax/Total card), fed live by Paddle's own checkout totals (`onCheckoutTotals` ← `checkout.loaded` / `checkout.updated`).

Why this is **newly built**: with the **overlay** checkout, Paddle rendered this order summary itself — it's part of Paddle's hosted overlay UI, so we never had to build it. The **inline** checkout only renders the payment form, so the summary has to be reconstructed on our side from the totals Paddle exposes through its checkout events.

All styling comes from the server-provided branding **appearance config** (nothing hardcoded): card background = `color_form_bg`, corner radius = `shapes`, amount color = `color_buttons_primary`, panel background = `color_product_info_bg`. New component: `src/ui/organisms/paddle-order-summary.svelte` (reuses the shared `ProductHeader` + `PricingTable`).

Note: the tax row label is generic ("Tax") since Paddle's event totals expose the tax amount but not a region-specific label (e.g. "VAT"); and the Paddle seller name still isn't exposed to the SDK, so the back link uses the branding app name.

---

**Update 4 — order summary restyled to match Paddle's overlay.** `PaddleOrderSummary` now mirrors the hosted overlay's summary: amount + "inc. VAT", "billed <frequency>", product name, price name, amount/period; and a breakdown card with Subtotal / VAT / Total + a recurring "Due on <date>" row. Generous padding, rounded cards. The extra data (recurring total, product name, price name) is forwarded from Paddle's checkout events via `onCheckoutTotals`. Card colors come from the appearance config; the rounded radius is fixed to match the overlay. Next-billing date is computed client-side from the billing period (the exact Paddle renewal date isn't exposed via events).

### DIFF
--- a/src/ui/organisms/paddle-order-summary.svelte
+++ b/src/ui/organisms/paddle-order-summary.svelte
@@ -1,113 +1,157 @@
 <script lang="ts">
   import { getContext } from "svelte";
   import { type Writable } from "svelte/store";
-  import ProductHeader from "../molecules/product-header.svelte";
-  import PricingTable from "../molecules/pricing-table.svelte";
   import Typography from "../atoms/typography.svelte";
   import { Theme } from "../theme/theme";
   import { translatorContextKey } from "../localization/constants";
   import { Translator } from "../localization/translator";
+  import { getInitialPriceFromPurchaseOption } from "../../helpers/purchase-option-price-helper";
   import type { BrandingInfoResponse } from "../../networking/responses/branding-response";
   import {
     type Product,
     type PurchaseOption,
     type SubscriptionOption,
-    type NonSubscriptionOption,
-    type PricingPhase,
   } from "../../entities/offerings";
-  import { type PriceBreakdown } from "../ui-types";
+  import { type Period } from "../../helpers/duration-helper";
+  import type { PaddleCheckoutTotals } from "../../paddle/paddle-service";
 
   interface Props {
     brandingInfo: BrandingInfoResponse | null;
     productDetails: Product;
     purchaseOption: PurchaseOption;
-    priceBreakdown: PriceBreakdown;
+    // Live order totals from Paddle's checkout events (null until the inline
+    // checkout reports them); drives the breakdown + recurring amount.
+    totals: PaddleCheckoutTotals | null;
   }
 
-  const {
-    brandingInfo,
-    productDetails,
-    purchaseOption,
-    priceBreakdown,
-  }: Props = $props();
+  const { brandingInfo, productDetails, purchaseOption, totals }: Props =
+    $props();
 
   const translator: Writable<Translator> = getContext(translatorContextKey);
 
-  // Card chrome comes from the same branding/appearance config the rest of the
-  // checkout uses: the form background for the cards, the shape preset for the
-  // corner radius. The amount uses the primary (buttons) color.
+  // Card chrome from the branding/appearance config: form background for the
+  // cards, primary (buttons) color for the amount. Corner radius is generous to
+  // match the hosted overlay summary.
   const theme = new Theme(brandingInfo?.appearance ?? null);
   const cardBackground = theme.formColors.background;
-  const cardRadius = theme.shape["input-border-radius"];
 
   const isSubscription = productDetails.productType === "subscription";
-  const subscriptionOption = isSubscription
-    ? (purchaseOption as SubscriptionOption)
+  const basePeriod: Period | null = isSubscription
+    ? ((purchaseOption as SubscriptionOption)?.base?.period ?? null)
     : null;
-  const nonSubscriptionOption = !isSubscription
-    ? (purchaseOption as NonSubscriptionOption)
-    : null;
-  const basePhase: PricingPhase | null = isSubscription
-    ? (subscriptionOption?.base ?? null)
-    : nonSubscriptionOption?.basePrice
-      ? {
-          periodDuration: null,
-          period: null,
-          cycleCount: 1,
-          price: nonSubscriptionOption.basePrice,
-          pricePerWeek: null,
-          pricePerMonth: null,
-          pricePerYear: null,
-        }
-      : null;
-  const trialPhase = subscriptionOption?.trial ?? null;
+
+  const toMicros = (amount: number): number => Math.round(amount * 1_000_000);
+
+  const fallbackPrice = getInitialPriceFromPurchaseOption(
+    productDetails,
+    purchaseOption,
+  );
+  const currency = $derived(totals?.currencyCode ?? fallbackPrice.currency);
+  const totalMicros = $derived(
+    totals ? toMicros(totals.totalAmount) : fallbackPrice.amountMicros,
+  );
+
+  const formatAmount = (micros: number): string =>
+    $translator.formatPrice(micros, currency);
+
+  const periodUnitLabel = $derived(
+    basePeriod ? $translator.translatePeriodUnit(basePeriod.unit) : null,
+  );
+  const billedFrequencyLabel = $derived(
+    basePeriod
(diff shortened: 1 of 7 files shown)


src/ui/organisms/paddle-order-summary.svelte
```diff
@@ -0,0 +1,231 @@
...
+
+  // Card chrome from the branding/appearance config: form background for the
+  // cards, primary (buttons) color for the amount. Corner radius is generous to
+  // match the hosted overlay summary.
+  const theme = new Theme(brandingInfo?.appearance ?? null);
+  const cardBackground = theme.formColors.background;
+
+  const isSubscription = productDetails.productType === "subscription";
+  const basePeriod: Period | null = isSubscription
+    ? ((purchaseOption as SubscriptionOption)?.base?.period ?? null)
```
vkartaviy: `as SubscriptionOption` asserts the type without checking it. Can we narrow it properly instead?

src/ui/organisms/paddle-order-summary.svelte
```diff
@@ -0,0 +1,231 @@
...
+      case "month":
+        date.setMonth(date.getMonth() + n);
+        break;
+      case "year":
+        date.setFullYear(date.getFullYear() + n);
+        break;
+      default:
+        return null;
+    }
+    return new Intl.DateTimeFormat(undefined, {
```
vkartaviy: `undefined` locale means this formats the date in the browser's locale, while prices above use the translator's locale. Should this use the configured locale too?

src/ui/organisms/paddle-order-summary.svelte
```diff
@@ -0,0 +1,231 @@
...
+        break;
+      case "month":
+        date.setMonth(date.getMonth() + n);
+        break;
+      case "year":
+        date.setFullYear(date.getFullYear() + n);
+        break;
+      default:
+        return null;
+    }
```
vkartaviy: This duplicates `getNextRenewalDate` in `duration-helper.ts`. The existing one also handles month-overflow (Jan 31 + 1mo) and leap-year Feb 29, which this version doesn't - so this is a less-correct copy. Can we call the helper and just keep the `Intl` formatting here?

src/ui/organisms/paddle-order-summary.svelte
```diff
@@ -0,0 +1,231 @@
...
+        {/if}
+      </div>
+    </div>
+  </div>
+
+  <!-- Totals breakdown card -->
+  {#if totals}
+    <div class="rcb-paddle-summary-card">
+      <div class="rcb-paddle-summary-row">
+        <span class="rcb-paddle-summary-muted">Subtotal</span>
```
vkartaviy: These look hardcoded - `pricing_table.subtotal` (and the sibling keys) already exist in the locale files. Worth running them through the translator like the period labels?

vkartaviy (changes requested):

src/ui/organisms/paddle-order-summary.svelte
rogersole: Done in ee09173 — replaced the cast with an \`isSubscriptionOption(option): option is SubscriptionOption\` type guard (\`"base" in option\`).

src/ui/organisms/paddle-order-summary.svelte
rogersole: Done in ee09173 — now formats with \`toBcp47Locale($translator.selectedLocale)\` so the date matches the locale used for prices.

src/ui/organisms/paddle-order-summary.svelte
rogersole: Done in ee09173 — now reuses \`getNextRenewalDate(new Date(), basePeriod, true)\` and only keeps the \`Intl.DateTimeFormat\` call, so the leap-year / month-overflow handling lives in one place.

src/ui/organisms/paddle-order-summary.svelte
rogersole: Done in ee09173 — the breakdown labels now go through the translator: \`PricingTableSubtotal\` / \`PricingTableTax\` / \`PricingTableTotalDueToday\`. Note this changes the wording from "VAT"/"Total" to the existing localized "Tax"/"Total due today"; also lowercased the card-1 "inc. VAT" → "inc. tax" for consistency. (The Paddle-specific "billed {x}" and "Due on {date}" strings have no existing keys yet — left for a follow-up if we want them localized.)

vkartaviy (approved):

src/ui/organisms/paddle-order-summary.svelte
rogersole: Created a follow-up ticket to handle the translations: https://linear.app/revenuecat/issue/WST-692/paddles-inline-checkout-follow-up-translations

PR #899, opened 2026-06-03
feat(paddle): map branding to inline checkout theme (WST-567 PR 4)

rogersole:
## Summary

Fourth PR in the stack switching Paddle from overlay to **inline** checkout (**WST-567**). Builds on #898.

Maps the merchant's RevenueCat branding onto Paddle's inline checkout so the embedded checkout visually matches the surrounding branded UI.

> ⚠️ Stacked on **#898 → #897 → #896**. Merge in order. This PR's diff is against `paddle-inline/03-inline-state-machine`.

## What changed

- **`src/ui/theme/utils.ts`** — exported `isHexColorLight(hex)`, reusing the existing private `hexToRGB` + WCAG luminance logic and `DEFAULT_LUMINANCE_THRESHOLD`.
- **`src/paddle/paddle-service.ts`** — added a `theme` option (`"light" | "dark"`, default `"light"`) to `buildPaddleCheckoutOptions()` / `purchase()`.
- **`src/ui/paddle-purchases-ui.svelte`** — derives the Paddle theme from the branding page background (`color_page_bg`, default `#ffffff` → light) and passes it through when inline.

## Scope rationale

Paddle's inline checkout only exposes **`theme`** and **`frameStyle`** in code — deeper colors/typography are configured in the Paddle dashboard, not via Paddle.js. So:

- **theme** → mapped from branding (this PR).
- **frameStyle** → kept `transparent` so the iframe inherits our page background; no per-brand value needed.
- **locale** → already forwarded via `selectedLocale` → `purchase()`.
- **success / error parity** → already rendered by the shared `PaddlePurchasesUiInner` (PR 2/3).
- **responsive height** → Paddle auto-resizes the iframe; `frameInitialHeight` is just the starting value.

Note: the existing `hexToRGB` doesn't expand 3-digit shorthand hex; real branding colors are 6-digit, so `isHexColorLight` is exercised/tested with 6-digit values.

## Safety

All inline behavior stays behind `useInlineCheckout` (default `false`); overlay is unchanged (still `theme: "light"`, no theme passed from the UI).

## Test plan

- [x] `tsc --noEmit` clean
- [x] `svelte-check` 0 errors
- [x] `vitest run` full suite green (+ tests: `isHexColorLight` light/dark/fallback; `buildPaddleCheckoutOptions` theme default + override; inline UI passes light/dark theme based on page background)

### DIFF
--- a/src/paddle/paddle-service.ts
+++ b/src/paddle/paddle-service.ts
@@ -3,7 +3,6 @@ import type {
   DisplayMode,
   Paddle,
   PaddleEventData,
-  Theme,
   Variant,
   Version,
 } from "@paddle/paddle-js";
@@ -105,7 +104,7 @@ export function buildPaddleCheckoutOptions({
   theme = "light",
 }: BuildPaddleCheckoutOptionsParams): CheckoutOpenOptions {
   const commonSettings = {
-    theme: theme as Theme,
+    theme,
     variant: "one-page" as Variant,
     locale,
     allowLogout: false,
--- a/src/paddle/paddle-service.ts
+++ b/src/paddle/paddle-service.ts
@@ -50,6 +50,9 @@ interface PaddlePurchaseParams {
  */
 export type PaddleCheckoutDisplayMode = "overlay" | "inline";
 
+/** Paddle checkout color theme. Mapped from the merchant's branding. */
+export type PaddleCheckoutTheme = "light" | "dark";
+
 /**
  * Order totals surfaced by Paddle's checkout events (`checkout.loaded` /
  * `checkout.updated`). Amounts are in major currency units (e.g. 9.99). Lets
@@ -85,6 +88,7 @@ interface BuildPaddleCheckoutOptionsParams {
   locale: string;
   customerEmail?: string;
   displayMode?: PaddleCheckoutDisplayMode;
+  theme?: PaddleCheckoutTheme;
 }
 
 /**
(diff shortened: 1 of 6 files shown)


src/paddle/paddle-service.ts
```diff
@@ -98,9 +102,10 @@ export function buildPaddleCheckoutOptions({
   locale,
   customerEmail,
   displayMode = "overlay",
+  theme = "light",
 }: BuildPaddleCheckoutOptionsParams): CheckoutOpenOptions {
   const commonSettings = {
-    theme: "light" as Theme,
+    theme: theme as Theme,
```
vkartaviy: I think `cast` is not needed here.

src/ui/theme/utils.ts
```diff
@@ -74,6 +74,20 @@ const isLightColor = ({
 
 export const DEFAULT_LUMINANCE_THRESHOLD = 0.37;
 
+/**
+ * Returns whether a hex color (e.g. "#ffffff" or "#fff") reads as light.
```
vkartaviy: The docstring lists `#fff` as an example, but per the scope note this only handles 6-digit hex - `hexToRGB` doesn't expand shorthand.

vkartaviy (approved):
Left some nits

src/paddle/paddle-service.ts
rogersole: Done in c01f2b7 — dropped the cast (and the now-unused \`Theme\` import). \`PaddleCheckoutTheme\` is \`"light" | "dark"\`, identical to Paddle's \`Theme\`, so it assigns directly.

src/ui/theme/utils.ts
rogersole: Good catch — fixed in c01f2b7. Removed the "#fff" example and noted the docstring only covers 6-digit hex (branding colors are always 6-digit; \`hexToRGB\` doesn't expand shorthand).

PR #900, opened 2026-06-03
feat(paddle): make inline checkout the default — cutover (WST-567 PR 5)

rogersole:
## Summary

Final functional PR in the stack switching Paddle from overlay to **inline** checkout (**WST-567**) — the **production cutover**. Builds on #899.

Flips `PaddlePurchasesUi`'s `useInlineCheckout` default to `true`. Since `main.ts` mounts the component without passing the flag, this single default change makes all Paddle purchases render **inline** (embedded in our branded UI) instead of the legacy overlay popup.

> ⚠️ This PR **changes production behavior**. Stacked on **#899 → #898 → #897 → #896**; merge in order.

## What changed

- **`src/ui/paddle-purchases-ui.svelte`** — `useInlineCheckout` now defaults to `true`. Overlay stays reachable via `useInlineCheckout={false}` as a one-release fallback (dead-code removal is a follow-up).
- **`src/tests/ui/paddle-purchases-ui.test.ts`** — the default-path test now asserts inline; overlay-specific tests (loading spinner, "no displayMode") pass `useInlineCheckout={false}` explicitly, preserving overlay regression coverage.

## Live validation

Re-ran the full flow against the Paddle sandbox via the demo app with **no flag passed** (i.e. exercising the new default through the real PR 1–5 code path, not the spike):

- ✅ Paddle iframe mounts inside `.paddle-checkout-container`
- ✅ checkout URL: `display_mode=inline&variant=one-page`
- ✅ renders embedded in our branded page (not an overlay popup)

## Test plan

- [x] `tsc --noEmit` clean
- [x] `vitest run` full suite green; Paddle UI suite 22 passed
- [x] Manual: sandbox purchase renders inline end-to-end

## Follow-ups (not in this stack)

- Remove the overlay fallback code once inline has soaked.
- PR 6 — Paddle upsells (depends on backend support).

### DIFF
--- a/src/ui/paddle-purchases-ui.svelte
+++ b/src/ui/paddle-purchases-ui.svelte
@@ -49,10 +49,10 @@
     unmountPaddlePurchaseUi: () => void;
     paddleService: PaddleService;
     /**
-     * Internal flag (defaults to false). When true, Paddle's checkout is
-     * embedded inline in our own container instead of opening as an overlay.
-     * Not yet exposed through the public configure() surface — wiring and the
-     * inline state machine land in follow-up PRs.
+     * Controls how Paddle's checkout is presented. Defaults to `true`
+     * (inline — embedded in our own container). Set to `false` to fall back to
+     * the legacy overlay (modal popup) presentation; this fallback is kept for
+     * one release and is expected to be removed once inline has soaked.
      */
     useInlineCheckout?: boolean;
   }
@@ -77,7 +77,7 @@
     attributionMetadata,
     unmountPaddlePurchaseUi,
     paddleService,
-    useInlineCheckout = false,
+    useInlineCheckout = true,
   }: Props = $props();
 
   let translator: Translator = new Translator(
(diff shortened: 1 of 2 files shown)


src/ui/paddle-purchases-ui.svelte
```diff
@@ -77,7 +77,7 @@
     attributionMetadata,
     unmountPaddlePurchaseUi,
     paddleService,
-    useInlineCheckout = false,
+    useInlineCheckout = true,
```
vkartaviy, line 80: Overlay fallback isn't reachable in prod since nothing passes this prop, so rollback means a full SDK release - should inline be gated on a server flag for remote disable instead?

vkartaviy (approved):

src/ui/paddle-purchases-ui.svelte
rogersole: we should just release this version when we are sure we can, meaning after the bug bash. So in that scenario we will have the inline version working. The follow-up PR will be to clean up all the outline-related stuff

rogersole:
Closing in favor of this ticket: https://linear.app/revenuecat/issue/WST-700/inline-checkout-release-to-customers

PR #906, opened 2026-06-05
fix: use createEmbeddedCheckoutPage for Stripe embedded checkout

vkartaviy:
## What

The Web SDK now calls `stripe.createEmbeddedCheckoutPage()` for embedded checkout instead of the removed `stripe.initEmbeddedCheckout()`.

## Why

Stripe removed `stripe.initEmbeddedCheckout()` in its dahlia release - calling it throws `IntegrationError`. Stripe.js is a single `window.Stripe` per page and the SDK's loader reuses whatever `js.stripe.com` script a merchant has already loaded, so a merchant whose page runs dahlia got a broken checkout (silent on live keys - Stripe's version-mismatch warning fires only for test keys).

`createEmbeddedCheckoutPage()` is present and works on every current Stripe.js release train (acacia through dahlia), so calling it directly fixes checkout on all of them. The basil-era types do not declare the method yet, so the call needs one cast in `stripe-service.ts`.

## Cross-version E2E

Adds `examples/webbilling-demo/src/tests/stripe-checkout/cross-version.test.ts`: for each train (basil / clover / dahlia) it self-loads that Stripe.js, runs a real embedded checkout, and asserts the SDK ran against the merchant's train. Verified locally against a Stripe sandbox - all three complete. It is skipped on CI because the stripe E2E suite is disabled there (`VITE_SKIP_STRIPE_TESTS=true`), so it serves as a local/manual check.

A small guard (`config-guard.test.ts`) fails loudly if the stripe E2E key is missing while the suite is enabled, so a dropped secret cannot pass as green.

## Scope

SDK-only. The backend already returns the embedded-session `client_secret`, so no backend or schema change.

## Test plan

- `pnpm test`, `tsc --noEmit`, `pnpm svelte-check` clean
- Cross-version matrix passes locally (real sandbox checkout on basil, clover, dahlia)

---
Linear: https://linear.app/revenuecat/issue/WST-695/purchases-js-feature-detect-stripe-embedded-checkout-method
Part of https://linear.app/revenuecat/issue/WST-694/purchases-js-stripejs-decouple-checkout-from-the-pages-stripejs

### DIFF
--- a/examples/webbilling-demo/src/tests/stripe-checkout/config-guard.test.ts
+++ b/examples/webbilling-demo/src/tests/stripe-checkout/config-guard.test.ts
@@ -6,9 +6,9 @@ import { skipStripeTestsIfDisabled } from "../helpers/test-helpers";
 integrationTest.describe("Stripe Checkout E2E configuration", () => {
   skipStripeTestsIfDisabled(integrationTest);
 
-  // A missing key makes the whole Stripe Checkout suite (including the
-  // cross-version matrix) skip while CI still reports green - hiding the gap.
-  // Fail loudly on CI instead so a dropped secret cannot pass unnoticed.
+  // A missing key skips the whole Stripe Checkout suite (including the
+  // cross-version matrix) while CI still reports green, so fail loudly here
+  // instead and surface the dropped secret.
   integrationTest(
     "Stripe Checkout E2E key is set on CI so the suite cannot silently skip",
     () => {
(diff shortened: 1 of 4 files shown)


vkartaviy:
@codex please review

vkartaviy:
Closing in favor of #907 - it carries this feature-detect change plus the v9 (dahlia) bump, targeting main directly.

PR #911, opened 2026-06-09
feat(paddle): gate inline checkout behind per-project backend flag (WST-700)

rogersole:
## Context

Adds a **gate** to the Paddle inline checkout rollout so it can be controlled per project (advance warning, staged rollout, opt-in for managed accounts) **without customers shipping a code change**. Supersedes the previous unconditional cutover.

Stacked on `paddle-inline/04-theming-localization`.

## How it's gated

The backend ([khepri#21157](https://github.com/RevenueCat/khepri/pull/21157)) delivers the per-project flag on the `rc_billing` **checkout start** response, directly on `paddle_billing_params.inline_checkout_enabled` (alongside `client_side_token` / `is_sandbox`). It's set by a dashboard toggle persisted on `app_config_settings`, and defaults to `false` when unset. The Paddle UI derives its presentation mode from that value.

> Earlier revisions explored the branding response, then a `checkout_config.paddle_config` wrapper. The final backend shape is `paddle_billing_params.inline_checkout_enabled`; the net diff vs `04` reflects that.

## Changes

- **`checkout-start-response.ts`** — add `inline_checkout_enabled?: boolean` to `PaddleCheckoutStartResponse.paddle_billing_params`.
- **`paddle-purchases-ui.svelte`** — derive `useInlineCheckout` from `startResponse.paddle_billing_params.inline_checkout_enabled` after `startCheckout`; `await tick()` so the inline frame-target container is in the DOM before `purchase()` opens the checkout; drop the `useInlineCheckout` prop.
- **Tests** — component-level (`paddle-purchases-ui.test.ts`): inline is driven by the start-response mock; absent/false ⇒ legacy overlay (no regression).

## Verification

- `vitest run` on `main.test.ts` + `paddle-purchases-ui.test.ts`: **99 passed**.
- `pnpm build` (`tsc && vite build`): passes.
- Wire shape matches khepri#21157 (`paddle_billing_params.inline_checkout_enabled`, default `false`).

## Out of scope (track separately)

- **Backend** (khepri#21157): the dashboard toggle + populating the flag. Merges independently; manual end-to-end is covered here by the start-response mock until it ships.
- **Customer comms:** advance-warning email + managed-account opt-in outreach.
- Eventual removal of the legacy overlay once inline has soaked.


src/main.ts
```diff
@@ -1693,6 +1693,11 @@ export class Purchases {
             metadata,
             unmountPaddlePurchaseUi,
             paddleService,
+            // Gate the inline checkout behind the per-project backend flag
+            // (WST-700). Absent/false => legacy overlay, so existing projects
```
vkartaviy: Lets maybe remove references to internal ticket system?

rogersole: Good call — removed the WST-700 refs from the code comments (and the test describe label) in 5d94a65.

vkartaviy (approved):

src/main.ts
rogersole: Update: the gate moved off the branding response entirely — it now reads checkout_config.paddle_config.inline_checkout_enabled from the checkout start response (per backend design), so there are no ticket refs (or branding changes) left here. See f9ec8fe.

vkartaviy (approved):

src/ui/paddle-purchases-ui.svelte
```diff
@@ -197,6 +194,17 @@
...
+
+      // Gate the inline presentation on the per-project backend flag. Absent =>
+      // legacy overlay, so projects that haven't been opted in are unaffected.
+      useInlineCheckout =
+        startResponse.paddle_billing_params.inline_checkout_enabled ?? false;
+      // Paddle injects its iframe into the inline container (frameTarget), so
+      // that element must be in the DOM before purchase() opens the checkout.
+      // Flush the render now that we know the presentation mode.
+      if (useInlineCheckout) {
+        await tick();
```
nicfix, line 211: Maybe we can avoid this call using $derive as I mentioned above

src/ui/paddle-purchases-ui.svelte
```diff
@@ -104,6 +96,11 @@
   // processing state instead of leaving an empty container on screen.
   let checkoutCompleted = $state(false);
 
+  // How Paddle's checkout is presented. Defaults to the legacy overlay; set to
+  // inline only when the per-project backend flag (returned on the checkout
+  // start response) enables it. Resolved after startCheckout, below.
+  let useInlineCheckout = $state(false);
```
nicfix: I think we can use $derived for this instead of $state and tick() 
```suggestion
let useInlineCheckout = $state(startResponse.paddle_billing_params.inline_checkout_enabled ?? false)
```

rogersole: Done in 52d41c2 — `useInlineCheckout` is now `$derived` off a `$state startResponse` instead of an imperative reassignment.

src/ui/paddle-purchases-ui.svelte
rogersole: Switched to `$derived` (52d41c2), but I had to keep the `await tick()` — it's a DOM-flush, not a value-computation concern. Paddle mounts its checkout iframe into the inline `frameTarget` element, which must already be in the DOM when `purchase()` opens the checkout. Since the presentation mode is only known after the async `startCheckout()`, the container renders asynchronously; `$derived` changes how the value is computed, not when Svelte flushes the DOM, so without the tick `purchase()` would run before the container exists. Open to a cleaner approach if you have one in mind (e.g. triggering purchase from the container's mount), but tick seemed the least invasive.

nicfix (approved):

PR #913, opened 2026-06-10
WFL-217 | Add support for workflows endpoints

RosieWatson:
## Motivation / Description

Add support for workflow endpoints to be used

Here's the flow for both paths in `presentPaywall`:

## Standard Paywall
1. Resolve `offering` (from param or `getOfferings().current`)
2. `workflowsEndpointEnabled` is `false` → skip workflow fetch
3. Validate `offering.paywallComponents` and `offering.uiConfig` exist
4. Mount `Paywall` component with the offering's paywall data

## Workflow
1. Resolve `offering` (same as above)
2. `workflowsEndpointEnabled` is `true` → call `getWorkflows(appUserId)` to get a list of all published workflows
3. Find one whose `offering_id` matches `offering.identifier`
4. If a match is found → call `getWorkflowById(appUserId, workflowId)` to fetch the full workflow data (handles inline or CDN redirect responses)
5. Convert the workflow data to nav data via `workflowDataToNavData`
6. If any step in 4–5 fails → warn and fall back to the standard paywall (or throw if there's no paywall to fall back to)
7. Mount `Workflow` component instead of `Paywall`

## Key differences
- **Two extra network requests** for workflows (`getWorkflows` + `getWorkflowById`) vs none for standard paywalls
- **Workflows are gated** behind `workflowsEndpointEnabled: true` in `FlagsConfig` (internal, defaults `false`)
- **The `offering` is still required** for both paths — workflows use `offering.identifier` to find the right workflow, and `offering.availablePackages` for the purchase flow
- **Locale**: workflow uses `selectedLocale` (raw from params/browser), paywall uses `finalLocale` (calculated from `offering.paywallComponents` localisations)

---

> [!NOTE]
> **Medium Risk**
> Changes core paywall presentation and purchase locale paths when the flag is on, plus extra network/CDN fetches; default-off flag limits exposure but fallback logic affects error behavior for workflow-only offerings.
> 
> **Overview**
> Adds **workflow-backed paywalls** behind the internal flag `workflowsEndpointEnabled` (default `false`). When enabled, `presentPaywall` loads paywall workflows from new subscriber APIs, picks one whose `offering_id` matches the offering, fetches full workflow payload (inline or CDN), and mounts the **`Workflow`** UI instead of **`Paywall`**, with warn-and-fallback to the classic paywall when fetch or nav conversion fails.
> 
> Introduces **`getWorkflows`** and **`getWorkflowById`** on the backend (including `use_cdn` follow-up fetches), response types, and MSW coverage. Paywall validation and locale handling are adjusted so workflow-only offerings can work without `paywallComponents` until fallback is required; purchase from workflows uses `selectedLocale` for checkout. Paywall event tests now **`waitFor`** mount because presentation can be async before the UI is mounted.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2376ba92bfa5f20dad3dc1d1e7d9782281a8f247. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/networking/backend.ts
```diff
@@ -480,4 +489,88 @@ export class Backend {
...
+        `Failed to fetch workflow from CDN: ${cdnResponse.statusText}`,
+      );
+    }
+    return (await cdnResponse.json()) as WorkflowDataResponse;
+  }
+
+  async getWorkflow(
+    workflowLinkId: string,
+    appUserId: string,
+  ): Promise<WorkflowDataResponse> {
```
vegaro: I am a bit confused between this function and the one above

```
  async getWorkflowById(
    appUserId: string,
    workflowId: string,
  ): Promise<WorkflowDataResponse> {
```

look like they accept the same inputs but different order and they return the same

src/main.ts
```diff
@@ -556,11 +564,17 @@ export class Purchases {
     if (!offering) {
       throw new Error("No offering found.");
     }
-    if (!offering.paywallComponents) {
+
+    // Check for a workflow associated with this offering by fetching it directly.
+    const workflowDataResponse = await this.backend
+      .getWorkflowById(this._appUserId, offering.identifier)
+      .catch(() => null);
```
vegaro: this is a change of behavior in the sense that this will always load on paywall presentation. Is that a problem? I am not familiar enough with the SDK but is similar to what we have to change in the native SDKs to make sure it behaved like offerings

src/networking/backend.ts
RosieWatson: So they were doing different things, but since I made some other changes this one isn't actually used anymore - so removing

src/main.ts
RosieWatson: I ummed and ahhed on this one for a while, because the approach i went with first was the norm for the web SDK - we're usually fine with a 404 and quick fallover, unlike in the native SDKs, buuut I think for this one as it's two requests you're right to push back, so have reverted it to being behind an internal feature flag, which is a bit manual but I think better

vegaro (approved):
thanks for making the changes!

PR #915, opened 2026-06-10
Adopt generated error codes

AlvaroBrey:
Replace hand-written `ErrorCode` enum with the file generated by [purchases-error-codes](https://github.com/RevenueCat/purchases-error-codes). No API changes.

To prevent a lot of refactoring, we preserve `ErrorCodeUtils` as a proxy


src/generated/error-codes.ts
```diff
@@ -0,0 +1,102 @@
+// @generated
+// DO NOT EDIT THIS FILE — generated from https://github.com/RevenueCat/purchases-error-codes
+
+/**
+ * Error codes for the Purchases SDK.
+ * @public
+ */
+export enum ErrorCode {
```
tonidero, line 8: Should we add one test that checks that we're using the latest version from the `purchases-error-codes` repo? (basically cloning that repo and diffing the versions) 
I guess we could leave it manually updated... Just afraid it might take a while to update and also that someone might want to modify this file by hand.

src/entities/errors.ts
```diff
@@ -2,113 +2,14 @@ import {
...
-  LogOutWithAnonymousUserError = 22,
-  ConfigurationError = 23,
-  UnsupportedError = 24,
-  EmptySubscriberAttributesError = 25,
-  CustomerInfoError = 28,
-  SignatureVerificationError = 36,
-  InvalidEmailError = 38,
-  TestStoreSimulatedPurchaseError = 42,
-}
+export { ErrorCode };
```
tonidero: Hmm I usually prefer to specify from where we're exporting it, even if we have it from the previous import... NABD I guess.
```suggestion
export { ErrorCode } from "../generated/error-codes";
```

tonidero (approved):
Looking great! Mostly just whether we should add some test for this generated file but could also be done separately. Thank you!!

src/generated/error-codes.ts
AlvaroBrey: The `purchases-error-codes` repo will automatically open PRs to consumers to update the files when the YAML is modified. We could technically forget about that PR and not merge it but I don't think it's a priority.

What I will do however is prevent manual modification of the generated files, that's coming as a follow up

tonidero: Ah cool, sounds good!

src/entities/errors.ts
AlvaroBrey: NABD to change it either! Done

PR #948, opened 2026-06-23
Expose express purchase button API

nicfix:
## Summary
- Use unique DOM ids for Stripe express checkout elements to avoid collisions when rendering multiple buttons
- Remove internal API docs from `presentExpressPurchaseButton`
- Remove the demo app `@ts-expect-error` now that the method is public

---

> [!NOTE]
> **Medium Risk**
> Public purchase API and required `htmlTarget` are breaking/type-surface changes for integrators; wallet checkout DOM id fix reduces multi-button bugs but touches payment UI wiring.
> 
> **Overview**
> **`Purchases.presentExpressPurchaseButton`** and related types (`PresentExpressPurchaseButtonParams`, `ExpressPurchaseButtonUpdater`) are now **public** (still marked experimental), with API Extractor reports and exports updated. Docs now point at `PresentExpressPurchaseButtonParams`, include a usage example for `onButtonReady` / `updatePurchase`, and **`htmlTarget` is required** in the public params (the redundant runtime guard was removed).
> 
> Purchase lifecycle callbacks are unified via a new **`PurchaseListener`** interface; **`PaywallListener`** is now a type alias to it, and express button params use `PurchaseListener` for `listener`.
> 
> Stripe express checkout elements use **`generateUUID()`** for DOM element ids instead of timestamps so multiple buttons on one page do not collide.
> 
> The web billing demo drops the `@ts-expect-error` on the express button call now that the method is public.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 99b90d2f0d6b0a991e5dd9a2d7ba5d2d862e4d14. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/api-report/purchases-js.api.md
+++ b/api-report/purchases-js.api.md
@@ -270,11 +270,7 @@ export enum PackageType {
 }
 
 // @public
-export interface PaywallListener {
-    onPurchaseCancelled?: () => void;
-    onPurchaseError?: (error: Error) => void;
-    onPurchaseStarted?: (rcPackage: Package) => void;
-}
+export type PaywallListener = PurchaseListener;
 
 // @public
 export interface PaywallPurchaseResult extends PurchaseResult {
@@ -319,8 +315,8 @@ export interface PresentedOfferingContext {
 export interface PresentExpressPurchaseButtonParams {
     customerEmail?: string;
     defaultLocale?: string;
-    htmlTarget?: HTMLElement;
-    listener?: PaywallListener;
+    htmlTarget: HTMLElement;
+    listener?: PurchaseListener;
     metadata?: PurchaseMetadata;
     onButtonReady?: (updater: ExpressPurchaseButtonUpdater, walletsAvailable: boolean) => void;
     purchaseOption?: PurchaseOption | null;
@@ -401,6 +397,13 @@ export enum ProductType {
     Consumable = "consumable",
     NonConsumable = "non_consumable",
     Subscription = "subscription"
+}
+
+// @public
+export interface PurchaseListener {
+    onPurchaseCancelled?: () => void;
+    onPurchaseError?: (error: Error) => void;
+    onPurchaseStarted?: (rcPackage: Package) => void;
 }
 
 // @public
--- a/api-report/purchases-js.api.md
+++ b/api-report/purchases-js.api.md
@@ -6,6 +6,7 @@
 
 import { CustomVariables } from '@revenuecat/purchases-ui-js';
 import { CustomVariableValue } from '@revenuecat/purchases-ui-js';
+import { WalletButtonTheme } from '@revenuecat/purchases-ui-js';
 
 // @public
 export interface BrandingAppearance {
@@ -142,6 +143,11 @@ export enum ErrorCode {
     UnsupportedError = 24,
     // (undocumented)
     UserCancelledError = 1
+}
+
+// @public
+export interface ExpressPurchaseButtonUpdater {
+    updatePurchase: (pkg: Package, purchaseOption?: PurchaseOption) => void;
 }
 
 // @public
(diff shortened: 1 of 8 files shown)


src/entities/present-express-purchase-button-params.ts
```diff
@@ -59,7 +57,6 @@ export interface PresentExpressPurchaseButtonParams {
    */
   labelsOverride?: CustomTranslations;
   /**
-   * @internal
    * Callback to be called when the express purchase button is ready to be clicked.
```
tonidero, line 95: Maybe we should mention some sample use cases for this property? Just wondering if it could be confusing to have this...

api-report/purchases-js.api.md
```diff
@@ -309,6 +315,20 @@ export interface PresentedOfferingContext {
     readonly targetingContext: TargetingContext | null;
 }
 
+// @public
+export interface PresentExpressPurchaseButtonParams {
+    customerEmail?: string;
+    defaultLocale?: string;
+    htmlTarget?: HTMLElement;
+    listener?: PaywallListener;
```
tonidero: Hmm it's a bit weird to have a `PaywallListener` as part of this which is not technically a paywall... I wonder if we should just typedef a different listener for this just in case?

api-report/purchases-js.api.md
```diff
@@ -309,6 +315,20 @@ export interface PresentedOfferingContext {
     readonly targetingContext: TargetingContext | null;
 }
 
+// @public
+export interface PresentExpressPurchaseButtonParams {
+    customerEmail?: string;
+    defaultLocale?: string;
+    htmlTarget?: HTMLElement;
```
tonidero: Hmm I'm thinking what's the behavior when there is no htmlTarget? Does it just show the express purchase buttons in a modal somehow? I would love to see some screenshot of this case... Because maybe we want to make this mandatory even if it makes the API a bit more difficult?

tonidero (commented):
Looks great! I'm just unsure how this will look when no `htmlTarget` is passed, and also a rename suggestion. But makes sense overall!

api-report/purchases-js.api.md
nicfix: I agree, this should be mandatory, and we are in time to make it so since we control all the usages right now.
I'll make it mandatory.

tonidero: 👍 Another thing then will be whether we want to rename the method from "present" to something like `injectExpressPurchaseButtons`?

api-report/purchases-js.api.md
nicfix: Maybe we should call it `PurchaseListener` ?

We might need to update the users of the same interface though or create a new one, since this interface is currently exposed by presentPaywall maybe creating a different one is a good idea, or create a parent interface and then just alias the name for paywalls to not break things.

src/entities/present-express-purchase-button-params.ts
nicfix: Will do.

api-report/purchases-js.api.md
tonidero: Yeah... I think `PurchaseListener` could make sense and alias them to avoid a breaking change could work indeed... I'm thinking whether there is any scenario where we would like some callback in one but not the other but nothing comes to mind... but as long as we leave ourselves that door open for the future, it should be fine I think

src/entities/present-express-purchase-button-params.ts
nicfix: Added an example in the latest commit.

api-report/purchases-js.api.md
nicfix: Done, converted to type alias to make ts happy.

api-report/purchases-js.api.md
nicfix: Made htmlTarget as mandatory.

I would prefer to still call it presentExpressPurchaseButton since it aligns with what we do with paywalls.

Also we expect the htmlTarget to be empty and `inject` might suggest that it gets added to whatever is there.

tonidero: Hmm I guess it's a different type of "presentation" 😅 but yeah, I think it's probably fine.

src/entities/present-express-purchase-button-params.ts
```diff
@@ -29,9 +62,9 @@ export interface PresentExpressPurchaseButtonParams {
    */
   purchaseOption?: PurchaseOption | null;
   /**
-   * The HTML element where the billing view should be added. If undefined, a new div will be created at the root of the page and appended to the body.
+   * The HTML element where the express purchase button should be rendered.
    */
-  htmlTarget?: HTMLElement;
+  htmlTarget: HTMLElement;
```
tonidero, line 63: Nitpick... but would it make sense to put the mandatory parameters first? Mostly so they are easier to find when reading the docs, that they are mandatory.

tonidero (approved):
Just a nitpick, but looks great overall!

src/entities/present-express-purchase-button-params.ts
nicfix: On it!

nicfix: Done

PR #951, opened 2026-06-24
WEB-4012 Add checkout back button visibility flag

nicfix:
## Summary
- Add `flags.hideBackButton` to the purchases-js configuration surface
- Propagate the flag through RevenueCat Web Billing and Paddle checkout UI paths
- Hide RevenueCat-rendered checkout back/return buttons when the flag is enabled
- Add coverage for Web Billing and Paddle inline checkout behavior

## Testing
- `pnpm typecheck`
- `pnpm test src/tests/main.test.ts src/tests/ui/paddle-purchases-ui.test.ts`

## Notes
- This does not change Paddle callback behavior; it only hides RevenueCat-rendered back/return UI.

---

> [!NOTE]
> **Low Risk**
> UI-only visibility for checkout navigation; purchase cancel/close logic is unchanged aside from reusing existing rcSource rules.
> 
> **Overview**
> Adds **`flags.hideBackButton`** on purchases-js configure so integrators can hide RevenueCat-rendered checkout back/return controls on Web Billing and Paddle inline checkout.
> 
> **`shouldHideCheckoutBackButton()`** in the purchase router sets visibility for both flows: explicit `hideBackButton: true`, or when **`rcSource`** is `app` / `embedded` (same sources that already alter close behavior). That value is passed into **`PurchasesUI`** / **`PaddleInlineCheckoutPage`**, which suppress the navbar close button and Paddle return button respectively—without changing cancel/callback semantics.
> 
> The webbilling demo maps **`hideCheckoutBackButton`** query params into the flag; unit and integration tests cover the new behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e10f360cabb382a98556136e3edb04a561d78ef2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/api-report/purchases-js.api.md
+++ b/api-report/purchases-js.api.md
@@ -6,7 +6,6 @@
 
 import { CustomVariables } from '@revenuecat/purchases-ui-js';
 import { CustomVariableValue } from '@revenuecat/purchases-ui-js';
-import { WalletButtonTheme } from '@revenuecat/purchases-ui-js';
 
 // @public
 export interface BrandingAppearance {
@@ -146,11 +145,6 @@ export enum ErrorCode {
 }
 
 // @public
-export interface ExpressPurchaseButtonUpdater {
-    updatePurchase: (pkg: Package, purchaseOption?: PurchaseOption) => void;
-}
-
-// @public
 export interface FlagsConfig {
     autoCollectUTMAsMetadata?: boolean;
     collectAnalyticsEvents?: boolean;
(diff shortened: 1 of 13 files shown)


api-report/purchases-js.api.md
```diff
@@ -154,6 +154,7 @@ export interface ExpressPurchaseButtonUpdater {
 export interface FlagsConfig {
     autoCollectUTMAsMetadata?: boolean;
     collectAnalyticsEvents?: boolean;
+    hideBackButton?: boolean;
```
vicfergar, line 157: Would it make sense to use an enum instead, with options like `auto` as the default and `never`?

vicfergar (approved):
LGTM, but I left a suggestion about using an enum instead of a boolean.

api-report/purchases-js.api.md
nicfix: I think it's a bit overkill, in the end it's just hiding or showing it and it overrides any other behaviour.

PR #954, opened 2026-06-27
feat(offerings): WST-823 build composite product id from the plan field

vkartaviy:
## What

Composite Stripe products carry their price on a separate `platform_product_plan_identifier` field (the same shape Google Play uses for base plans). The web SDK now keys a package's product by the composite id `{product}:{plan}` whenever that plan field is present, so two price variants of one Stripe product resolve to their own products instead of collapsing onto the shared bare product id. Packages with no plan field (legacy / non-composite) key by the product id alone, unchanged.

## Changes

- `PackageResponse` models `platform_product_plan_identifier`.
- New `packageProductIdentifier(pkg)` helper - the single source of truth for a package's product key.
- `getAllOfferings` fetches products by that key, and `toPackage` looks them up by the same key. Same helper on both sides, so the fetch and the map can't drift.

## Behavior

| Package | Product key |
|---------|-------------|
| Composite (plan field present) | `prod:price` |
| Legacy / non-composite | `prod` (unchanged) |

## Tests

- `packageProductIdentifier` returns the composite id with a plan field, the bare id without.
- `toOffering` resolves a composite package's product by the composite id.
- Existing offerings suite (32 tests) green - non-composite packages unchanged.

## Backend dependency + version coordination

Pairs with the khepri offerings gate `offering-stripe-composite` (`sdk='js'`): the backend serves the bare id + plan field only to clients at or above that gate version, and folds the composite id into `platform_product_identifier` for everything below. So this SDK must release at the gate's `js` min version (currently `1.47.0`) - confirm the two line up when it ships. Composite price resolution in `/products` requires khepri WST-730.

---

> [!NOTE]
> **Medium Risk**
> Touches core offerings/product resolution for purchases; legacy non-composite behavior is preserved, but mismatched backend/SDK versions could drop packages if composite products are missing from `/products`.
> 
> **Overview**
> Composite Stripe packages can now resolve **distinct products per price** instead of sharing one bare product id when multiple packages point at the same Stripe product.
> 
> `PackageResponse` gains optional `platform_product_plan_identifier`, and **`packageProductIdentifier`** centralizes the lookup key: `{platform_product_identifier}:{platform_product_plan_identifier}` when a plan is present, otherwise the bare product id (unchanged for legacy packages). **`getAllOfferings`** requests products with that key, and **`toPackage`** reads from the product map with the same key so fetch and resolution stay aligned.
> 
> Tests cover the helper’s composite vs bare behavior and **`toOffering`** wiring for a composite package.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d175b7b70ff459e5cdef44425c0e75adadd2bcf8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/entities/offerings.ts
+++ b/src/entities/offerings.ts
@@ -890,13 +890,25 @@ const toSubscriptionProduct = (
   };
 };
 
+/**
+ * The product key for a package. Composite Stripe products carry the price on a
+ * separate plan field, so they key by `{product}:{plan}` and each price variant
+ * resolves to its own product.
+ */
+export const packageProductIdentifier = (
+  packageData: PackageResponse,
+): string =>
+  packageData.platform_product_plan_identifier
+    ? `${packageData.platform_product_identifier}:${packageData.platform_product_plan_identifier}`
+    : packageData.platform_product_identifier;
+
 const toPackage = (
   presentedOfferingContext: PresentedOfferingContext,
   packageData: PackageResponse,
   productDetailsData: { [productId: string]: ProductResponse },
 ): Package | null => {
   const webBillingProduct =
-    productDetailsData[packageData.platform_product_identifier];
+    productDetailsData[packageProductIdentifier(packageData)];
   if (webBillingProduct === undefined) return null;
   const product = toProduct(webBillingProduct, presentedOfferingContext);
   if (product === null) return null;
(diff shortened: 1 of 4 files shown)


vkartaviy:
Closing - the composite id is now served directly in `platform_product_identifier` by the offerings backend (https://github.com/RevenueCat/khepri/pull/21999), so the existing SDK resolves the right price variant through the `/products` lookup with no SDK change needed.

PR #960, opened 2026-06-29
[WEB-4434] feat: collect full billing address for tax-relevant countries

vicfergar:
## Motivation / Description

Stripe's tax calculation needs the full billing address (not just country +
postal code) to resolve the rate in certain countries. Until now the full
address was only collected when the project had `full_address_collection_mode`
set to `"always"`. This PR makes the checkout form switch to full-address
collection automatically when it's actually required for taxes, while keeping
the minimal country + postal code form for everyone else.

## Changes introduced

- The full billing address element is now shown when **tax collection is enabled**
  and the country selected in the payment element is one that requires the full
  address to resolve taxes (`CA`, `PR`, `IN`).
- The transition is driven by the payment element's `change` event: when a
  tax-relevant country is selected we reveal the dedicated address element and
  seed it with the already-selected country so it opens on the right one.
- Once the full address has been required, it **latches on** for the rest of the
  session. There is no way to push a country back into the payment element, so
  toggling back to the minimal form would lose the selection and flip-flop the
  UI; latching keeps the address element as the single source of truth and lets
  the customer freely change the country afterwards (e.g. Canada → Bulgaria).
- `shouldCollectFullAddress(brandingInfo, selectedCountryCode?)` gains the
  tax + country logic via `FULL_ADDRESS_REQUIRED_TAX_COUNTRY_CODES` (`CA`, `PR`,
  `IN`); `stripe-elements.svelte` and `payment-entry-page.svelte` apply the latch
  (rendering and pay-button gating respectively).
- The existing `full_address_collection_mode: "always"` behavior is unchanged.
- Tests: unit coverage for the dynamic trigger, not-tax-relevant / tax-disabled
  cases, the latch (CA → BG), and the address-element country seed; plus an
  integration test in `tax-calculation.test.ts` where selecting Canada reveals
  the full address form and performs the tax calculation (mocked + real,
  Chromium). The billing-address test helper now fills country-specific labels
  (Province / Postal code).

## Linear ticket (if any)

WEB-4434

## Additional comments

Trade-off: after a tax-relevant country has been selected, switching to a
non-required country keeps the full-address fields rather than collapsing back to
the minimal form. Collapsing would require recreating the payment element (losing
the entered card details), so latching is the deliberate choice that avoids the
mount/unmount loop.

---

> [!NOTE]
> **Medium Risk**
> Changes checkout tax/address UX and when tax recalculations run, which can affect displayed totals and pay-button enablement for international customers; behavior is covered by new unit and integration tests.
> 
> **Overview**
> When gateway tax collection is on and `full_address_collection_mode` is `if_required`, checkout now **reveals the Stripe Address Element** after the customer picks a country in the payment form that Stripe Tax cannot rate from country alone (**CA**, **PR**, **IN**), instead of only showing full address when the project forces `always`.
> 
> **`shouldCollectFullAddress`** now considers tax collection plus whether the selected country needs a full address; **`StripeService.countryRequiresFullAddressForTaxes`** encodes those countries. The payment element’s country is tracked, the address element is **seeded** with that country, and once full address is required it **latches** for the session (so switching to a non-required country does not collapse the form). Pay-button and tax-refresh gating follow the same dynamic full-address requirement.
> 
> Tests add Canada fixtures, country-specific E2E helpers (Province / Postal code), a Chromium integration path for Canada → full address → tax, and unit coverage for the new trigger, latch, and seed behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8530d7ab9b86b647451166d4b9da62c3bbfc717d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/networking/responses/branding-response.ts
+++ b/src/networking/responses/branding-response.ts
@@ -28,22 +28,15 @@ export type BrandingInfoResponse = {
 };
 
 /**
- * Country codes (ISO 3166-1 alpha-2) that require the full billing address to be
- * collected when tax collection is enabled, because country alone
- * is not enough to resolve the tax rate.
- * see https://docs.stripe.com/tax/customer-locations?#supported-formats
- */
-export const FULL_ADDRESS_REQUIRED_TAX_COUNTRY_CODES = ["CA", "PR", "IN"];
-
-/**
  * Resolves whether the full billing address UI should be presented, based on the
- * branding configuration and the currently selected country.
+ * branding configuration and whether the selected country requires the full
+ * address to resolve taxes.
  *
  * The full address is collected when either:
  * - `full_address_collection_mode` is `always`, or
- * - tax collection is enabled and the selected country is one that requires the
- *   full address to resolve taxes (see
- *   {@link FULL_ADDRESS_REQUIRED_TAX_COUNTRY_CODES}).
+ * - tax collection is enabled and the selected country requires the full address
+ *   to resolve taxes (determined by the payment gateway, e.g.
+ *   `StripeService.countryRequiresFullAddressForTaxes`).
  *
  * Treats a missing mode and any unknown/future mode as `if_required`.
  */
@@ -55,7 +48,7 @@ export function shouldCollectFullAddress(
       >
     | null
     | undefined,
-  selectedCountryCode?: string | null,
+  selectedCountryRequiresFullAddressForTaxes: boolean = false,
 ): boolean {
   if (brandingInfo?.full_address_collection_mode === "always") {
     return true;
@@ -63,7 +56,6 @@ export function shouldCollectFullAddress(
 
   return (
     !!brandingInfo?.gateway_tax_collection_enabled &&
-    !!selectedCountryCode &&
-    FULL_ADDRESS_REQUIRED_TAX_COUNTRY_CODES.includes(selectedCountryCode)
+    selectedCountryRequiresFullAddressForTaxes
   );
 }
--- a/src/networking/responses/branding-response.ts
+++ b/src/networking/responses/branding-response.ts
@@ -28,15 +28,42 @@ export type BrandingInfoResponse = {
 };
 
 /**
- * Resolves whether the full billing address UI should always be presented,
- * based on the branding configuration. Treats a missing mode and any
- * unknown/future mode as `if_required`.
+ * Country codes (ISO 3166-1 alpha-2) that require the full billing address to be
+ * collected when tax collection is enabled, because country alone
+ * is not enough to resolve the tax rate.
+ * see https://docs.stripe.com/tax/customer-locations?#supported-formats
+ */
+export const FULL_ADDRESS_REQUIRED_TAX_COUNTRY_CODES = ["CA", "PR", "IN"];
+
+/**
+ * Resolves whether the full billing address UI should be presented, based on the
+ * branding configuration and the currently selected country.
+ *
+ * The full address is collected when either:
+ * - `full_address_collection_mode` is `always`, or
+ * - tax collection is enabled and the selected country is one that requires the
+ *   full address to resolve taxes (see
+ *   {@link FULL_ADDRESS_REQUIRED_TAX_COUNTRY_CODES}).
+ *
+ * Treats a missing mode and any unknown/future mode as `if_required`.
  */
 export function shouldCollectFullAddress(
   brandingInfo:
-    | Pick<BrandingInfoResponse, "full_address_collection_mode">
+    | Pick<
+        BrandingInfoResponse,
+        "full_address_collection_mode" | "gateway_tax_collection_enabled"
+      >
     | null
     | undefined,
+  selectedCountryCode?: string | null,
 ): boolean {
-  return brandingInfo?.full_address_collection_mode === "always";
+  if (brandingInfo?.full_address_collection_mode === "always") {
+    return true;
+  }
+
+  return (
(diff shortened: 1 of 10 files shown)


src/ui/pages/payment-entry-page.svelte
```diff
@@ -176,7 +176,22 @@
...
+  // address is required (a tax-relevant country was selected in the payment
+  // element) it stays required for the rest of the session.
+  let collectFullBillingAddress = $state(
+    shouldCollectFullAddress(brandingInfo),
+  );
+  $effect(() => {
+    if (shouldCollectFullAddress(brandingInfo, selectedCountry)) {
+      collectFullBillingAddress = true;
+    }
+  });
```
nicfix, line 200: Nit: Maybe you can achieve this with a $derived instead of a $state + $effect.

src/ui/molecules/stripe-elements.svelte
```diff
@@ -76,9 +77,25 @@
...
+  // needed for tax purposes.
+  let selectedCountry: string | undefined = $state(undefined);
+
+  // Once the full billing address is required (a tax-relevant country was
+  // selected) we keep collecting it for the rest of the session. There is no way
+  // to push a country back into the payment element, so switching back to the
+  // minimal form would lose the selection; latching keeps the address element as
+  // the single source of truth and lets the customer freely change the country
+  // afterwards (e.g. Canada -> Bulgaria).
+  let collectFullBillingAddress = $state(
```
nicfix, line 95: Nit: you might be able to use $derived for this instead of $state + $effect

src/networking/responses/branding-response.ts
```diff
@@ -28,15 +28,42 @@ export type BrandingInfoResponse = {
...
 /**
- * Resolves whether the full billing address UI should always be presented,
- * based on the branding configuration. Treats a missing mode and any
- * unknown/future mode as `if_required`.
+ * Country codes (ISO 3166-1 alpha-2) that require the full billing address to be
+ * collected when tax collection is enabled, because country alone
+ * is not enough to resolve the tax rate.
+ * see https://docs.stripe.com/tax/customer-locations?#supported-formats
+ */
+export const FULL_ADDRESS_REQUIRED_TAX_COUNTRY_CODES = ["CA", "PR", "IN"];
```
nicfix: Shouldn't they be \["CA", "PR", "GB"\] ?<br>[https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2#GB](<https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2#GB>)

src/networking/responses/branding-response.ts
nicfix, line 1: NIT: Maybe these changes should be somewhere closer to the Stripe Elements, and/or the StripeService.<br><br>In the end of the day they are very Stripe specific, wdyt?

nicfix (commented):
A couple of nits and a question on the `FULL_ADDRESS_REQUIRED_TAX_COUNTRY_CODES`, I think we should use or add GB for the UK.<br><br>Also the current behaviour never removes the full address element when we switch back to a non full address country.<br><br>Might be good enough but I just wanted to be sure it was intentional.

src/ui/pages/payment-entry-page.svelte
vicfergar: I tested UK tax collection, and it worked for me with only the country entered, so I’d prefer to keep the UK out of this path to reduce friction.

I used the attached Stripe documentation page to decide which countries to include in this path.

src/networking/responses/branding-response.ts
vicfergar: As mentioned above, I only included Canada, Puerto Rico and India in this path.

Source: https://docs.stripe.com/tax/customer-locations?#supported-formats

src/ui/molecules/stripe-elements.svelte
vicfergar: I'm not sure about this TBH, this is the justification from Cursor:

> Not correct. collectFullBillingAddress is intentionally latching — once true, it must stay true even if the customer later switches to a non-tax-relevant country (e.g. Canada → Bulgaria), as the comment explains and a test asserts.
> 
> A $derived recomputes on every dependency change, so it would flip back to false and tear down the address element. $derived can't express "only ever go false → true", which is exactly why $state + $effect is used here.

src/ui/pages/payment-entry-page.svelte
nicfix: Makes sense, thanks!

src/ui/molecules/stripe-elements.svelte
nicfix: That's true, $derived will make it change back to false.

However, maybe that could be a way to hide the address element once the country changes back to a non full address tax one.

If we are fine to keep the address on once it's shown once then I am fine with keeping it like this.

nicfix: PS: this is a nit and it's absolutely non blocking

vicfergar: As explained in the description, we cannot change back to a non full address UI, as we cannot push a country back into the payment element

src/networking/responses/branding-response.ts
vicfergar: Addressed in 4340eff75414db62171a0daba1797cb0a223160d

vicfergar:
@claude review this PR

nicfix (approved):

PR #976, opened 2026-07-15
[WST-851] Pass discount codes to Stripe Billing checkout

nicfix:
## Why

Developers need to pass an initial `discountCode` to Stripe Billing checkout. The SDK must resolve that code into the purchase option returned by Khepri and use the selected option when starting checkout.

## What changed

- When a Stripe purchase receives `discountCode`, request the selected product from `/products` with that code.
- Use the default purchase option returned by Khepri. A valid Stripe Promotion Code is represented by an opaque ID such as `stripe_promo;code=SUMMER20`.
- Send every non-base option to `/checkout/start` as `purchase_option_id`.
- Keep the checkout-start request based on `purchase_option_id`; it does not send a separate Stripe `discount_code`.
- Keep the Web Billing example's optional discount-code input so the complete flow can be smoke-tested.

## Compatibility and scope

- Calls without `discountCode` keep using the package's selected or default purchase option without an additional product request.
- If Khepri returns the base option because promotion codes are disabled or the code is invalid, checkout continues without a pre-applied discount.
- The existing RevenueCat Web Billing discount input and checkout-pricing refresh flow remain unchanged.
- Purchase options returned for Stripe promotion codes retain the product's normal displayed price in this first step. Synthetic discounted pricing and product eligibility are deferred.
- `discountCode` remains an experimental public SDK field in this PR. Stabilizing the related discount-code APIs will be handled separately.

Backend support is provided by [khepri#22676](https://github.com/RevenueCat/khepri/pull/22676).


nicfix:
This PR is no more needed since we moved to the purchase_option_id mechanism on backend side.

PR #979, opened 2026-07-17
[WEB-4481] Support consent checkbox at checkout

james-od:
## Motivation / Description

Adds the option to opt in to display an explicit consent checkbox to unlock the purchase. If opted into the payment will not be able to be triggered until the checkbox is checked. This also means wallet payment methods are hidden until the checkbox is checked.

Made the `termsAndConditionsUrl` param not-internal - it was added a few months back as a customer request, I don't think theres a reason not to do this? Unlocks support for using the consent checkbox via the SDK.

Bumped translation model to `"gemini-2.5-flash",`.

Checkout consent only displays if we have the terms URL, the explicit opt-in, and if the product is a subscription. 

We need to do some mildly clever string manip to get the link on the correct part of the string, while remaining localisation friendly.

See https://github.com/RevenueCat/revenuecat-app/pull/10858 for recording.

## Additional comments

---

> [!NOTE]
> **Medium Risk**
> Changes payment submission and wallet express-checkout gating on the payment entry page, which is checkout-critical but scoped to an opt-in branding flag and subscriptions only.
> 
> **Overview**
> Adds an **optional checkout consent checkbox** for subscription purchases when branding sets `require_checkout_consent`, a **`termsAndConditionsUrl`** is provided, and the product is a subscription. Until the box is checked, **Pay stays disabled** and **express/wallet checkout is blocked** (hidden and non-interactive).
> 
> **`termsAndConditionsUrl`** on `PurchaseParams` is now **public** (no longer `@internal`) so SDK integrators can supply the terms link for the consent UI.
> 
> Branding responses gain **`require_checkout_consent`**. A new **`CheckoutConsent`** molecule renders localized copy with an inline terms link (placeholder-based splitting for i18n). **`isCheckoutConsentRequired`** centralizes the gating rules.
> 
> Locale strings were added across languages; the locale translation script model was bumped to **gemini-2.5-flash**. Tests, Storybook, and API reports were updated accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 65e9d9d6332be7ac0703078daf9cb2af6b9b7a0a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/stories/pages/payment-entry-page.stories.svelte
+++ b/src/stories/pages/payment-entry-page.stories.svelte
@@ -65,6 +65,7 @@
     full_address_collection_mode:
       args.fullAddressCollectionMode ??
       brandingInfos[context.globals.brandingName].full_address_collection_mode,
+    require_checkout_consent: args.requireCheckoutConsent ?? false,
   }}
   <PurchasesInner
     isSandbox={args.isSandbox}
@@ -103,6 +104,21 @@
 <Story
   name="Default"
   args={{ ...defaultArgs, currentPage: "payment-entry" }}
+  parameters={{
+    chromatic: {
+      delay: 1000,
+    },
+  }}
+/>
+
+<Story
+  name="With Checkout Consent"
+  args={{
+    ...defaultArgs,
+    currentPage: "payment-entry",
+    requireCheckoutConsent: true,
+    termsAndConditionsUrl: "https://example.com/terms",
+  }}
   parameters={{
     chromatic: {
       delay: 1000,
--- a/src/ui/molecules/checkout-consent.svelte
+++ b/src/ui/molecules/checkout-consent.svelte
@@ -49,10 +49,10 @@
   };
 </script>
 
-<label class="rc-checkout-consent" data-testid="CheckoutConsent">
+<label class="rcb-checkout-consent" data-testid="CheckoutConsent">
   <input
     type="checkbox"
-    class="rc-checkout-consent-checkbox"
+    class="rcb-checkout-consent-checkbox"
     data-testid="CheckoutConsentCheckbox"
     {checked}
     onchange={handleChange}
@@ -71,7 +71,7 @@
 </label>
 
 <style>
-  .rc-checkout-consent {
+  .rcb-checkout-consent {
     display: flex;
     align-items: flex-start;
     gap: 8px;
@@ -79,12 +79,12 @@
     cursor: pointer;
   }
 
-  .rc-checkout-consent-checkbox {
+  .rcb-checkout-consent-checkbox {
     margin-top: 2px;
     flex-shrink: 0;
   }
 
-  .rc-checkout-consent :global(a) {
+  .rcb-checkout-consent :global(a) {
     color: inherit;
     text-decoration: underline;
   }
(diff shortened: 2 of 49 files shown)


src/stories/pages/payment-entry-page.stories.svelte
```diff
@@ -110,6 +111,21 @@
   }}
 />
 
+<Story
```
lburdock, line 114: I noted this in the chromatic screenshot as well:
The snapshot cuts off the bottom of the form on mobile, so the checkbox isn't shown. We should investigate how to make these stories display the entire form, but that doesn't need to block this PR!

src/ui/molecules/stripe-express-checkout-element.svelte
```diff
@@ -124,13 +129,22 @@
...
+  <TextSeparator
+    text={$translator.translate(
+      LocalizationKeys.PaymentEntryPageExpressCheckoutDivider,
+    )}
+  />
 {/if}
+
+<style>
+  .rc-express-checkout-hidden {
+    display: none;
```
lburdock, line 148: Not blocking: Would disabling the button until the checkbox is selected create a smoother experience by making it clear that express checkout is an option (once the terms and conditions are accepted)? Or would that add more confusion?

src/ui/molecules/checkout-consent.svelte
```diff
@@ -0,0 +1,91 @@
...
+  });
+
+  const handleChange = (event: Event) => {
+    const target = event.currentTarget as HTMLInputElement;
+    checked = target.checked;
+    onChange(target.checked);
+  };
+</script>
+
+<label class="rc-checkout-consent" data-testid="CheckoutConsent">
```
lburdock: [nit] Most of the one-off classes start with `rcb-` instead of `rc-`.

lburdock (approved):

src/stories/pages/payment-entry-page.stories.svelte
james-od: Yeah I did spot that, agree it should be investigated but opted not to follow up in this scope. Still get most of the coverage from the desktop snaps.

src/ui/molecules/stripe-express-checkout-element.svelte
james-od: Maybe, we considered a few different options here. It's definitely not super smooth atm agree, but I'm not sure disabling helps. I think this is a reasonable slice to ship, and allows us to play with it a bit more and think about potential iterations from here. So I'm going to roll with what we already have to unblock.

src/ui/molecules/checkout-consent.svelte
james-od: Good spot!

src/ui/pages/payment-entry-page.svelte
```diff
@@ -498,6 +510,7 @@
     e?.preventDefault();
 
     if (processing) return;
+    if (checkoutConsentRequired && !checkoutConsentAccepted) return;
```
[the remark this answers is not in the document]
james-od: I don't think either of these cases are possible in practise for a human

PR #996, opened 2026-07-28
[WEB-4522] Upgrade checkout UI

james-od:
## Motivation / Description

Linked with https://github.com/RevenueCat/khepri/pull/23189

This PR establishes the new SDK method that will support the new subscription upgrade flow, as well as a test harness for local development.

A new SDK method `Purchases.presentProductChange({ newProductId, subscriberToken, subscriptionId })` is added, marked as @internal at the moment due to it's ongoing development.
This method operates similarly to other methods which inject some UI into the page. In this case we show a screen similar to the checkout which provides a way to upgrade/downgrade an ongoing subscription.

The page shows the current subscription, the target subscription, the prices of both, the direction, line items, tax status (including whether or not this is deferred), payment method, email and billing address.
I've included several stories to add coverage here.
Note that the UI currently is very much in a PoC stage, with designs to be confirmed later this week. So copy, fonts, positioning etc are all arbitrary. I do still think this is mergeable as-is though. The key point is whether the correct data is being passed in.

The test harness is a super minimal node JS server that handles the token requests. It's run separately with instructions provided in the README.
Components in pricing table are reused in the new page.

https://github.com/user-attachments/assets/e99fb3ee-1923-48ac-b5c5-322e4f18201e

Screenshot 2026-07-28 at 16 43 32

## Additional comments

---

> [!NOTE]
> **Medium Risk**
> Touches subscription changes and confirm-time charging on the payment method on file, but the API is internal/experimental and subscriber tokens are validated to avoid misusing API keys; primary risk is incorrect confirm behavior or auth handling in the new checkout path.
> 
> **Overview**
> Adds **experimental, `@internal`** upgrade-mode checkout so Web Billing subscribers can change products along configured product change paths.
> 
> **SDK:** New `Purchases.changeProduct` mounts upgrade checkout UI (modal or `htmlTarget`), runs **start → confirm** via `ProductChangeOperationHelper`, and returns `ProductChangeResult` (`immediate` vs `deferred`). Calls new RC Billing endpoints with **Bearer subscriber tokens** (not API keys); `validateSubscriberToken` blocks key-shaped values. Success clears in-memory caches; flow reuses existing checkout close/error/history behavior.
> 
> **UI:** `UpgradeCheckoutUi` shows from/to products, pricing (immediate due today vs deferred next-renewal estimate), email, payment method, and billing address; `PricingTable` gains optional `pendingTaxLabel` / `totalRowLabel` for upgrade copy.
> 
> **Demo:** `webbilling-demo` adds `/upgrade/:app_user_id`, a standalone **token server** (`pnpm run token-server`) that mints tokens via Developer API `authenticate`, and Vite proxy `/api` → token server so secrets stay out of the frontend bundle.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 63698164d1a96ab98391e568ab048e1f4aa12291. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/main.ts
```diff
@@ -2138,6 +2148,114 @@ export class Purchases {
...
+   * end of the current billing cycle.
+   *
+   * @param params - The {@link PresentProductChangeParams} for the change.
+   * @returns The {@link ProductChangeResult} describing the applied change.
+   * @throws {@link PurchasesError} if the token is invalid, the subscription
+   * cannot be changed, or no product change path is configured.
+   * @internal
+   */
+  @requiresLoadedResources
+  public async presentProductChange(
```
nicfix: Maybe we can just call this `changeProduct` as we do with `purchase`.

Wdyt @tonidero ?

src/main.ts
```diff
@@ -2138,6 +2148,114 @@ export class Purchases {
...
+      isPaddleApiKey(subscriberToken) ||
+      isStripeApiKey(subscriberToken) ||
+      isSimulatedStoreApiKey(subscriberToken) ||
+      subscriberToken.startsWith("sk_");
+
+    if (looksLikeApiKey || !subscriberToken) {
+      throw new PurchasesError(
+        ErrorCode.ConfigurationError,
+        "Invalid subscriber token.",
+        "The subscriberToken must be a short-lived subscriber access token " +
```
nicfix: Will this error message reach the UI?
If so, I think it explains too much of how internally the upgrade works to a subscriber.

src/main.ts
```diff
@@ -2138,6 +2148,114 @@ export class Purchases {
...
+          unmount(component);
+        }
+        certainHTMLTarget.innerHTML = "";
+      };
+
+      const onClose = this.createCheckoutOnCloseHandler(reject, unmountUi);
+
+      if (!isInElement && onClose) {
+        win.addEventListener("popstate", onClose as EventListener);
+      }
```
nicfix, line 2215: Nit: this should probably be something reused across the various methods that support the back/close behavior. 

Can you create a ticket for that with eng-backlog label?

nicfix (approved):
Looks good overall, I would just change the name of the method in `changeProduct`.

The `present` prefix is something we kept from the old sdks for presenting the paywall, and eventually made it also in the ExpressPurchaseButton but I think the change of product shouldn't carry it forward.

src/main.ts
rickvdl: My 2cts: I like `changeProduct` as you suggested, seems more consistent with the naming of other APIs, and it conceptually does more than just presenting it.

src/entities/product-change-params.ts
```diff
@@ -0,0 +1,53 @@
...
+   * The product identifier of the Web Billing product to change the
+   * customer's current subscription to. A product change path from the
+   * current product to this product must be configured in RevenueCat.
+   */
+  newProductId: string;
+  /**
+   * A short-lived subscriber access token authenticating the current
+   * customer. This must be minted server-side using a secret API key via the
+   * RevenueCat Developer API `authenticate` endpoint, and passed to the
+   * browser. Never use a RevenueCat API key here.
```
tonidero: I don't understand this point about never using a RevenueCat API key here... I don't think anyone should use that here? I'm just thinking that might be more confusing...

src/entities/product-change-params.ts
```diff
@@ -0,0 +1,53 @@
...
+   */
+  subscriberToken: string;
+  /**
+   * RevenueCat subscription public id (`sub…`) of the subscription to change.
+   *
+   * Typically obtained by the developer backend via the Developer API
```
tonidero: Maybe add a link to the docs for the developer API? Also
```suggestion
   * Typically obtained by your backend via the Developer API
```

src/main.ts
tonidero: Yeah, I like `changeProduct` 👍

src/main.ts
```diff
@@ -213,6 +218,10 @@ export type {
   PurchaseResponseAttributionMetadata,
   PurchaseParams,
 } from "./entities/purchase-params";
+export type {
+  PresentProductChangeParams,
+  ProductChangeResult,
+} from "./entities/product-change-params";
```
tonidero, line 225: There is something wrong... I see we didn't update the api extractor files but the tests are passing... This is supposed to make the tests fail so we get notified of accidental API changes/things we're exposing. I think we should fix that to help visualize what we're exporting as part of the public API.

src/entities/product-change-params.ts
```diff
@@ -0,0 +1,53 @@
...
+   * customer subscriptions list and passed to the client with the token.
+   */
+  subscriptionId: string;
```
tonidero, line 26: One thought I had is that, if we need both the `subscriberToken` + `subscriptionId` from the developer API, would it make sense to just join them together somehow, so it's easier on the developer side? But well, it could also make sense to keep them separate I guess, since they represent different information after all, so please ignore this, just rambling :P

tonidero (commented):
I think we should run api-extractor to generate the files and figure out why tests are passing without that... Other than that, the API makes sense to me!

src/main.ts
james-od: It's because they're marked @internal no?

src/main.ts
james-od: Ah ok cool - I'd intentionally opted for consistency with the prefix because this function will actually 'present' (i.e mount in the UI) and not immediately 'change' but happy to go with the preference!
https://github.com/RevenueCat/purchases-js/pull/996/commits/80acc76db33eb5499b340e7257e958e18b610198

src/main.ts
james-od: Mm, it would be raised up to the client calling the SDK and then up to them whether it reaches the UI. But either way I think just "Invalid subscriber token" is fine. 
https://github.com/RevenueCat/purchases-js/pull/996/commits/c761b3b4f14c03931eecacb4415cb7696f35c836

src/main.ts
james-od: +1 , I'd started to look at this but there are some variances. 
https://linear.app/revenuecat/issue/WEB-4540/sdk-refactor-reusable-method-for-backclose-behaviour

src/entities/product-change-params.ts
james-od: Agree! https://github.com/RevenueCat/purchases-js/pull/996/commits/152051ae65881453675d18b823003f5f1d0c95e1

src/entities/product-change-params.ts
james-od: Agree! https://github.com/RevenueCat/purchases-js/pull/996/commits/152051ae65881453675d18b823003f5f1d0c95e1

src/entities/product-change-params.ts
nicfix: The idea it's not completely off.

@james-od is the token right now carrying information about the subscription_id ?

james-od: Not currently. We could consider returning some additional data from the call to the auth endpoint, but we'd still run into the 'which subscription' limitation somewhere.

src/main.ts
tonidero: Ahhh 🤦 yeah sorry. We're good then!

src/ui/upgrade-checkout-ui.svelte
```diff
@@ -0,0 +1,242 @@
...
+      const error =
+        e instanceof PurchaseFlowError
+          ? e
+          : new PurchaseFlowError(
+              PurchaseFlowErrorCode.ErrorChargingPayment,
+              "Failed to confirm product change.",
+              e instanceof Error ? e.message : String(e),
+            );
+      loadError = error.message;
+      onError(error);
```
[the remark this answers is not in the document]
james-od: Rolling this into https://linear.app/revenuecat/issue/WEB-4542/sdk-investigate-edge-caseserror-handling

src/entities/product-change-params.ts
nicfix: Ok so I would keep it for now.

nicfix (approved):

PR #1008, opened 2026-07-31
[WEB-4545] Improved product changes API

james-od:
## Motivation / Description

This PR:
- Folds the `changeProduct` method into `Purchases.purchase()` via a new prop on that method `productChangeInfo`
    - (Note both the before and after state are marked @internal)
- Accepts a subscriber access token on both Purchases.configure({ subscriberToken }) and in productChangeInfo.subscriberToken. Note that the latter will override the former. 

One key note here is that the target product is now always taken from rcPackage.webBillingProduct.identifier. I think this is the most consistent with the regular checkout as well as the other SDKs.
This is somewhat more restrictive that the previous `changeProduct` method which instead took a productIdentifier directly. This allowed more control over the product change as any destination could be specified (as long as a change path existed) but we now require a valid offering to be passed. I imagine this is ok as it will be familiar to developers as the regular checkout path already requires a package (and therefore offering). 

The example demo has been updated to reflect these changes. No changes needed in Khepri.

Getting the subscription id in order to pass as the source is still not ideal, will see if we can improve that but in a separate follow up.

https://github.com/user-attachments/assets/076e37d4-11f3-42b1-85e8-f8dc34c7b454

## Additional comments

---

> [!NOTE]
> **Medium Risk**
> Changes the internal product-change entry point and purchase return shape for upgrades; misconfigured tokens or subscription ids still fail at checkout, but callers must migrate from `changeProduct` to `purchase` + package selection.
> 
> **Overview**
> **Product changes now go through `Purchases.purchase()`** instead of a separate `changeProduct` API. Set **`productChangeInfo`** on purchase params (subscription id, optional per-call subscriber token); the destination product comes from **`rcPackage.webBillingProduct.identifier`**, matching normal checkout and other SDKs.
> 
> **Subscriber tokens** can be set on **`Purchases.configure({ subscriberToken })`** or overridden per call via **`productChangeInfo.subscriberToken`**.
> 
> Successful product-change flows return a standard **`PurchaseResult`** (refreshed **`customerInfo`**, synthetic **`storeTransaction`**, **`redemptionInfo: null`**) plus an internal **`productChange.changeType`** (`immediate` | `deferred`). **`ChangeProductParams`** is replaced by **`ProductChangeInfo`**; **`changeProduct`** is removed.
> 
> The webbilling demo upgrade page picks a package from the offering and calls **`purchase`** with **`productChangeInfo`**; tests cover validation and token precedence.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fb983c1849f1425287eaa8e77780c167d7490f16. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/main.ts
```diff
@@ -2226,10 +2256,27 @@ export class Purchases {
...
+              productIdentifier: result.newProductId,
+              purchaseDate: new Date(),
+            },
+            productChange: {
+              changeType: result.changeType,
+            },
+          });
+        } catch (error) {
+          reject(error);
+        }
```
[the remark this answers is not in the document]
james-od: I'm going to roll this into https://linear.app/revenuecat/issue/WEB-4540/sdk-refactor-reusable-method-for-backclose-behaviour, and move that ticket from backlog into the project milestone. I think reuse is probably the easiest way to make sure this all works as expected, but I'd rather handle that refactor separately.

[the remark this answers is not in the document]
james-od: https://github.com/RevenueCat/purchases-js/pull/1008/commits/0bf89f5cea6de15d3abf1ef608bae0986e8e8938

src/entities/product-change-params.ts
```diff
@@ -25,14 +13,17 @@ export interface ChangeProductParams {
    */
   subscriptionId: string;
```
tonidero, line 14: Not directly related to this, but I wonder if it would be possible to just pass the old subscription/product id, as it appears in the CustomerInfo. FWIW, this is what we do in the mobile SDKs. I would only do this though with `subscriberToken` which already needs a backend authenticated operation to perform the operation though... (Maybe I'm missing some other reason why this should be a value provided by the backend, if so ignore me!)

Also, if we did this change, we would need to add a verification that the customer does indeed have that subscription before proceeding with the change.

src/main.ts
```diff
@@ -2226,10 +2256,27 @@ export class Purchases {
         win.addEventListener("popstate", onClose as EventListener);
       }
 
-      const onFinished = (result: ProductChangeResult) => {
+      const onFinished = async (result: ProductChangeResult) => {
         this.inMemoryCache.invalidateAllCaches();
         unmountUi();
-        resolve(result);
+        try {
+          const customerInfo = await this.getCustomerInfo();
```
tonidero, line 2263: I wonder if we should include the new customer info in the change product operation result so we avoid an extra request and avoid any possible race conditions... But not a blocker for this.

tonidero (approved):
Looks great to me! I would love if we could also change the `subscriptionId` with the old product id, so it's easier to use from hybrids. But maybe we can iterate on that on follow-up PRs since it's internal...

src/entities/product-change-params.ts
james-od: Yeah the problem is that in RC Billing we can have multiple active subscriptions for an app user, so `CustomerInfo` could return multiple and we would need some way to decide.
In iOS you can only have one subscription per subscription group so the source product is always deterministic. Though in android for the paywall path it looks like we just take the first if there are multiple: https://github.com/RevenueCat/purchases-android/blob/c622d81365fdb3f578a157968d6f17436acf7565/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/data/ProductChangeCalculator.kt#L20-L27.
So we could adopt some heuristic to always be able to pick a value, but I think ideally there would still be a way for the developer to set it too - which currently would be the `subscriptionId`. 
But I think you're right that the problem fundamentally is more that this existing approach means a call to API v2 and not via the SDK.
Intuitively the two ways to solve it are:
1. Get the `productId` from `CustomerInfo` and pass to the productChange endpoint
2. Return subscriptionIds in `CustomerInfo` (we don't currently -`activeSubscriptions` is `productId`s, not `subscriptionId`s).

After looking into it, it looks like we can reliably use `productId`s here as the source for the change because an app user can't have two active subscriptions to the same product. 
So I think we can use #1 above, have the SDK get the productIds and pick one in some manner - but also still allow the developer to pass an override productId which again they can get from `CustomerInfo` without needing a separate fetch.

I can work on this as a follow up if you think it's sensible? @tonidero

src/main.ts
james-od: Will look into this as part of the above response :)

src/entities/product-change-params.ts
tonidero: That makes sense to me. The important part there is that we only allow one single subscription per productId in web/stripe/paddle so if that's true, I think it would indeed make the API easier to use and it would be closer to what we do in the mobile SDKs 🙏 

One side note, in Android, Google actually doesn't allow multiple subscriptions to the same product Id for the same user either, so I'm unsure on why we have that comment... cc @vegaro in case you remember why we added that?

james-od: Hmm yeah maybe doesn't hold for Stripe - at least as is. Will talk to the team about that, see if theres something we can do.
As an aside in that block^ for Android they aren't necessarily keyed to the same product iiuc, it's checking for multiple subscriptions, so I think it makes sense.

PR #1018, opened 2026-08-03
feat(vega): include/import vega IAP SDK

fire-at-will:
## Changes introduced
Adds the Amazon Vega IAP SDK as a dependency to Purchases JS.

When the SDK is configured with an Amazon API key, it starts loading the Vega IAP SDK immediately without changing the synchronous Purchases.configure() API. The loader caches the promise, logs when loading succeeds, and turns failures into a configuration error through the Amazon billing wrapper.

The diff on this PR looks huge, but almost all of it is due to lockfile changes, not actual code changes.

### Packaging
The Vega IAP SDK is a pinned build dependency of Purchases JS, matching the existing Stripe/Paddle packaging model. Apps do not install the Vega IAP SDK themselves. The IAP SDK is emitted as a lazy bundle chunk: non-Amazon web apps ship the asset but do not load or execute it, and Amazon apps load it on demand when `configure()` is called.

The IAP SDK depends on Kepler’s native React Native runtime, whose entry point cannot be parsed by browser bundlers. A build-time shim defers that native lookup until the IAP SDK is used in a Kepler runtime. This prevents ordinary Vite web builds from resolving the native package.

### Purchase Tester
The Purchase Tester no longer declares the Vega IAP SDK directly. Instead, it receives the bundled SDK through `purchases-js`. Its existing Kepler runtime dependency remains part of the Amazon app setup.

---

> [!NOTE]
> **Medium Risk**
> Changes purchase packaging and configuration-time loading for Amazon store keys, but the SDK stays lazy and gated on API key type, limiting blast radius for typical web billing apps.
> 
> **Overview**
> **Purchases JS now ships the Amazon Vega IAP SDK** (pinned dev/build dependency, same packaging idea as Stripe/Paddle). Apps do not add the IAP package themselves; the demo **Purchase Tester** drops its direct `@amazon-devices/keplerscript-appstore-iap-lib` dependency and relies on `purchases-js`.
> 
> For **Amazon API keys**, `configure()` still returns synchronously but **starts background loading** of the Vega SDK via `AmazonVegaBillingWrapper` and a **cached dynamic import**. Success is logged; failures surface as **configuration errors** through the wrapper. Non-Amazon keys never import the SDK.
> 
> **Web bundling** aliases Kepler’s `react-native-kepler` to a **runtime shim** so ordinary Vite builds do not parse the native entry; Kepler/Metro resolves the real module when the IAP chunk loads. The internal API report adds `amazonVegaBillingWrapper` on `Purchases`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2e9a1521252b0b741bad4326e49043f60938037a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/amazon-vega/amazon-vega-sdk-loader.ts
+++ b/src/amazon-vega/amazon-vega-sdk-loader.ts
@@ -1,0 +1,44 @@
+import type * as AmazonVegaSdkModule from "@amazon-devices/keplerscript-appstore-iap-lib";
+
+/**
+ * The Vega SDK should only be available when configured with an Amazon API Key.
+ * This loader manages importing the SDK at runtime so web and other store
+ * configurations never resolve it.
+ * @internal
+ */
+export type AmazonVegaSdk = typeof AmazonVegaSdkModule;
+
+type AmazonVegaSdkImporter = () => Promise<AmazonVegaSdk>;
+
+const importAmazonVegaSdk: AmazonVegaSdkImporter = async () =>
+  await import("@amazon-devices/keplerscript-appstore-iap-lib");
+
+/** @internal */
+export interface AmazonVegaSdkLoader {
+  load(): Promise<AmazonVegaSdk>;
+}
+
+/** @internal */
+export function createAmazonVegaSdkLoader(
+  importer: AmazonVegaSdkImporter = importAmazonVegaSdk,
+): AmazonVegaSdkLoader {
+  let sdkPromise: Promise<AmazonVegaSdk> | undefined;
+
+  return {
+    load(): Promise<AmazonVegaSdk> {
+      sdkPromise ??= importer();
+      return sdkPromise;
+    },
+  };
+}
+
+const amazonVegaSdkLoader = createAmazonVegaSdkLoader();
+
+/**
+ * Starts loading the Vega SDK and returns the cached loading promise.
+ * Future Amazon purchase handling can await this same promise.
+ * @internal
+ */
+export function loadAmazonVegaSdk(): Promise<AmazonVegaSdk> {
+  return amazonVegaSdkLoader.load();
+}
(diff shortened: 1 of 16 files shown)


src/amazon-vega/amazon-vega-sdk-loader.ts
```diff
@@ -0,0 +1,24 @@
...
+ * This loader manages importing the SDK at runtime so web and other store
+ * configurations never resolve it.
+ * @internal
+ */
+export type AmazonVegaSdk = typeof AmazonVegaSdkModule;
+
+type AmazonVegaSdkImporter = () => Promise<AmazonVegaSdk>;
+
+const importAmazonVegaSdk: AmazonVegaSdkImporter = async () =>
+  await import("@amazon-devices/keplerscript-appstore-iap-lib");
```
tonidero, line 14: So dynamic imports are a pain later in hybrids... We've had several issues before with them and we try to avoid them... Mostly the issues are around the fact that we create a umd file for the library to be used in flutter and that we have [this setting](https://github.com/RevenueCat/purchases-hybrid-common/blob/615b298dd490d14d0aa62385eba5ecdae808d78b/purchases-js-hybrid-mappings/rollup.config.js#L14) in `purchases-js-hybrid-mappings`. 

Having said this, I don't think we need to support Vega OS from hybrids? If so, maybe there could be a way to not include this in `purchases-js-hybrid-mappings` somehow? (Not sure if it's the best solution TBH, but something that came to mind)

tonidero (commented):
On my side, I'm mostly a bit concerned about how that dynamic import would work on hybrids... Also, I think it would be great to get a review from the web team on this one 🙏

src/amazon-vega/amazon-vega-sdk-loader.ts
fire-at-will: Yeah, right now, we don't have any solid plans to support Vega OS from the hybrids. If we ever did, we'd only support it on React Native, since Vega OS apps are built in React Native. I'll do some experimenting here...

PR #1019, opened 2026-08-04
[WEB-4548] Use product_id over sub_id for product changes

james-od:
## Motivation / Description

See the ticket: https://linear.app/revenuecat/issue/WEB-4548/sdkkhepri-key-product-change-on-productid-over-subscriptionid

Essentially, we think it makes more sense to key the product change on the product id and not the subscription id. This is feasible because an app user should not be subscribed to the same product more than once (though it's theoretically possible in Stripe billing, so we carve out explicit handling for it).

Additionally, we now infer the source subscription if no source product id is provided - though error if 0 or >1 are found for the app user.

The demo remains built in a pretty client side manner, but the product id can still be fetched from API v2 and handled via a backend if preferred.

https://www.loom.com/share/deccb210702541ba8eef05713311bf9a

## Additional comments

---

> [!NOTE]
> **Medium Risk**
> This is a breaking API change for experimental `productChangeInfo` (removes `subscriptionId`) and shifts source resolution to product id plus optional backend inference, which can mis-target changes if callers omit `fromProductIdentifier` when multiple Web Billing subscriptions exist.
> 
> **Overview**
> Product change checkout now identifies the **source** subscription with an optional **`fromProductIdentifier`** (from `CustomerInfo`) instead of a required RevenueCat **`subscriptionId` (`sub…`)**. The change/start API body sends **`from_product_id`** when set and omits it so the backend can infer the source when the user has exactly one active Web Billing subscription.
> 
> **`Purchases.purchase`** with **`productChangeInfo`** no longer validates **`subscriptionId`**; **`subscriberToken`** is still required. **`UpgradeCheckoutUi`**, **`ProductChangeOperationHelper`**, and **`Backend.postSubscriptionChangeCheckoutStart`** are aligned on the new parameter.
> 
> The **webbilling-demo** upgrade page drops the subscription id field, lists active **`rc_billing`** products for “change from,” and documents inference vs explicit selection when multiple subscriptions exist. Tests and Storybook props are updated accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ee247b9a17ac14dfc641daceaea6d80dd3296965. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/entities/product-change-params.ts
+++ b/src/entities/product-change-params.ts
@@ -5,13 +5,18 @@
  */
 export interface ProductChangeInfo {
   /**
-   * RevenueCat subscription public id (`sub…`) of the subscription to change.
+   * Product identifier of the Web Billing subscription to change.
    *
-   * Typically obtained by your backend via the
-   * [Developer API](https://www.revenuecat.com/docs/api-v2)
-   * customer subscriptions list and passed to the client with the token.
+   * Obtainable from {@link CustomerInfo.activeSubscriptions} or
+   * {@link CustomerInfo.subscriptionsByProductIdentifier} via
+   * {@link Purchases.getCustomerInfo}.
+   *
+   * When omitted, the source will be inferred on the assumption that the app
+   * user has exactly one active Web Billing subscription.
+   * However if it is possible for an app user to have zero or multiple
+   * active Web Billing subscriptions then this field is required.
    */
-  subscriptionId: string;
+  fromProductIdentifier?: string;
   /**
    * Optional short-lived subscriber access token override for this call.
    * Falls back to {@link PurchasesConfig.subscriberToken} when omitted.
--- a/examples/webbilling-demo/README.md
+++ b/examples/webbilling-demo/README.md
@@ -49,8 +49,9 @@ It uses a small token server to serve as a backend: it holds a **secret** API ke
 2. Start the token server: `pnpm run token-server`
 3. In another terminal, start the demo: `pnpm run dev` (Vite proxies `/api` → the token server)
 4. Open `/upgrade/<app_user_id>` for a customer with an active Web Billing subscription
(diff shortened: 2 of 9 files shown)


src/entities/product-change-params.ts
```diff
@@ -5,13 +5,18 @@
...
-   * Typically obtained by your backend via the
-   * [Developer API](https://www.revenuecat.com/docs/api-v2)
-   * customer subscriptions list and passed to the client with the token.
+   * Obtainable from {@link CustomerInfo.activeSubscriptions} or
+   * {@link CustomerInfo.subscriptionsByProductIdentifier} via
+   * {@link Purchases.getCustomerInfo}.
+   *
+   * When omitted, the source will be inferred on the assumption that the app
+   * user has exactly one active Web Billing subscription.
+   * However if it is possible for an app user to have zero or multiple
```
tonidero, line 16: I would say having zero active subscriptions would be VERY common 😅. So we should just make sure we handle it nicely. Probably just fail with a meaningful error when this happens.

tonidero (approved):
Like this! This will make the eventual integration in the hybrids much easier :)

nicfix (approved):

src/entities/product-change-params.ts
james-od: Very common but hopefully they don't find themselves in an upgrade path in that case 😅 
The next step I'm going to take is to roll both '/start' endpoints we have in Khepri (I.e we have one to 'start' a normal checkout and one to 'start' an upgrade checkout) into a single endpoint, which will then decide based on the params whether to perform an upgrade or a regular checkout. So if they have zero then we should just perform a regular checkout. Maybe you're right it's better to just not mention the zero case here.

PR #1025, opened 2026-08-05
[WEB-4567] Support subId and/or prodId for subscription changes

james-od:
## Motivation / Description

This PR replaces the previous https://github.com/RevenueCat/purchases-js/pull/1019.

Linked with https://github.com/RevenueCat/khepri/pull/23583.

The goal here is to support subscription changes keyed by either a product ID _or_ a subscription ID.

Generally, we would recommend the subscription ID as it's ultimately the identifier of the subscription that will actually be _changed_. The product ID provided is mapped to a subscription ID, and this isn't necessarily in all possible cases 1:1 (mostly thinking ahead to Stripe support). However in the majority of cases it's likely good enough, eases development for users preferring to use primarily the SDK, and should make support in hybrids easier.

Additionally, this PR also introduces inference if _neither_ is provided, where we use the app_user_id to see if they only have one active relevant subscription - and if so we'll use that.

If both are provided, we just assert that they need to match.

Updates the demo app UI accordingly to support either input.

The video below shows the example demo UI, in which there are inputs for product id and subscription id. Subscription id is currently fetched manually, but we plan to return this alongside the auth token in future which we could then apply in this demo to provide a dropdown similar to what we do with product id. 
The appuser33 app_user_id is subscribed to SubChangesYearlyBasic and the video shows some sample (arbitrary) upgrades/downgrades from that position, given various ids explicitly provided or omitted.

https://github.com/user-attachments/assets/76e8d85c-47a6-4259-8a0b-13f6251e3372

## Additional comments

---

> [!NOTE]
> **Medium Risk**
> Changes how the upgrade flow identifies the source subscription and relies on backend inference when ids are omitted, which can mis-target users with multiple active Web Billing subscriptions if callers omit ids incorrectly.
> 
> **Overview**
> Upgrade checkout no longer **requires** `subscriptionId` in `productChangeInfo`. Callers can pass **`productIdentifier`** (from customer info), **`subscriptionId`**, **both** (must match the same subscription), or **neither** so the backend infers a single active Web Billing subscription.
> 
> The SDK forwards optional ids through `ProductChangeOperationHelper` and `postSubscriptionChangeCheckoutStart`, which now sends `from_product_id` when a product id is set and omits both fields when inference is intended. Client validation that blocked missing `subscriptionId` was removed; **`subscriberToken`** is still required.
> 
> The webbilling demo adds a **change from product** dropdown (active `rc_billing` subs) and treats subscription id as optional, with README steps updated accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3fef92eb08efdf0c3f0a40deb4b314811a2029e7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/examples/webbilling-demo/README.md
+++ b/examples/webbilling-demo/README.md
@@ -49,8 +49,9 @@ It uses a small token server to serve as a backend: it holds a **secret** API ke
 2. Start the token server: `pnpm run token-server`
 3. In another terminal, start the demo: `pnpm run dev` (Vite proxies `/api` → the token server)
 4. Open `/upgrade/<app_user_id>` for a customer with an active Web Billing subscription
-5. Pick the target package from the current offering and enter the RevenueCat subscription public id (`sub…`)
-6. Click **Open upgrade checkout**. The SDK starts a checkout session, shows from→to / pricing / PM on file, then confirms on CTA.
+5. Optionally pick a source product from active Web Billing subscriptions and/or enter a subscription public id (`sub…`). Either, both (must match), or neither (we infer a single active Web Billing subscription) is allowed. Provide at least one id if the user may have multiple active Web Billing subscriptions.
+6. Pick the target package from the current offering.
+7. Click **Open upgrade checkout**. The SDK starts a checkout session, shows from→to / pricing / PM on file, then confirms on CTA.
 
 ### Payment Methods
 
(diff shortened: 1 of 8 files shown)


nicfix (approved):
Looks great!

PR #1029, opened 2026-08-06
[WEB-4571] Fold sub-change checkout into /start and /complete

james-od:
## Motivation / Description

The idea is to combine the /start route for the normal checkout and the /start route for the upgrade checkout into a single path. And then from the SDK we just call /start optionally with auth token + source prod/sub id. If that object is passed, attempt an upgrade response (using the auth), if not or if the upgrade isn't possible, return a normal checkout response. Similarly for /complete.
This means that we can fully offload the decision of 'which checkout to show' to Khepri and keep the SDK side clean(er).
We do have to do a bit of managing the different request/response shapes - but thats now the bulk of the complexity.
We also return an explicit param denoting the intended checkout type.

It's worth noting that the behaviour built here is "Try to perform the upgrade if at all possible- but if anything is missing, just perform a normal checkout".

The other potentially contentious point here is auth. I've opted to add a new X-RC-Subscriber-Token header. The problem with the auth is that the Authorization header we use for RC Billing can only hold one type of auth (the public API key). I think this broadly similar to how we use obo_token. Happy to iterate here though if it looks off.

I've rolled the upgrade checkout into a new 'page' and now the ui will show that page if we have the right response shape. Imo this a lot better than the previous approach where it was fully dedicated. It avoids having any back and forth when rendering, which was complicating some earlier attempts.

I'll follow this SDK change up with an /examples page that shows how a paywall can now be used to trigger an upgrade checkout.

Everything here is still marked @internal. 

In the recording below we show appUser35 be subscribed to YearlyBasic, and then begin some purchases/upgrades of other products. Upgrades trigger the upgrade checkout, if no upgrade is applicable then the regular checkout is shown.
Then appUser36 has no subscriptions so all resolve as fresh purchases.

https://github.com/user-attachments/assets/99a7b10a-b852-4375-9068-a48c1a78cd61

## Additional comments

---

> [!NOTE]
> **Medium Risk**
> Touches the core Web Billing purchase path, subscription-change completion, and new subscriber-token auth alongside the API key; behavior change when upgrade isn't possible (fallback vs dedicated flow) needs careful backend/SDK alignment.
> 
> **Overview**
> **Subscription upgrades/downgrades now ride the same checkout session as new purchases**, instead of a separate upgrade UI and dedicated `/subscription/change/checkout/*` APIs.
> 
> When `productChangeInfo` and a subscriber token are present (Web Billing only), `purchase` / paywall checkout calls unified **`/checkout/start`** with optional `product_change` (`subscription_id` / `from_product_id`) and **`X-RC-Subscriber-Token`**. The backend returns either `checkout_mode: "subscription_change"` (upgrade-confirm page inside the standard checkout shell) or `checkout_mode: "purchase"` (normal payment flow in the **same** session—no second start). Confirmation uses **`/checkout/{session}/complete`** with the subscriber token header instead of a separate confirm endpoint.
> 
> `ProductChangeOperationHelper`, `UpgradeCheckoutUi`, and the old networking endpoints are removed; logic lives in `PurchaseOperationHelper` (`completeProductChange`, response guards). Missing subscriber token no longer errors—it **falls back** to a normal purchase. Responses are distinguished via `checkout_mode` and type guards.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dfd0c27b94a0df3122281431eb1d7a90e7a626c1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/networking/responses/subscription-change-response.ts
```diff
@@ -41,10 +41,34 @@ export interface SubscriptionChangeCheckoutStartResponse {
...
+export function isSubscriptionChangeCompleteResponse(
+  response: unknown,
+): response is SubscriptionChangeCompleteResponse {
+  return (
+    typeof response === "object" &&
+    response !== null &&
+    "checkout_mode" in response &&
+    response.checkout_mode === "subscription_change"
+  );
 }
```
lburdock: These two functions are identical except for the type they assert. Should each one also check a field unique to the shape it's narrowing to, like `to_product` for start and `new_product_id` for complete?

src/networking/responses/subscription-change-response.ts
```diff
@@ -41,10 +41,34 @@ export interface SubscriptionChangeCheckoutStartResponse {
...
   billing_address: SubscriptionChangeBillingAddressSummary | null;
+  checkout_mode?: "subscription_change";
 }
 
-export interface SubscriptionChangeConfirmResponse {
+export interface SubscriptionChangeCompleteResponse {
   operation_session_id: string;
   change_type: "immediate" | "deferred";
   new_product_id: string;
+  checkout_mode?: "subscription_change";
```
lburdock: Is `checkout_mode` optional? Could it just be `checkout_mode: "subscription_changes"` to make the response easier to distinguish, and then `CheckoutCompleteResponse` could use `checkout_mode: "purchase"`?

lburdock (approved):

src/networking/responses/subscription-change-response.ts
james-od: Yeah good point! https://github.com/RevenueCat/purchases-js/pull/1029/changes/26cb85b6ee384a9b47397568d576d4d944b2cbd6

src/networking/responses/subscription-change-response.ts
james-od: Ah yeah I think we can get away with firming this up! https://github.com/RevenueCat/purchases-js/pull/1029/changes/d0ca00726d095ac2842cea53a9cd4e1901dd8934

PR #1043, opened 2026-08-12
feat(stripe): send supports_price_id capability on product fetches (WST-932)

rogersole:
## Summary

When a customer views a web paywall in a non-default currency, the backend will soon be able to serve the Stripe price that matches that currency instead of always serving the base one. To do that safely, the backend needs to know the client can handle it: an SDK that displays a substituted price must also tell checkout exactly which price it displayed, otherwise the customer could be charged a different amount or currency than what they saw.

This PR makes purchases-js declare that capability. Every product fetch now includes `supports_price_id=true`, which tells the backend it's safe to substitute currencies because this SDK version always echoes the returned `price_id` back at checkout. The echo itself already existed — this change is the opt-in signal plus tests that pin both sides of the contract.

Old SDK versions never send the flag, so they keep getting the base price and remain unaffected.

## Changes

- `GetProductsEndpoint.urlPath()` appends `&supports_price_id=true` (literal string) to `GET /rcbilling/v1/subscribers/{id}/products`. The param lives in `urlPath()` so the 2000-char URL batching in `backend.getProducts` accounts for it automatically.
- Verified the `price_id` echo to `POST /rcbilling/v1/checkout/start`: already implemented (`price_id: purchaseOption.priceId` in `backend.ts`, same for `/checkout/prepare`) — no code change needed. Added a body assertion on the `/checkout/prepare` test so both echo sites are now pinned.
- `currency` in the products URL is now passed through `encodeURIComponent` like every other query param in `urlPath()` (no behavior change for valid currency codes).

Notes:
- The ticket also mentions `GET .../offering_products`, but purchases-js never calls that endpoint (verified repo-wide) — that half of the contract applies to khepri's other clients.
- Deployment order: khepri ([WST-930](https://linear.app/revenuecat/issue/WST-930/backend-pr-5-currency-aware-serving-checkout) / khepri#23838) ships first; until this change ships, khepri serves the base currency to all clients.

Linear: [WST-932](https://linear.app/revenuecat/issue/WST-932/purchases-js-send-supports-price-id-capability-and-echo-price-id-at)

## Testing

- Updated the six exact-URL assertions in `endpoints.test.ts` (the wire-format spec for `urlPath()`).
- The shared `getProducts` msw helper in `backend.test.ts` now rejects any request missing `supports_price_id=true`, so every getProducts test enforces the flag. While there, fixed its discount check, which read `discountCode` instead of the emitted `discount_code` and so never validated anything.
- Placement tests in `purchase.offerings.test.ts` build expected URLs from `GetProductsEndpoint.urlPath()` instead of hardcoding query shape, so future param changes don't churn tests unrelated to URL format.
- `pnpm test` (835 tests), `prettier`, `eslint`, `tsc` (src + tests) all pass.

---

> [!NOTE]
> **Medium Risk**
> Changes the products API contract and checkout price alignment path; misalignment could show one price and charge another, though older clients without the flag are unchanged.
> 
> **Overview**
> Product fetches now advertise **`supports_price_id=true`**, telling the backend this SDK can echo the returned Stripe **`price_id`** at checkout so currency-substituted prices stay consistent with what the paywall shows.
> 
> **`GetProductsEndpoint`** appends that query flag on every products URL and runs **`currency`** through **`encodeURIComponent`** like other params. Tests pin the wire format, require the flag in MSW **`getProducts`** mocks (and fix discount validation to use **`discount_code`**), assert **`POST /checkout/prepare`** sends **`price_id`**, and build placement expected URLs from **`GetProductsEndpoint.urlPath()`** so future query changes don’t drift.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a676313b871eb22510ae464e4b3cb70296c9af3e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


rogersole:
@cursor review

nicfix (changes requested):
Why do we need this change in the sdk?

If it's for some backwards compatibility from before having multi currency in the SDK, we didn't even have Stripe Billing before multi currency.

PR #1045, opened 2026-08-12
Block multiple purchase inits

james-od:
## Motivation / Description

Noticed this while testing the upgrade mode checkout, but it affects the regular checkout too.
Currently, clicking the CTA on a paywall multiple times in quick succession will trigger checkout `start` calls for each click. Khepri will return `EQUIVALENT_REQUEST_IN_FLIGHT` 409s (in both scenarios once https://github.com/RevenueCat/khepri/pull/23865 is merged) but these are displayed as errors to an end user depending on the return order. Clicking 'back' solves the problem but this is obviously not ideal.

https://github.com/user-attachments/assets/d24e0ffd-5048-4466-9d23-5bb439122f9e

https://github.com/user-attachments/assets/50e399a7-b7c2-4d84-9077-025550668b59

We now use `purchaseInFlight` to block this scenario:

https://github.com/user-attachments/assets/3222ba6e-92db-4113-a416-c2eb25d7d190

(I promise that in all the videos I'm clicking the CTA rapidly)

## Additional comments

---

> [!NOTE]
> **Low Risk**
> Localized guard around paywall purchase initiation with tests; no auth or billing API contract changes beyond reducing duplicate starts.
> 
> **Overview**
> **Prevents rapid paywall CTA clicks from starting multiple checkout sessions**, which could trigger backend `EQUIVALENT_REQUEST_IN_FLIGHT` responses and surface confusing errors.
> 
> `presentPaywall` now tracks a `purchaseInFlight` flag and routes both standard **Paywall** and **Workflow** purchase callbacks through `createPurchaseClickHandler`. Extra clicks are ignored until the current `startPurchaseFlow` attempt settles (success, error, or cancel) in a `finally` block. Workflow and standard paywall paths share the same handler so locale is passed consistently to checkout.
> 
> Tests cover that only one `purchase` / `onPurchaseStarted` fires during rapid clicks, and that a retry is allowed after a failed attempt completes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2f9603b3262f8fa623d58bd46dd50dd9ba3bdf23. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/main.ts
```diff
@@ -1239,15 +1262,7 @@ export class Purchases {
...
-            onPurchaseClicked: (selectedPackageId: string) => {
-              const pkg = offering.packagesById[selectedPackageId];
-              if (pkg) {
-                notifyPurchaseStarted(pkg);
-              }
-              startPurchaseFlow(selectedPackageId)
-                .then(onSuccess)
-                .catch(onError("Error performing purchase"));
-            },
+            onPurchaseClicked: handlePurchaseClicked,
```
lburdock: Should this be:
```suggestion
            onPurchaseClicked: (selectedPackageId: string) =>
              handlePurchaseClicked(selectedPackageId),
```

It looks like onPurchaseClicked in purchases-ui-js has the type `onPurchaseClicked?: (selectedPackageId: string, actionId: string) => void;` which means `onPurchaseClicked: handlePurchaseClicked` will lead the actionId to be passed in as the locale.

lburdock (approved):

src/main.ts
james-od: Very good spot! I reworked it so createPurchaseClickHandler(locale) closes over the locale and returns a (selectedPackageId) => handler, used for both. So the click handler only ever takes a package id

PR #1048, opened 2026-08-12
feat(vega): support purchasing Amazon products

fire-at-will:
## Motivation / Description
Adds the ability to purchases packages using the Amazon Store on the Vega OS.

## Changes introduced
- Adds the ability to purchases packages using the Amazon Store on the Vega OS. When a purchase occurs, the SDK:
   1. Purchases the package with the Amazon IAP SDK
   2. Posts the receipt to the RevenueCat backend
   3. Notifies Amazon of the purchase with the `notifyFulfillment()` function
- Removes the `WebBillingHandler` abstraction, since moving web billing purchase functionality from `main.ts` -> `WebBillingHandler` turned out to be a pretty large refactor. We can revisit in the future if we want to.
- Adds a "Purchase" button next to products in offerings in the Vega Purchase Tester app.
- Adds unit tests

When posting receipts, the SDK posts the following for the product ID:
- Entitlements/consumables: `receipt.sku`
- Subscriptions: `receipt.termSku`
This is a big improvement over the implementation in the Android SDK, where the SDK must query the RevenueCat backend to obtain the `termSku` before posting the receipt to the RevenueCat backend.

I was able to verify that purchases go through successfully end-to-end with the RevenueCat backend!

---

> [!NOTE]
> **High Risk**
> Touches the purchase and receipt-posting path (entitlements and store fulfillment). Amazon-only, but incorrect SKU/receipt handling could grant or fail purchases.
> 
> **Overview**
> Adds Amazon Store purchases on Vega: `purchase()` with an `amzn_` API key now goes through `AmazonBillingWrapper`, which buys via the Amazon IAP SDK, posts the receipt to RevenueCat (including `store_user_id`), then calls `notifyFulfillment`. Fulfillment failures are logged and do not fail the purchase.
> 
> Subscriptions post `receipt.termSku`; consumables/entitlements post `receipt.sku`. Amazon response codes map to existing `PurchasesError` codes. Receipt ID is returned as `operationSessionId`.
> 
> Removes the unused `WebBillingWrapper` passthrough. Product fetches for Amazon keys still use the Amazon wrapper; other keys call `backend.getProducts` directly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f767e7d7f3e9aa1401c264e8ae1d57f0c7ff5402. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


fire-at-will:
@claude please review

src/amazon/amazon-billing-wrapper.ts
```diff
@@ -42,6 +47,123 @@ export class AmazonBillingWrapper implements BillingWrapper {
...
+
+    switch (response.responseCode) {
+      case NotifyFulfillmentResponseCode.SUCCESSFUL:
+        Logger.debugLog(
+          `Successfully fulfilled receipt ID ${receiptId} with the Amazon Store.`,
+        );
+        break;
+      case NotifyFulfillmentResponseCode.NOT_SUPPORTED:
+        Logger.warnLog(
+          `Failed to fulfill receipt ID ${receiptId} with the Amazon Store: fulfillment is not supported.`,
```
fire-at-will, line 162: Fulfillment is "fire and forget" in the Android SDK, so I kept it that way here too instead of propagating an error.

src/main.ts
```diff
@@ -1677,6 +1680,14 @@ export class Purchases {
       return await this.performStripePurchase(params);
     }
 
+    const isAmazon = isAmazonApiKey(this._API_KEY);
+    if (isAmazon) {
+      return await this.unwrappedAmazonBillingWrapper().purchase(
+        params,
+        this._appUserId,
+      );
+    }
```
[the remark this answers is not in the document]
fire-at-will: addressed in 1c03bdfa4f9d1c15ece4cf8fc9d5a6bdcc6d0771

src/amazon/amazon-billing-wrapper.ts
```diff
@@ -42,6 +47,123 @@ export class AmazonBillingWrapper implements BillingWrapper {
...
+
+    return {
+      customerInfo: toCustomerInfo(subscriberResponse),
+      redemptionInfo: null,
+      operationSessionId: receipt.receiptId,
+      storeTransaction: {
+        storeTransactionId: receipt.receiptId,
+        productIdentifier: receipt.sku,
+        purchaseDate: receipt.purchaseDate,
+      },
```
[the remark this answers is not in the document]
fire-at-will: addressed in 459420ed6a7a2a604873f15ba606573e2de0bdae

src/amazon/amazon-billing-wrapper.ts
```diff
@@ -42,6 +47,134 @@ export class AmazonBillingWrapper implements BillingWrapper {
...
+      Logger.warnLog(
+        `Failed to fulfill receipt ID ${receipt.receiptId} with the Amazon Store: ${String(error)}`,
+      );
+    }
+    Logger.debugLog("Amazon purchase completed successfully.");
+
+    return {
+      customerInfo: toCustomerInfo(subscriberResponse),
+      redemptionInfo: null,
+      operationSessionId: receipt.receiptId,
```
tonidero, line 129: Hmm not sure if we should use the receiptId for the operationSessionId... but I'm not sure of a better idea TBH 😓

src/amazon/amazon-billing-wrapper.ts
tonidero: Hmm TBH, not sure if we should... what are the consequences of not completing the fullfillment? I guess if the backend did actually succeed, they should get the entitlement/updated customer info if any... But what I'm concerned is the lack of any retry mechanism right now... In the Android SDK, we would retry on the next app foreground... I don't think we have such a mechanism here in web? (not sure if we plan to add something like that later but it might be tricky since we don't cache things on the web SDK right now, though maybe we should consider it for this). I'm also not sure in what situations this might fail TBH... But yeah, I think this might be ok for now, as long as we have some retry mechanism in place.

src/amazon/amazon-billing-wrapper.ts
```diff
@@ -42,6 +47,134 @@ export class AmazonBillingWrapper implements BillingWrapper {
...
+    const subscriberResponse = await this.backend.postReceipt(
+      appUserId,
+      productIdentifier,
+      rcPackage.webBillingProduct.price.currency,
+      receipt.receiptId,
+      rcPackage.webBillingProduct.presentedOfferingContext,
+      "purchase",
+      undefined,
+      storeUserId,
+    );
```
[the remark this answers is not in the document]
tonidero: Hmm I guess we should support paywalls with the amazon SDK right? If so, this seems valid?

tonidero (approved):
I think the only thing that we might want to really fix is the missing paywallId propagation. Other than that, just some questions, but I think we can 🚢 this

src/amazon/amazon-billing-wrapper.ts
fire-at-will: The only options I could think of were to use the receiptId or to leave it as an empty string since I believe `operationSessionId` is a web billing concept 🤷‍♂️

src/amazon/amazon-billing-wrapper.ts
fire-at-will: Right. The Amazon docs don't appear to have many details on what notifying the fulfillment actually does, but I can find two things:
- For OTPs, the `getPurchaseUpdates()` function "retrieves only unfulfilled and canceled consumable purchases" (source: https://community.amazondeveloper.com/t/vega-iap-getpurchaseupdates/20231)
- It is documented for Quick Subscribe that Amazon automatically cancels subscriptions that aren't fulfilled. I can't find confirmation in the documentation about this behavior for non-Quick Subscribe subscriptions, but in my testing, I haven't been able to make this happen in the DevTest/LAT environments.

I think we should probably do some caching for retry logic, but that might be a bit complicated. The only way that I can find to cache data in a Vega app is with the `@amazon-devices/react-native-async-storage__async-storage` library (docs: https://developer.amazon.com/docs/vega-api/0.23/react-native-async-storage.html). This library is pre-installed on-device, but we still need to declare it in our package.json, so there could be some complications we'll need to make sure it's either installed as a peerDependency and dynamically loaded, or, if dynamically loading causes issues with the PHC JS mappings, we'll need to import it in a separate amazon module. 

Additionally, it seems like this dependency's version needs to be kept in sync with the version of React Native that is used by the app, according to this line from the docs:
> This library is system-deployed and available to React Native for Vega apps without a separate installation process. The library is deployed as an autolinking library, which your app links to at runtime. The library is guaranteed to be compatible only with the version of React Native for Vega for which it's built. When you uplevel your app's version of React Native for Vega, consider the best practice of upleveling its library dependencies.

So we may need to find a way to keep our version of this dependency in sync with the version of React Native used by apps 🤔

If it's okay with you, I think I'll explore this in a separate PR in this PR stack so that it can be scrutinized in a standalone PR :) At the very least, if the persistent caching proves to be too big of a headache, we could implement retries in the `notifyFulfillment` function and make the purchase function throw if it fails.

src/amazon/amazon-billing-wrapper.ts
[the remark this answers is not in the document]
fire-at-will: I don't think that they are supported unless.

Vega apps are built in React Native, so we can't use the re-use the existing mobile paywall engines here. We might be able to make them work by showing web paywalls in a web view in a Vega app, but IMO I think that's probably out of scope for this MVP implementation. Can you think of another way that they might be implemented easily here?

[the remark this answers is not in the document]
tonidero: Hmm right... it would indeed be a bit tricky since paywalls assumes the presence of a browser... So yeah, I think we can tackle this later 🙏. We indeed might be able to use a web view or something like that to present the paywall, but likely out of scope for this PR

src/amazon/amazon-billing-wrapper.ts
tonidero: Yeah... I also didn't have a better idea... Could be good to document it for now.

fire-at-will: Documented in 8bf44175baa2954b7cbc158f5729c5e726afdded 👍

fire-at-will:

PR #1054, opened 2026-08-14
WEB-4597 Allow runtime purchase appearance overrides

nicfix:
## Summary

Apps need to adapt checkout branding at runtime, for example when switching between light and dark presentation. Previously, `purchases-js` only used the Dashboard branding configuration and its internal appearance override was neither public nor sent to checkout start.

This change exposes a partial `brandingAppearanceOverride` through `Purchases.configure()`, `presentPaywall()`, and `purchase()`. Configure-time values become instance defaults, while values passed for a specific paywall or purchase take precedence. The resolved override also updates the SDK checkout UI without mutating cached branding and is forwarded to `/checkout/start` as `appearance_override`.

The override is intentionally per SDK instance or purchase and is not persisted through checkout sessions.

## Compatibility and rollout

- Existing integrations are unchanged when no override is supplied.
- Stripe's mobile wallet presentation can use the supported branding values, but the Stripe-hosted "Pay another way" fallback remains light and cannot be fully themed.
- End-to-end Stripe customization depends on Khepri accepting and applying `appearance_override` on `/rcbilling/v1/checkout/start`.

Closes WEB-4597.

---

> [!NOTE]
> **Medium Risk**
> Touches purchase/checkout start and all payment paths (Web Billing, Stripe, Paddle) with new API surface; behavior is backward-compatible when overrides are omitted, but incorrect merge or backend handling of `appearance_override` could affect checkout presentation.
> 
> **Overview**
> Adds public **`brandingAppearanceOverride`** (`Partial<BrandingAppearance>`) on **`Purchases.configure()`**, **`purchase()`**, and **`presentPaywall()`** so checkout can be themed at runtime without relying only on Dashboard branding.
> 
> **Precedence:** instance configure-time defaults merge with per-purchase/per-paywall overrides; operation-level fields win. Overrides are applied for checkout UI via merge helpers that **do not mutate** cached branding info, and the resolved partial override is sent to checkout start as **`appearance_override`**. **`presentPaywall`** forwards paywall overrides into the nested **`purchase()`** call.
> 
> **API shift:** `PurchaseParams.brandingAppearanceOverride` moves from an internal full override to this public partial model (with Stripe wallet theming notes in docs). **`DEFAULT_BRANDING_APPEARANCE`** backs merges when dashboard appearance is missing.
> 
> The webbilling demo adds a **Runtime appearance demo** route and loader path to compare default, configure-time, and per-operation palettes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b938960ff5b978762838ec914d4db823400a28e3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/examples/webbilling-demo/src/pages/appearance_overrides/index.tsx
+++ b/examples/webbilling-demo/src/pages/appearance_overrides/index.tsx
@@ -107,7 +107,9 @@ const AppearanceOverridesPage = () => {
       <LogoutButton />
 
       <header className="appearance-demo__header">
-        <span className="appearance-demo__eyebrow">WEB-4597 demo</span>
+        <span className="appearance-demo__eyebrow">
+          Appearance Override demo
+        </span>
         <h1>Runtime appearance overrides</h1>
         <p>
           Compare the Dashboard/default appearance with colors stored when
--- a/examples/webbilling-demo/src/pages/appearance_overrides/index.tsx
+++ b/examples/webbilling-demo/src/pages/appearance_overrides/index.tsx
@@ -1,0 +1,220 @@
+import type { Package } from "@revenuecat/purchases-js";
+import { useState } from "react";
+import LogoutButton from "../../components/LogoutButton";
+import { usePurchasesLoaderData } from "../../util/PurchasesLoader";
+import {
+  configuredAppearanceOverride,
+  operationAppearanceOverride,
+} from "../../util/runtime-appearance-overrides";
+
+type DemoStatus = {
+  kind: "idle" | "running" | "success" | "error";
+  message: string;
+};
+
+const getErrorMessage = (error: unknown) =>
+  error instanceof Error ? error.message : String(error);
+
+const AppearanceOverridesPage = () => {
+  const { purchases, offering } = usePurchasesLoaderData();
+  const packages =
+    offering?.availablePackages.filter((pkg) => pkg.webBillingProduct) ?? [];
+  const [selectedPackageId, setSelectedPackageId] = useState(
+    packages[0]?.identifier ?? "",
+  );
+  const [status, setStatus] = useState<DemoStatus>({
+    kind: "idle",
+    message: "Choose a flow to open checkout.",
+  });
+
+  const selectedPackage = packages.find(
+    (pkg) => pkg.identifier === selectedPackageId,
+  );
+  const isRunning = status.kind === "running";
+
+  const runFlow = async (label: string, operation: () => Promise<unknown>) => {
+    setStatus({ kind: "running", message: `Opening ${label}…` });
+
+    try {
+      await operation();
+      setStatus({ kind: "success", message: `${label} completed.` });
+    } catch (error) {
+      setStatus({
+        kind: "error",
+        message: `${label} closed or failed: ${getErrorMessage(error)}`,
+      });
+    }
+  };
+
+  const purchase = (pkg: Package, useOperationOverride: boolean) =>
+    runFlow(
+      useOperationOverride ? "per-purchase checkout" : "configured checkout",
+      () =>
+        purchases.purchase({
+          rcPackage: pkg,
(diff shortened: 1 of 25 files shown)


src/tests/networking/backend.test.ts
```diff
@@ -820,6 +820,35 @@ describe("postCheckoutStart request", () => {
     expect(result).toEqual(checkoutStartResponse);
   });
 
+  test("includes a partial appearance override when provided", async () => {
```
james-od, line 823: This covers passthrough but not specifically `partial` right? I think thats worth covering.

src/networking/backend.ts
```diff
@@ -384,6 +388,10 @@ export class Backend {
       requestBody.attribution_metadata = attributionMetadata;
     }
 
+    if (appearanceOverride) {
+      requestBody.appearance_override = appearanceOverride;
+    }
```
james-od, line 393: ~Doesn't this need to be added to the /start endpoint in Khepri first, otherwise will fail on `check_params`? Couldn't see an open PR~
Ah this is in the description, still would we not want to deploy that first?

src/main.ts
```diff
@@ -1973,11 +1996,10 @@ export class Purchases {
       ...(params.metadata || {}),
     };
 
-    const finalBrandingInfo: BrandingInfoResponse | null = this._brandingInfo;
-
-    if (finalBrandingInfo && params.brandingAppearanceOverride) {
-      finalBrandingInfo.appearance = params.brandingAppearanceOverride;
-    }
```
james-od, line 1987: Weird that `finalBrandingInfo` in this function was never actually used before? Its the same for the other functions above.

src/networking/backend.ts
nicfix: Yeah this shouldn't be here.

We don't need to get it from the backend, I'll remove it.

Sorry for not noticing this got added.

src/main.ts
nicfix: I think we never ended up using that config since we never split the WPL appearance from the sdk appearance.

And we never noticed that we were not using it since... ...we were not using it.

src/networking/backend.ts
nicfix: We actually need this for Stripe, I followed up with a khepri PR.

src/tests/networking/backend.test.ts
nicfix: It technically does given that all the appearance fields can be set (or not) in the appearance override.

But I know what you mean, in the end Stripe supports only the 2 passed.

examples/webbilling-demo/src/pages/appearance_overrides/index.tsx
```diff
@@ -0,0 +1,250 @@
...
+      </main>
+    );
+  }
+
+  return (
+    <main className="appearance-demo">
+      <LogoutButton />
+
+      <header className="appearance-demo__header">
+        <span className="appearance-demo__eyebrow">WEB-4597 demo</span>
```
james-od: Nit: can probably remove the ticket name here

james-od (approved):

src/main.ts
```diff
@@ -1655,9 +1670,21 @@ export class Purchases {
...
+      ? { ...params, brandingAppearanceOverride: appearanceOverride }
+      : params;
+    const effectiveBrandingInfo = applyBrandingAppearanceOverride(
+      this._brandingInfo,
+      appearanceOverride,
+    );
+
     if (isSimulatedStoreApiKey(this._API_KEY)) {
       const purchaseResult = await purchaseSimulatedStoreProduct(
-        params,
```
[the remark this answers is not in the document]
nicfix: It's fine the button cannot be styled freely but has to follow the Apple/Google guidelines.

PR #1071, opened 2026-08-20
feat(vega) sync/restore purchases enhancements

fire-at-will:
## Changes introduced
This PR introduces a few enhancements to the `syncPurchases()`/`restorePurchases()` functions for the Amazon Store that bring them more in-line with the functions' behavior on Android. Specifically:

### syncPurchases/restorePurchases
- Now post the receipts sorted by purchase date
- Post all receipts during `syncPurchases()`
- Cache successfully synced receipt IDs on device
- Perform fulfillment for restored receipts and cache IDs after successful fulfillment;
- Explicitly fetch CustomerInfo from the backend after processing all receipts to match the behavior found in the Android SDK.

### Caching of Processed Receipts
The SDK now caches successfully processed receipts using a new `VegaDeviceCache` class, which uses [@amazon-devices/kepler-file-system](https://developer.amazon.com/docs/vega-api/0.23/README.amazon-devices_kepler-file-system.html) to persist the cache on the fire stick's disk.
- Adds a dependency on `@amazon-devices/kepler-file-system`
   - It's a peer dependency in the main module itself
   - The Vega SDK entrypoint injects the actual dependency at runtime
- Cache files are scoped per API key
- Receipts are cached:
   - After successful purchase/post receipt/fulfillment
   - `syncPurchases()`: after each receipt is successfully posted to the backend. `syncPurchases()` does not call fulfillment
   - `restorePurchases()`: after each receipt is successfully posted to the backend and Amazon fulfillment is successful
- Unlike in the Android SDK, the receiptIds are not hashed using SHA-1, and are stored as raw strings. We could use the [kepler crypto library](https://developer.amazon.com/docs/vega-api/0.23/README.amazon-devices_keplercrypto.html), but:
   - This library is not supported in the simulator
   - The library doesn't support SHA-1, which is used by the Android SDK
- `@amazon-devices/kepler-file-system` doesn't support overwriting files, so when update a file, we delete the existing one and rewrite it. 
 
If we think that hashing the receipts is necessary, I can do that in a follow-up PR.

---

> [!NOTE]
> **Medium Risk**
> Touches Amazon purchase posting, fulfillment, and on-device persistence of receipt IDs. Scoped to the Vega entry point, but billing correctness and file-system cache behavior are easy to get wrong.
> 
> **Overview**
> Aligns Amazon Vega `syncPurchases` / `restorePurchases` with Android-style behavior: collect all purchase-update pages, post receipts oldest-first, then always refetch `CustomerInfo`. Restore now notifies Amazon fulfillment; sync posts every receipt each run instead of skipping in-memory duplicates.
> 
> Adds `VegaDeviceCache` (Kepler File System, optional peer `@amazon-devices/kepler-file-system`) wired only through the Vega entry point. Successfully posted (and, on restore/purchase, fulfilled) receipt IDs are written under `/data` per API key. Cache failures are logged and do not fail the purchase/sync. The cache is not yet used to skip re-posting.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fed4e4d2ba55812a089b10ee70ca30fb4c283d91. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/amazon/kepler-file-system-loader.ts
+++ b/src/amazon/kepler-file-system-loader.ts
@@ -1,0 +1,53 @@
+import type { KeplerFileSystem as AmazonKeplerFileSystem } from "@amazon-devices/kepler-file-system";
+import { ErrorCode, PurchasesError } from "../entities/errors";
+
+/**
+ * Shared Kepler File System loader contract used by VegaDeviceCache.
+ *
+ * Keeping this as an injected dependency lets the default purchases-js module
+ * avoid importing the Vega-only File System package. The Vega entry point
+ * installs the native module implementation before re-exporting the public
+ * SDK API.
+ */
+export type KeplerFileSystem = typeof AmazonKeplerFileSystem;
+export type KeplerFileSystemLoader = () => Promise<KeplerFileSystem>;
+
+const missingKeplerFileSystemLoader: KeplerFileSystemLoader = async () => {
+  throw new PurchasesError(
+    ErrorCode.ConfigurationError,
+    "Kepler File System is supported only by the @revenuecat/purchases-js/vega entry point.",
+  );
+};
+
+let keplerFileSystemLoader: KeplerFileSystemLoader =
+  missingKeplerFileSystemLoader;
+
+/**
+ * Installs the runtime-specific Kepler File System implementation.
+ *
+ * @internal
+ */
+export function setKeplerFileSystemLoader(
+  loader: KeplerFileSystemLoader,
+): void {
+  keplerFileSystemLoader = loader;
+}
+
+/**
+ * Restores the loader used when no Vega-capable entry point has configured one.
+ *
+ * @internal
+ */
+export function resetKeplerFileSystemLoader(): void {
+  keplerFileSystemLoader = missingKeplerFileSystemLoader;
+}
+
+/**
+ * Gets the Kepler File System through the implementation selected by the
+ * active entry point.
+ *
+ * @internal
+ */
+export function loadKeplerFileSystem(): Promise<KeplerFileSystem> {
+  return keplerFileSystemLoader();
+}
--- a/src/amazon/vega-device-cache.ts
+++ b/src/amazon/vega-device-cache.ts
@@ -39,11 +39,33 @@ export class VegaDeviceCache {
       }
 
       const fileSystem = await this.fileSystemLoader();
-      await fileSystem.writeStringToFile(
-        this.tokensCachePath,
-        JSON.stringify([...receiptIds, receiptId]),
-        "UTF-8",
-      );
+      const updatedReceiptIds = [...receiptIds, receiptId];
+      if (await fileSystem.exists(this.tokensCachePath)) {
+        await fileSystem.removeFile(this.tokensCachePath);
+      }
+      try {
+        await fileSystem.writeStringToFile(
+          this.tokensCachePath,
+          JSON.stringify(updatedReceiptIds),
+          "UTF-8",
+        );
+      } catch (error: unknown) {
+        if (!isAlreadyExistsError(error)) {
+          throw error;
+        }
+
+        const currentReceiptIds = await this.getReceiptIds();
+        if (currentReceiptIds.includes(receiptId)) {
+          return;
+        }
+
+        await fileSystem.removeFile(this.tokensCachePath);
+        await fileSystem.writeStringToFile(
+          this.tokensCachePath,
+          JSON.stringify([...currentReceiptIds, receiptId]),
+          "UTF-8",
+        );
+      }
     });
 
     this.writeQueue = operation.catch(() => undefined);
@@ -74,4 +96,8 @@ function isReceiptCache(value: unknown): value is ReceiptCache {
   return (
     Array.isArray(value) && value.every((entry) => typeof entry === "string")
   );
+}
+
+function isAlreadyExistsError(error: unknown): boolean {
+  return String(error).includes("AlreadyExistsError");
 }
--- a/src/amazon/vega-device-cache.ts
+++ b/src/amazon/vega-device-cache.ts
@@ -20,7 +20,7 @@ export class VegaDeviceCache {
     private readonly apiKey: string,
     private readonly fileSystemLoader: KeplerFileSystemLoader = loadKeplerFileSystem,
   ) {
-    this.tokensCachePath = `/data/revenuecat/${this.sharedPreferencesPrefix}${this.apiKey}.${this.tokensCacheKey}`;
+    this.tokensCachePath = `/data/${this.sharedPreferencesPrefix}${this.apiKey}.${this.tokensCacheKey}`;
   }
 
   public async getPreviouslySentReceiptIds(): Promise<Set<string>> {
--- a/src/amazon/vega-device-cache.ts
+++ b/src/amazon/vega-device-cache.ts
@@ -11,8 +11,16 @@ type ReceiptCache = string[];
  * @internal
  */
 export class VegaDeviceCache {
-  private readonly rcStoragePrefix = "com.revenuecat.purchases.";
-  private readonly tokensCacheKey = "tokens";
+  // Apps operate in a sandboxed environment with a restricted view of the device file system.
+  // Vega makes the following app-specific directories available in the sandbox.
+  //
+  // The /data directory is a writable location for app data. It is persistent across
+  // device reboots and package upgrades, and is not shared between apps or services.
+  // The directory isn't persistent on app uninstall.
+  //
+  // More info: https://developer.amazon.com/docs/vega-api/0.24/README.amazon-devices_kepler-file-system.html
+  private readonly dataDirectory = "data";
+  private readonly rcStoragePrefix = "com.revenuecat.purchases";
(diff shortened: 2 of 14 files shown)


src/amazon/vega-device-cache.ts
```diff
@@ -0,0 +1,103 @@
...
+  private readonly sharedPreferencesPrefix = "com.revenuecat.purchases.";
+  private readonly tokensCacheKey = "tokens";
+  private readonly tokensCachePath: string;
+  private writeQueue: Promise<void> = Promise.resolve();
+
+  public constructor(
+    private readonly apiKey: string,
+    private readonly fileSystemLoader: KeplerFileSystemLoader = loadKeplerFileSystem,
+  ) {
+    this.tokensCachePath = `/revenuecat/${this.sharedPreferencesPrefix}${this.apiKey}.${this.tokensCacheKey}`;
```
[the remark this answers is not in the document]
fire-at-will: fixed in 9fb8e0d61669f1bc8cc344cb32126a84606cc4cd

src/amazon/vega-device-cache.ts
```diff
@@ -0,0 +1,95 @@
...
+  ) {
+    this.tokensCachePath = `/data/${this.sharedPreferencesPrefix}${this.apiKey}.${this.tokensCacheKey}`;
+  }
+
+  public async getPreviouslySentReceiptIds(): Promise<Set<string>> {
+    await this.writeQueue;
+    return new Set(await this.getReceiptIds());
+  }
+
+  public async addSuccessfullyPostedReceiptId(
```
tonidero, line 42: So I see we're only adding receipts now... I'm wondering if we should mimick Android and clear these when changing user... Though not sure if that would mean we would need to be caching the user id and data as well... and we would be entering a bit of a rabbit hole 🤔

src/amazon/vega-device-cache.ts
```diff
@@ -0,0 +1,95 @@
...
+    receiptId: string,
+  ): Promise<void> {
+    const operation = this.writeQueue.then(async () => {
```
tonidero, line 45: Hmm TBH, not sure if this is standard (implementing a queue as concatenated promises 😅)... I guess my concern is possible memory leaks... But I guess it shouldn't happen?

src/amazon/vega-device-cache.ts
```diff
@@ -0,0 +1,95 @@
...
+
+type ReceiptCache = string[];
+
+/**
+ * Class capable of caching items on a Vega device.
+ *
+ * @internal
+ */
+export class VegaDeviceCache {
+  private readonly sharedPreferencesPrefix = "com.revenuecat.purchases.";
```
tonidero: I guess `sharedPreferences` doesn't really mean anything here and could be more confusing 😅... maybe just `rcStoragePrefix`?

src/amazon/kepler-file-system-loader.ts
```diff
@@ -0,0 +1,53 @@
...
+}
+
+/**
+ * Gets the Kepler File System through the implementation selected by the
+ * active entry point.
+ *
+ * @internal
+ */
+export function loadKeplerFileSystem(): Promise<KeplerFileSystem> {
+  return keplerFileSystemLoader();
```
tonidero, line 52: I have no idea how this works 😅 But quick question would files stored here be accessible to users and/or other apps?

src/amazon/vega-device-cache.ts
```diff
@@ -0,0 +1,95 @@
...
+  private readonly tokensCacheKey = "tokens";
+  private readonly tokensCachePath: string;
+  private writeQueue: Promise<void> = Promise.resolve();
+
+  public constructor(
+    private readonly apiKey: string,
+    private readonly fileSystemLoader: KeplerFileSystemLoader = loadKeplerFileSystem,
+  ) {
+    this.tokensCachePath = `/data/${this.sharedPreferencesPrefix}${this.apiKey}.${this.tokensCacheKey}`;
```
tonidero: I would probably add some folder to contain these files, in case we need to add more files later.

src/amazon/vega-device-cache.ts
```diff
@@ -0,0 +1,95 @@
...
+      const receiptIds = await this.getReceiptIds();
+
+      if (receiptIds.includes(receiptId)) {
+        return;
+      }
+
+      const fileSystem = await this.fileSystemLoader();
+      const updatedReceiptIds = [...receiptIds, receiptId];
+      if (await fileSystem.exists(this.tokensCachePath)) {
+        await fileSystem.removeFile(this.tokensCachePath);
```
tonidero, line 55: Could this throw some type of IO exception? Also, no idea if we should be threading all these to be executed on a background thread 😅

tonidero (commented):
Mostly questions... but I think this is starting to get complicated 😨

src/amazon/kepler-file-system-loader.ts
fire-at-will: They shouldn't be accessible to users unless they go digging! Not sure if it'd be possible without jailbreaking or enabling developer mode. According to the [KeplerFileSystem docs](https://developer.amazon.com/docs/vega-api/0.23/README.amazon-devices_kepler-file-system.html),

> Apps operate in a sandboxed environment with a restricted view of the device file system. Vega makes the following app-specific directories available in the sandbox: 

> `/data`: Writable location for the app data. Persistent across device reboots and package upgrades, not shared between apps or services. This directory isn't persistent on app uninstall.

src/amazon/vega-device-cache.ts
fire-at-will: Renamed in 85a3297967e0bcb770d0ddffb3622b8a0e0a5889 👍

src/amazon/vega-device-cache.ts
fire-at-will: I agree, but believe it or not...there is no API to create new directories. `@amazon-devices/kepler-file-system`'s APIs expose functions to _delete_ directories, but none to create them 😅 

If you try to write a file to a location that doesn't exist (like `/data/com.revenuecat.purchases/<apiKey>.tokens`), the `writeToString()` function throws an error:
```
{ 
   "message" : "[com.amazon.kepler.io.IoError]: Unable to open file: No such file or directory"
}
```

All of that said, I've added in some comments in 7aa385f481bd136ef374003cd789b0ba06b672c2 to clarify this for future agents/humans who read the code

fire-at-will: There's a list of the file system operations that we can perform here: https://developer.amazon.com/docs/vega-api/0.24/interfaces.KeplerFileSystem.amazon-devices_kepler-file-system.html

src/amazon/vega-device-cache.ts
fire-at-will: Maybe? I think that might be overkill for now. These calls are all `awaiting` their results and return promises, so the rest of the code/app won't be blocking on I/O operations. Let me know if you think it's worth looking into further now!

src/amazon/vega-device-cache.ts
fire-at-will: The promise queue is the only way I can think of to guarantee that we aren't doing multiple writes at the same time without importing some sort of mutex/lock system, and as far as I can find, the Vega runtime doesn't expose any kind of synchronization primitive. We could maybe write our own lock primitive, but it would probably use something similar under the hood.

As far as memory leaks go, the only thing that I can see happening is if the write operations get enqueued faster than the file system can serve them.

In a typical `syncPurchases()`/`restorePurchases()` execution, we do call `addSuccessfullyPostedReceiptId()` in a loop as the receipts are processed, but those loops are slow because of the POST /receipt and notifyFulfillment() calls, and they also await the `addSuccessfullyPostedReceiptId()` calls before continuing their loops.

It's _possible_ that if `syncPurchases()`/`restorePurchases()` got called in multiple threads in a loop, maybe the queue good grow too much over time, but if that happens, I think we might have bigger problems 🤷‍♂️

Let me know what you think!

src/amazon/vega-device-cache.ts
tonidero: Hmm well... I'm just unsure if this could cause frame drops, which wouldn't be great. I'm not sure if this library already handles the threading themselves with the async/await APIs though... I would say this is not a blocker if we want to have something ready for the beta this week though, but probably we should confirm this before considering it stable.

src/amazon/vega-device-cache.ts
tonidero: Wow, that's pretty crazy 😅 Oh well, let's go with this then!

src/amazon/vega-device-cache.ts
tonidero: Ok... let's go with this for now, and we can revisit later... I think it should be ok for now.

src/amazon/vega-device-cache.ts
tonidero: Wdyt about this? TBH, I'm not sure if it's required though, so happy to go with the current implementation for now.

tonidero (approved):

fire-at-will:

PR #1075, opened 2026-08-21
fix(vega): post product price for purchases

fire-at-will:
## Changes introduced
Before this PR, we weren't posting the product's price when calling POST /receipt. This means that if we don't already have a price for the product in the backend, the backend may never be able to get one.

This PR modifies the SDK to include the Amazon product's price when posting a receipt for a new purchase.

---

> [!NOTE]
> **Medium Risk**
> Touches receipt posting for Amazon purchases, which feeds backend revenue/price data. Restore and other callers still send `price: null`.
> 
> **Overview**
> Amazon purchases now send **price and currency** on `POST /receipts` so the backend can record a product price when it does not already have one.
> 
> `postReceipt` takes an optional `price` (major units) and always includes `price` in the body (`null` if omitted). On purchase, the SDK uses the selected option’s (or default option’s) price from the package—currency from that price, amount from `amountMicros`—instead of the package’s web-billing currency alone. Restore/sync still omit price.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ff12f3441ce9ccc6c41859c81841a25980661f05. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/amazon/amazon-billing-wrapper.ts
+++ b/src/amazon/amazon-billing-wrapper.ts
@@ -5,11 +5,11 @@ import type {
   Receipt,
 } from "@amazon-devices/keplerscript-appstore-iap-lib";
 import { ErrorCode, PurchasesError } from "../entities/errors";
-import type {
-  NonSubscriptionOption,
-  Price,
-  PurchaseOption,
-  SubscriptionOption,
+import {
+  ProductType as RevenueCatProductType,
+  type NonSubscriptionOption,
+  type Price,
+  type SubscriptionOption,
 } from "../entities/offerings";
 import { Logger } from "../helpers/logger";
 import type { BillingWrapper } from "../helpers/billing-wrapper";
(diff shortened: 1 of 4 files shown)


src/amazon/amazon-billing-wrapper.ts
```diff
@@ -129,15 +135,19 @@ export class AmazonBillingWrapper implements BillingWrapper {
...
       productIdentifier,
-      rcPackage.webBillingProduct.price.currency,
+      price?.currency ?? null,
       receipt.receiptId,
       rcPackage.webBillingProduct.presentedOfferingContext,
       PostReceiptInitiationSource.PURCHASE,
       undefined,
       storeUserId,
+      false,
+      price ? price.amountMicros / 1_000_000 : undefined,
```
tonidero, line 150: With so many parameters, it could be worth refactoring this method to get some object with named parameters 😅 But not for this PR.

tonidero (approved):

src/amazon/amazon-billing-wrapper.ts
```diff
@@ -644,6 +654,15 @@ export class AmazonBillingWrapper implements BillingWrapper {
   }
 }
 
+function priceForPurchaseParams(params: PurchaseParams): Price | null {
+  const product = params.rcPackage.webBillingProduct;
+  const purchaseOption = params.purchaseOption ?? product.defaultPurchaseOption;
+
+  return product.productType === RevenueCatProductType.Subscription
+    ? (purchaseOption as SubscriptionOption).base.price
```
tonidero, line 662: Actually, I guess I was thinking this doesn't account intro prices or free trials right? Not sure if it's something we should consider...

fire-at-will: Yeah, it doesn't take intro prices or free trials into account, but neither does the Android implementation of Amazon. I'll look and see if this is something we should take into account...

fire-at-will:

PR #1077, opened 2026-08-22
[EXTERNAL] Allow Paddle checkout settings to carry non-boolean values (unblocks express checkout)

### LINKED ISSUE #1072: Paddle: express checkout is unreachable — checkout_settings is typed Record<string, boolean>
### Summary

Paddle supports [express checkout](https://developer.paddle.com/concepts/sell/express-checkout), which presents Apple Pay / Google Pay first and captures customer details from the wallet and IP geolocation rather than a form. For a mobile-first funnel this removes the email/country step entirely.

Because `purchases-js` owns the Paddle.js integration, apps can't enable it themselves. Today it looks unreachable for two small reasons.

### 1. `variant` is hardcoded, but the override path exists

In `src/paddle/paddle-service.ts`:

```ts
const commonSettings = {
  theme,
  variant: "one-page" as Variant,
  locale,
  allowLogout: false,
  showAddDiscounts: false,
  showAddTaxId: false,
  allowDiscountRemoval: false,
  ...checkoutSettings,
};
```

`...checkoutSettings` is spread *after* the defaults, so a backend-supplied `variant` would already win. The mechanism is in place.

### 2. The type can't express it

`checkoutSettings` is `Record<string, boolean>`, matching `checkout_settings` in `src/networking/responses/checkout-prepare-response.ts` and `checkout-start-response.ts`. Paddle's `variant` is a string, so it can't be represented — and `showNonExpressPaymentMethods` (Paddle's express fallback control) isn't forwarded either.

### Suggested change

- Widen `checkout_settings` / `checkoutSettings` to something like `Record<string, boolean | string>`, or a typed subset of Paddle's checkout settings.
- Forward express checkout from the dashboard's Paddle checkout settings, alongside the existing `showAddDiscounts` passthrough.

The SDK half looks like a small type change; the dashboard/backend half is the part only RevenueCat can do. Happy to open a PR for the SDK side if that direction is welcome.

### Notes

- Apple Pay domain verification is a Paddle-side prerequisite for the one-click experience and is independent of this.
- `presentExpressPurchaseButton` is Stripe-backed, so it isn't an alternative for Paddle Billing apps.

### Environment

`@revenuecat/purchases-js` 1.53.1, Paddle Billing.

alepacheco:
## Motivation / Description

Fixes #1072.

Paddle supports [express checkout](https://developer.paddle.com/concepts/sell/express-checkout), which leads with Apple Pay / Google Pay and takes customer details from the wallet instead of a form. Because `purchases-js` owns the Paddle.js integration, apps can't turn it on themselves — it has to come through `checkout_settings`.

The passthrough mechanism is already correct: `buildPaddleCheckoutOptions` spreads `...checkoutSettings` *after* the SDK defaults, so a backend-supplied `variant` already wins over the hardcoded `"one-page"`. The only thing in the way is the type — `checkout_settings` / `checkoutSettings` are `Record<string, boolean>`, and Paddle's `variant` is `"one-page" | "multi-page" | "express"`, so express checkout can't be expressed at all.

This PR is the SDK half only: it makes the setting representable. Actually forwarding it from the dashboard's Paddle checkout settings (alongside the existing `showAddDiscounts` passthrough) is the backend/dashboard half, which only RevenueCat can do.

## Changes introduced

- Added `PaddleCheckoutSettings` (`Record<string, boolean | string>`) in `src/networking/responses/paddle-checkout-settings.ts`, mirroring how `stripe-elements.ts` holds gateway-specific types shared between the prepare and start responses.
- Used it for `checkout_settings` in `CheckoutPreparePaddleBillingParams` and `PaddleCheckoutStartResponse`, and for the `checkoutSettings` params in `paddle-service.ts`.
- Tests covering `variant: "express"` overriding the default, and an unknown setting (`showNonExpressPaymentMethods`) reaching `Checkout.open()` untouched — both through `buildPaddleCheckoutOptions` and through `PaddleService.purchase`.

Notes on the shape of the type:

- It stays an **open** record rather than a closed union of the settings known today, so the backend can start forwarding a newly supported Paddle setting without waiting on an SDK release. `showNonExpressPaymentMethods` is a concrete example: it's documented by Paddle but isn't in `@paddle/paddle-js@1.6.4`'s `CheckoutSettings` type yet.
- Type safety on the settings the SDK controls is unchanged. The explicit defaults (`theme`, `locale`) and the layout settings (`displayMode`, `frameTarget`, `frameInitialHeight`, `frameStyle`) keep their exact types and are still checked against Paddle's `CheckoutSettings` — I verified that a bad `theme` literal still fails to compile after the widening.
- No public API surface change; `api-report/` is unchanged.

Happy to adjust the direction — e.g. a closed union of the settings you intend to support, or `Partial<CheckoutSettings>` from `@paddle/paddle-js` — if you'd rather constrain it.

## Linear ticket (if any)

N/A — external contribution.

## Additional comments

- `pnpm typecheck`, `pnpm test:typecheck`, `pnpm lint`, `pnpm svelte-check` (0 errors), `pnpm build` and `pnpm test` (816 passing) all pass locally.
- Apple Pay domain verification is a Paddle-side prerequisite for the one-click experience and is independent of this change.
- `presentExpressPurchaseButton` is Stripe-backed, so it isn't an alternative for Paddle Billing apps.

---

> [!NOTE]
> **Low Risk**
> Type-only widening of an existing passthrough; merge order and SDK-controlled layout settings are unchanged. No public API or payment-flow logic change.
> 
> **Overview**
> Widens Paddle `checkout_settings` so dashboard/backend values can include strings (e.g. `variant: "express"`), unblocking express checkout without an SDK release for each new Paddle setting.
> 
> Adds shared `PaddleCheckoutSettings` (`Record<string, boolean | string>`) and uses it on prepare/start responses and in `buildPaddleCheckoutOptions`. Settings still spread after SDK defaults, so a backend `variant` overrides `"one-page"`. Tests cover express override and forwarding unknown keys like `showNonExpressPaymentMethods`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 432e24e756aa7cdb0ba68931d33b55c02521013b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/networking/responses/paddle-checkout-settings.ts
+++ b/src/networking/responses/paddle-checkout-settings.ts
@@ -2,12 +2,7 @@
  * Checkout settings configured on the RevenueCat dashboard and forwarded by the
  * backend, merged over the SDK's defaults before being handed to Paddle's
  * `Checkout.open({ settings })`.
- *
- * Keys and values map 1:1 to Paddle's own checkout settings, which are not all
- * booleans: `variant`, for example, is `"one-page" | "multi-page" | "express"`,
- * so express checkout can only be expressed once strings are allowed here.
- * Kept as an open record (rather than a closed union of the settings we know
- * about today) so the backend can forward newly supported Paddle settings
+ * Intentionally an open record so we can forward newly supported Paddle settings
  * without requiring an SDK release.
  */
 export type PaddleCheckoutSettings = Record<string, boolean | string>;
--- a/src/networking/responses/paddle-checkout-settings.ts
+++ b/src/networking/responses/paddle-checkout-settings.ts
@@ -1,0 +1,13 @@
+/**
+ * Checkout settings configured on the RevenueCat dashboard and forwarded by the
+ * backend, merged over the SDK's defaults before being handed to Paddle's
+ * `Checkout.open({ settings })`.
+ *
+ * Keys and values map 1:1 to Paddle's own checkout settings, which are not all
+ * booleans: `variant`, for example, is `"one-page" | "multi-page" | "express"`,
+ * so express checkout can only be expressed once strings are allowed here.
+ * Kept as an open record (rather than a closed union of the settings we know
+ * about today) so the backend can forward newly supported Paddle settings
+ * without requiring an SDK release.
+ */
+export type PaddleCheckoutSettings = Record<string, boolean | string>;
--- a/src/networking/responses/checkout-prepare-response.ts
+++ b/src/networking/responses/checkout-prepare-response.ts
@@ -1,4 +1,5 @@
 import type { GatewayParams } from "./stripe-elements";
+import type { PaddleCheckoutSettings } from "./paddle-checkout-settings";
 
 export type CheckoutPrepareStripeGatewayParams = GatewayParams;
 
@@ -9,7 +10,7 @@ export interface CheckoutPreparePayPalGatewayParams {
 export interface CheckoutPreparePaddleBillingParams {
   client_side_token: string;
   is_sandbox: boolean;
-  checkout_settings?: Record<string, boolean>;
+  checkout_settings?: PaddleCheckoutSettings;
 }
 
 export interface CheckoutPrepareResponse {
--- a/src/networking/responses/checkout-start-response.ts
+++ b/src/networking/responses/checkout-start-response.ts
@@ -1,4 +1,5 @@
 import type { GatewayParams } from "./stripe-elements";
+import type { PaddleCheckoutSettings } from "./paddle-checkout-settings";
 
 export interface StripeBillingParams {
   client_secret: string;
@@ -26,7 +27,7 @@ export interface PaddleCheckoutStartResponse {
     client_side_token: string;
     is_sandbox: boolean;
     transaction_id: string;
-    checkout_settings?: Record<string, boolean>;
+    checkout_settings?: PaddleCheckoutSettings;
     /**
      * Per-project gate for the Paddle inline checkout rollout. When `true`,
      * present Paddle's checkout inline (embedded in our own container); when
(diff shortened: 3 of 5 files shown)


src/networking/responses/paddle-checkout-settings.ts
```diff
@@ -0,0 +1,13 @@
...
+ * Checkout settings configured on the RevenueCat dashboard and forwarded by the
+ * backend, merged over the SDK's defaults before being handed to Paddle's
+ * `Checkout.open({ settings })`.
+ *
+ * Keys and values map 1:1 to Paddle's own checkout settings, which are not all
+ * booleans: `variant`, for example, is `"one-page" | "multi-page" | "express"`,
+ * so express checkout can only be expressed once strings are allowed here.
+ * Kept as an open record (rather than a closed union of the settings we know
+ * about today) so the backend can forward newly supported Paddle settings
+ * without requiring an SDK release.
```
james-od, line 6: As-is this comment gives lots of context for the review, but for the longer term could we change this to:
```suggestion
 * Checkout settings configured on the RevenueCat dashboard and forwarded by the
 * backend, merged over the SDK's defaults before being handed to Paddle's
 * `Checkout.open({ settings })`.
 * Intentionally an open record so we can forward newly supported Paddle settings
 * without requiring an SDK release.
```

james-od (approved):
Very nice, thanks for raising! I'll handle the backend + dashboard side next but I think this can be merged separately. Once you're able to fix the nit on the comment I'll handle merging this in and will make sure the changelog shows you as a contributor! 

> Happy to adjust the direction — e.g. a closed union of the settings you intend to support, or Partial<CheckoutSettings> from @paddle/paddle-js — if you'd rather constrain it.

Since this isn't part of the public API I think it's probably better to leave it open for now and keep any narrowing in the backend, so happy to leave what you have. We can tighten as a followup if needed.

Thanks!

src/networking/responses/paddle-checkout-settings.ts
alepacheco: Updated to the suggested wording in 432e24e. Thanks!

PR #1082, opened 2026-08-24
feat(vega): run sync on app foregrounding

fire-at-will:
## Changes introduced
Runs the unposted receipt sync whenever the vega app is foregrounded in addition to when the SDK is configured. This is a crucial step in supporting Quick Subscribe.

The foregrounding transition is detected using the React Native `AppState` APIs, which means that we needed to add React Native as a dependency for the `AmazonBillingWrapper`, while keeping it out of the base `purchases-js` module.

If the sync triggered by foregrounding occurs while a sync is ongoing from the SDK being configured (or vise versa), the existing promise is awaited so that we don't run to sync operations at the same time.

### Testing
- Added a few unit tests
- Manually verified behavior

---

> [!NOTE]
> **Medium Risk**
> Changes automatic Amazon receipt posting and backend sync timing on foreground; mistakes could duplicate work or miss receipts, though coalescing and cache checks limit overlap.
> 
> **Overview**
> **Amazon / Vega billing** now re-syncs unposted receipts when the app returns to the foreground, in addition to the existing sync at SDK initialization. That foreground path is meant to support Quick Subscribe flows where purchases may need posting after the user comes back to the app.
> 
> Foreground detection uses React Native **`AppState`** through a new **`react-native-app-state-loader`** (same pluggable pattern as the Amazon IAP loader). **`react-native`** is added as an optional peer dependency and wired only from the **`@revenuecat/purchases-js/vega`** entry point so the default web bundle stays unchanged.
> 
> **`AmazonBillingWrapper`** takes a **`getAppUserId`** callback instead of a fixed user id, coalesces overlapping background syncs onto one in-flight promise, and swallows sync failures into **`warnLog`** so fire-and-forget callers don’t get unhandled rejections. **`close()`** on the billing wrapper (and **`Purchases.configure` / `Purchases.close`**) removes the AppState listener so reconfiguration doesn’t leave stale background syncs running.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f4024945ad8f48d8d371f12e6e590b574c8bd430. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,6 +67,7 @@
   "devDependencies": {
     "@amazon-devices/kepler-file-system": "0.0.7",
     "@amazon-devices/keplerscript-appstore-iap-lib": "2.12.16",
+    "react-native": "0.72.0",
     "@eslint/js": "^9.16.0",
     "@microsoft/api-extractor": "^7.48.0",
     "@paddle/paddle-js": "^1.6.4",
@@ -114,13 +115,17 @@
   },
   "peerDependencies": {
     "@amazon-devices/kepler-file-system": "0.0.7",
-    "@amazon-devices/keplerscript-appstore-iap-lib": "2.12.16"
+    "@amazon-devices/keplerscript-appstore-iap-lib": "2.12.16",
+    "react-native": ">=0.72.0"
   },
   "peerDependenciesMeta": {
     "@amazon-devices/kepler-file-system": {
       "optional": true
     },
     "@amazon-devices/keplerscript-appstore-iap-lib": {
+      "optional": true
+    },
+    "react-native": {
       "optional": true
     }
   },
--- a/src/amazon/amazon-billing-wrapper.ts
+++ b/src/amazon/amazon-billing-wrapper.ts
@@ -32,6 +32,10 @@ import {
   type AmazonAppstoreIAPSDK,
 } from "./amazon-appstore-iap-sdk-loader";
 import { VegaDeviceCache } from "./vega-device-cache";
+import {
+  loadReactNativeAppState,
+  type AppStateSubscription,
+} from "./react-native-app-state-loader";
 
 type ReceiptWithStoreUserId = {
   receipt: Receipt;
@@ -45,25 +49,29 @@ type ReceiptWithStoreUserId = {
  */
 export class AmazonBillingWrapper implements BillingWrapper {
   private readonly deviceCache: VegaDeviceCache;
+  private appStateSubscription: AppStateSubscription | undefined;
+  private pendingSync: Promise<void> | undefined;
+  private closed = false;
 
   constructor(
     private readonly backend: Backend,
     apiKey: string,
-    appUserId?: string,
+    private readonly getAppUserId: () => string | undefined,
   ) {
     this.deviceCache = new VegaDeviceCache(apiKey);
-    if (appUserId !== undefined) {
-      void this.syncPendingPurchases(appUserId).catch((error: unknown) => {
-        Logger.warnLog(
-          `Failed to sync pending Amazon purchases in the background: ${String(error)}`,
-        );
-      });
-    }
+    void this.triggerSyncPendingPurchases();
+    void this.observeAppState();
   }
 
   private amazonAppstoreIAPSDKPromise:
     | Promise<AmazonAppstoreIAPSDK>
     | undefined;
+
+  public close(): void {
+    this.closed = true;
+    this.appStateSubscription?.remove();
+    this.appStateSubscription = undefined;
+  }
 
   public async getProducts(
     _appUserId: string,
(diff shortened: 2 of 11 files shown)


src/main.ts
```diff
@@ -621,7 +621,7 @@ export class Purchases {
       this.amazonBillingWrapper = new AmazonBillingWrapper(
         this.backend,
         this._API_KEY,
-        this._appUserId,
+        () => this._appUserId,
```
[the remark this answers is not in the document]
fire-at-will: Fixed in f4024945ad8f48d8d371f12e6e590b574c8bd430

src/amazon/amazon-billing-wrapper.ts
```diff
@@ -189,13 +197,33 @@ export class AmazonBillingWrapper implements BillingWrapper {
...
-   * Posts and fulfills receipts which have not been successfully handled on
-   * this device. This is intended for the background automatic sync performed
-   * when the app starts or returns to the foreground.
-   * @internal
-   */
-  public async syncPendingPurchases(appUserId: string): Promise<void> {
+  private async triggerSyncPendingPurchases(): Promise<void> {
+    if (this.pendingSync !== undefined) {
+      await this.pendingSync;
+      return;
```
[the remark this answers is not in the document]
fire-at-will: It's a bit complicated, but addressed in 19058f14728191a9687ae06483eae77f3bdb278b

src/amazon/amazon-billing-wrapper.ts
```diff
@@ -45,26 +49,30 @@ type ReceiptWithStoreUserId = {
...
   private readonly deviceCache: VegaDeviceCache;
+  private appStateSubscription: AppStateSubscription | undefined;
+  private pendingSync: Promise<void> | undefined;
+  private closed = false;
 
   constructor(
     private readonly backend: Backend,
     apiKey: string,
-    appUserId?: string,
+    private readonly getAppUserId: () => string | undefined,
```
fire-at-will, line 59: This protects us from an app user ID change post-configure

package.json
```diff
@@ -114,14 +115,18 @@
   },
   "peerDependencies": {
     "@amazon-devices/kepler-file-system": "0.0.7",
-    "@amazon-devices/keplerscript-appstore-iap-lib": "2.12.16"
+    "@amazon-devices/keplerscript-appstore-iap-lib": "2.12.16",
+    "react-native": ">=0.72.0"
```
tonidero, line 119: We might want to increase the min version we support here? Not sure the min RN version amazon vega supports... But this way we could make sure we avoid breaking changes later... Though on the other hand, I'm not sure if it would result in some incompatibilities with `react-native-purchases` which has `>=0.73.0`... So maybe just use the same as that SDK just in case?

tonidero (approved):
I think this makes sense! I have a slight concern of having react-native as a peer dependency might put off some folks looking at the dependencies used by purchases-js... But hopefully it's not an issue 🤞

package.json
fire-at-will: @tonidero So the answer here is actually a bit complicated 

React Native is bundled in the Vega OS itself, and right now, the Vega OS only ships two different versions of React Native: 0.72.0 and 0.83.0, so we have to choose between those two versions.

Also, support for 0.83.0 wasn't introduced until Vega OS version 1.2, so if we require 0.83.0, our SDK's minOS version will be Vega OS 1.2. FWIW, the most recent version of the Vega SDK (0.24), only supports Vega OS 1.2 ([source](https://developer.amazon.com/docs/vega/0.24/vega-release-notes)).

I'm not sure yet if declaring a different RN version here will affect `react-native-purchases`, since it's just a peerDependency. I'll do some testing this morning to see.

References:
- RN/OS coupling: https://developer.amazon.com/docs/vega/0.24/target-os-version
- RN overview for Vega: https://developer.amazon.com/docs/vega/0.24/vega-rn-overview

tonidero: Ohhh ok 😅 that is indeed complicated... I guess I do wonder if we want to only support newer Vega OS, so we are not forced to support such older RN versions... But let's go with this for now I would say :)

fire-at-will: Okay, I think that this won't cause any issues for a few reasons:

#### 1. `react-native-purchases`'s dependency requirement is greater than `purchases-js`'s
- `purchases-js` has a peerDependency on `>=0.72.0`
- `react-native-purchases` has a peerDependency on `>= 0.73.0`

Luckily for us, any RN version that satisfies `react-native-purchases`'s `>= 0.73.0` requirement will satisfy `purchases-js`'s `>=0.72.0` peerDependency requirement.

#### 2. Hybrids don't use the vega entry point
The hybrid mappings import the normal `@revenuecat/purchases-js` entry point, not the `/vega` entry point. The normal entry point has no static react-native import. That import exists only in the Vega entry point, whose build keeps it external.

#### 3. The dependency in purchases-js is only required for building/testing purchases-js
The exact 0.72.0 dependency in `purchases-js` is a dev dependency for building/testing this repo, and isn't forced on consuming apps.

All of that said, you never know till you try it, so I'll make a local build of an app using this branch of purchases-js in the `react-native-purchases` SDK and confirm that it builds :)

fire-at-will: Okay, I double checked with a build, and the React Native purchase tester app compiles for both iOS/Android with this branch of purchases-js :)

fire-at-will: Thanks for calling this out, that could have been a serious issue that we might not have found for a while!

PR #1083, opened 2026-08-24
fix(vega): handle exists() function not available on OS

fire-at-will:
## Motivation / Description
The file system's `exist()` function wasn't introduced until the most recent version of the Kepler file system, `0.0.7`. This library is baked in to the Vega OS, so if the SDK is running on an OS version with a file system version <0.0.7, then the `exists()` function will be `undefined` at runtime, causing issues.

## Changes introduced
- We introduce a new `safeExists()` function that serves as a wrapper around `exists()`. It checks to see if `exists()` is available, and if so, uses it. If it isn't available, it gracefully falls back to attempting to read the file, while throws if the file isn't present on the disk.
- This introduces a dependency on `@amazon-devices/kepler-compatibility`, which is safely injected to the `purchases-js` SDK in the same manner as other dependencies to ensure that it won't be included for web/hybrid users of the SDK.

## Additional comments
I audited other dependencies' function calls that we use and couldn't find any other instances of functions be introduced that we need to work around.

---

> [!NOTE]
> **Medium Risk**
> Changes Vega on-device receipt cache behavior on older OS versions; incorrect fallback could affect duplicate receipt handling, but scope is limited to the Vega Amazon path.
> 
> **Overview**
> Fixes **Vega receipt token caching** on OS builds where Kepler File System predates **`exists`** (added in **0.0.7**). `VegaDeviceCache` no longer calls `exists` unconditionally; it routes through **`doesCacheFileExist`**, which uses `exists` when the runtime reports support and otherwise infers presence via a **`readFileAsString`** attempt.
> 
> A new **`kepler-compatibility-loader`** holds an injectable OS capability check (default entry point still errors if used without the Vega bundle). The **`@revenuecat/purchases-js/vega`** entry registers **`isPresentOnOS("@amazon-devices/kepler-file-system", "0.0.7")`** from **`@amazon-devices/kepler-compatibility`**, added as an **optional peer** and demo dev dependency, with the Vega Vite build marking that package **external**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2892e94e2c6b1a65edf59d4835687b480033bcc1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/amazon/vega-device-cache.ts
+++ b/src/amazon/vega-device-cache.ts
@@ -1,7 +1,12 @@
 import {
   loadKeplerFileSystem,
+  type KeplerFileSystem,
   type KeplerFileSystemLoader,
 } from "./kepler-file-system-loader";
+import {
+  isKeplerLibraryPresentOnOS,
+  type IsPresentOnOS,
+} from "./kepler-compatibility-loader";
 
 type ReceiptCache = string[];
 
@@ -27,6 +32,7 @@ export class VegaDeviceCache {
   public constructor(
     private readonly apiKey: string,
     private readonly fileSystemLoader: KeplerFileSystemLoader = loadKeplerFileSystem,
+    private readonly isPresentOnOS: IsPresentOnOS = isKeplerLibraryPresentOnOS,
   ) {
     // As of KeplerFileSystem SDK version 0.24, there is no way to create a directory, and any attempts
     // to write to one throw a com.amazon.kepler.file_system.NotFoundError error. Therefore, we write
@@ -51,7 +57,7 @@ export class VegaDeviceCache {
 
       const fileSystem = await this.fileSystemLoader();
       const updatedReceiptIds = [...receiptIds, receiptId];
-      if (await fileSystem.exists(this.tokensCachePath)) {
+      if (await this.safeExists(fileSystem)) {
         await fileSystem.removeFile(this.tokensCachePath);
       }
       try {
@@ -79,10 +85,26 @@ export class VegaDeviceCache {
     await operation;
   }
 
+  /**
+   * `exists` arrived in Kepler File System 0.0.7. On older Vega OS versions,
+   * use a read attempt to determine whether the cache file is present.
+   */
+  private async safeExists(fileSystem: KeplerFileSystem): Promise<boolean> {
+    if (this.isPresentOnOS("@amazon-devices/kepler-file-system", "0.0.7")) {
+      return await fileSystem.exists(this.tokensCachePath);
+    }
+
+    try {
+      await fileSystem.readFileAsString(this.tokensCachePath, "UTF-8");
+      return true;
+    } catch {
+      return false;
+    }
+  }
+
   private async getReceiptIds(): Promise<ReceiptCache> {
     const fileSystem = await this.fileSystemLoader();
-    const cacheExists = await fileSystem.exists(this.tokensCachePath);
-    if (!cacheExists) {
+    if (!(await this.safeExists(fileSystem))) {
       return [];
     }
 
--- a/src/amazon/vega-device-cache.ts
+++ b/src/amazon/vega-device-cache.ts
@@ -95,6 +95,7 @@ export class VegaDeviceCache {
     }
 
     try {
+      // Throws if the file isn't found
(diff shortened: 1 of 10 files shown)


src/amazon/vega-device-cache.ts
```diff
@@ -79,10 +85,27 @@ export class VegaDeviceCache {
...
+   * use a read attempt to determine whether the cache file is present.
+   */
+  private async safeExists(fileSystem: KeplerFileSystem): Promise<boolean> {
+    if (this.isKeplerFileSystemExistsSupported()) {
+      return await fileSystem.exists(this.tokensCachePath);
+    }
+
+    try {
+      // Throws if the file isn't found
+      await fileSystem.readFileAsString(this.tokensCachePath, "UTF-8");
```
tonidero: Hmm so since we're just discarding the read here, I wonder if we should have a `safeRemoveFile` and `safeReadFile` instead, so we perform the operations "safely" and do not have to read the file an extra time to ignore the result?

tonidero (approved):
A possible optimization... But I think it makes sense 🙏

src/amazon/vega-device-cache.ts
fire-at-will: Right...

Luckily, removeFile/readFile have been around since the beginning, so we don't actually need a `safeReadFile`/`safeRemoveFile` in the same way that we need a `safeExists` function. Luckily, since this is the fallback, it should only execute on non-current OS versions, and over time it should be run less and less as users update their OS.

I'll see if I can come up with a way to combine the "exists and read if exists" operations into one read operation. I'm trying to ship the beta today, so I'll do this in a follow-up PR that may or may not make it into today's release - we'll see!

src/amazon/vega-device-cache.ts
```diff
@@ -79,10 +85,27 @@ export class VegaDeviceCache {
     await operation;
   }
 
+  /**
+   * `exists` arrived in Kepler File System 0.0.7. On older Vega OS versions,
+   * use a read attempt to determine whether the cache file is present.
+   */
+  private async safeExists(fileSystem: KeplerFileSystem): Promise<boolean> {
```
joshdholtz: Nit: A little unsure about this function name 🤷‍♂️  Is "safe" the way its accessing this? Just not sure what safe is 🙃

src/amazon/vega-device-cache.ts
```diff
@@ -79,10 +85,27 @@ export class VegaDeviceCache {
...
+    }
+
+    try {
+      // Throws if the file isn't found
+      await fileSystem.readFileAsString(this.tokensCachePath, "UTF-8");
+      return true;
+    } catch {
+      return false;
+    }
+  }
```
joshdholtz: Also, does this assume that we are also storing cache as a UTF-8 file? Should that be either in the function name and/or param type?

joshdholtz (approved):
Overall looks good! Just a few small comments but otherwise :shipit:

src/tests/vega.test.ts
```diff
@@ -1,30 +1,38 @@
...
+//   KeplerFileSystem: keplerFileSystem,
+// }));
+
+// vi.mock("@amazon-devices/kepler-compatibility", () => ({
+//   isPresentOnOS,
+// }));
+
+// vi.mock("react-native", () => ({
+//   AppState: appState,
+// }));
```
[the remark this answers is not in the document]
fire-at-will: fixed :)

src/amazon/vega-device-cache.ts
fire-at-will: Good idea! I renamed it to `doesCacheFileExist()` in ca470338aee55e3e86ed3ea277a434eb007b41b9 since the whole fallback bit is really an implementation detail

src/amazon/vega-device-cache.ts
fire-at-will: Good catch! I think that we can assume that since we can control the encoding for both reads/writes across the cache in this file. To make sure that it's standardized, I extracted the encoding to a constant in the file here 2892e94e2c6b1a65edf59d4835687b480033bcc1 so that we always use the same format :)

PR #1080, opened 2026-08-24
feat(singular): add $singularDeviceId reserved customer attribute

guillemcorominas:
## Motivation / Description

Singular's Event Endpoint V2 (mandatory for Singular accounts created on or after
2026-07-15) requires an `sdid`, which is generated by the Singular SDK and must be
forwarded to RevenueCat by the app.

## Changes introduced

Adds `ReservedCustomerAttribute.SingularDeviceId` (`$singularDeviceId`) so it can be
set through `setAttributes`. API report regenerated.

## Linear ticket (if any)

[INT-101](https://linear.app/revenuecat/issue/INT-101/support-singular-event-endpoint-v2-with-sdid)

## Additional comments

Companion PRs in `purchases-ios` and `purchases-android`. Backend support (Singular V2
payload) lands separately in Khepri. PHC, hybrids and KMP follow once the native
releases propagate.

---

> [!NOTE]
> **Low Risk**
> Additive enum and API-report change only; no purchase, auth, or attribute-validation logic is modified in this diff.
> 
> **Overview**
> Adds **`ReservedCustomerAttribute.SingularDeviceId`** (`$singularDeviceId`) so apps can pass Singular’s SDK-generated device ID through **`setAttributes`**, matching Singular Event Endpoint V2’s required `sdid`.
> 
> The public API report is updated so the new enum member appears in the published **`@revenuecat/purchases-js`** types.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5877e0e22c19ae587c1ef4ea11dabb9c940727cd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #1055, opened 2026-08-14
Paddle Metadata

HaleyRevcat:
## Motivation / Description
Currently devs cannot pass metadata to Paddle. This is necessary for some integrations like Tolt.
Backend: khepri/pull/23967

## Changes introduced
src/entities/present-paywall-params.ts
-  Added metadata?: PurchaseMetadata

src/main.ts
- startPurchaseFlow now passes metadata: paywallParams.metadata into this.purchase(...)

api-report/purchases-js.api.md
- ran extract-api to generate it

## Linear ticket (if any)
ticket: 82631

## Additional comments

---

> [!NOTE]
> **Low Risk**
> Additive optional API and pass-through wiring with no auth or breaking signature changes beyond an extended internal wallet render helper.
> 
> **Overview**
> Adds optional **`metadata`** on **`PresentPaywallParams`** so integrators can attach key/value purchase metadata (e.g. partner referral IDs) when a purchase starts from the paywall.
> 
> That value is now threaded into the standard paywall **`purchase`** call and into **express wallet** checkout via **`getWalletButtonRender`** → **`presentExpressPurchaseButton`**, so it can reach the backend, payment gateway, and RC transaction metadata. The public API surface is updated accordingly, with a unit test asserting metadata is passed to the express purchase button.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0a1d4d4c0a001ac92fcb70b7733343fa7848ba14. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #1085, opened 2026-08-25
chore: drop the unused api-extractor doc model

AlvaroBrey:
The `extract-api` script generates both a Markdown and a json file, but CI only checks the markdown. The JSON file has no consumer but it pollutes diffs and drifts (since it's not gated) as seen in [this PR](https://github.com/RevenueCat/purchases-js/pull/1080/changes#diff-f9192e79dce34eebcaff5ca682d02f804c8aa3525c8cbd9bfc11fe433d258b89.) 

This PR disables the JSON export while keeping the markdown, and deletes the JSON file.

<details><summary>Agent description</summary>

## Motivation / Description

One `extract-api` run emits two artifacts: `purchases-js.api.md`, the human-readable report meant to be reviewed as a diff, and `purchases-js.api.json`, the doc model api-extractor hands to `api-documenter` to build a docs site. This repo dropped `api-documenter` long ago and generates docs with typedoc, which reads the TypeScript sources instead, so the json has been output with no input side for a while.

It also drifts silently. The API check runs `extract-api` and then diffs a single path, `api-report/purchases-js.api.md`, so a PR that regenerates the report and commits only the md leaves the json stale with nothing to catch it. That is how #1055 left `PresentPaywallParams.metadata` out of the json for a week, until #1080 regenerated it and carried the catch-up as an unrelated 49-line addition. The only thing that has been repairing it is the `update-error-codes` job, which happens to commit the whole `api-report` directory.

## Changes introduced

- `api-extractor.json`: `docModel.enabled` set to `false`; `apiJsonFilePath`, `includeForgottenExports` and `projectFolderUrl` go with it.
- Deletes `api-report/purchases-js.api.json`.
- `apiReport` is untouched, so `purchases-js.api.md` is still generated and still the file the API check diffs.

**Rejected:**

- Adding the json to the API check (`git diff --exit-code api-report/`): closes the drift but keeps a 10k-line artifact nobody reads, and makes every public-API PR carry its churn on top of the md.
- Removing the `docModel` block outright: fails api-extractor's config schema validation, since `enabled` is required and its defaults file does not supply one.

## Linear ticket (if any)

None.

## Additional comments

The same `docModel` config exists in `purchases-capacitor`, `purchases-hybrid-common` (`typescript/` and `purchases-js-hybrid-mappings/`) and `purchases-ui-js`. Whether their doc models have consumers is unchecked, so this PR does not touch them.

</details>

---

> [!NOTE]
> **Low Risk**
> Build/tooling configuration only; no changes to runtime SDK behavior or the reviewed public API markdown check.
> 
> **Overview**
> Turns off api-extractor’s **doc model** output by setting `docModel.enabled` to `false` and removing the related `apiJsonFilePath`, `includeForgottenExports`, and `projectFolderUrl` settings. The unused `purchases-js.api.json` artifact is removed.
> 
> **Markdown API reporting stays the same** — `apiReport` is unchanged, so `extract-api` still writes `api-report/purchases-js.api.md`, which CI continues to gate with `git diff --exit-code`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ac1c8566e4a4eea607c9cb69fe40264822b39b5c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ajpallares (approved):
Makes sense! Thanks for the cleanup!

tonidero (approved):
Right... I think it makes sense yeah... The markdown version is much nicer to read and parse, so that's the reason we only use that, and yeah, agreed this was just noise. Thank you!!

PR #1088, opened 2026-08-25
fix(vega): set isRestore based on anon status when auto syncing receipts

fire-at-will:
## Motivation / Description
When auto-syncing receipts on SDK configure/app foregrounding, the iOS and Android SDKs post receipts with an `is_restore` value according to the following logic:
- Is `allowSharingAppStoreAccount`/`allowSharingPlayStoreAccount` set?
   - If so, use that value
   - If not, use `is_restore = isUserAnonymous()`

In the current vega implementation, we currently always set this value to `false`, leading to a divergence in auto-sync behaviors across SDKs

## Changes introduced
When posting a receipt from an auto sync, sets the `is_restore` value to true if the user is anonymous, otherwise sets it to `false`. 

This PR intentionally doesn't add a `allowSharingAmazonAccount` since `allowSharingAppStoreAccount`/`allowSharingPlayStoreAccount` are deprecated in `purchases-ios`/`purchases-android`.

### DIFF
--- a/src/amazon/amazon-billing-wrapper.ts
+++ b/src/amazon/amazon-billing-wrapper.ts
@@ -57,6 +57,7 @@ export class AmazonBillingWrapper implements BillingWrapper {
     private readonly backend: Backend,
     apiKey: string,
     private readonly getAppUserId: () => string | undefined,
+    private readonly getIsAnonymous: () => boolean,
   ) {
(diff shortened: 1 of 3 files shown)


joshdholtz (approved):
:shipit: :shipit: :shipit:

PR #1092, opened 2026-08-26
feat: add trackCustomPaywallImpression() function

fire-at-will:
## Changes introduced
Ports over the `trackCustomPaywallImpression()` from the mobile SDKs. When called, it allows developers to track paywall impression events from their custom paywalls.

### New APIs
Introduces a new `trackCustomPaywallImpression()` function:

```typescript
public trackCustomPaywallImpression(
    params: CustomPaywallImpressionParams = {},
  ): void

export interface CustomPaywallImpressionParams {
  /** An optional identifier for the custom paywall shown to the customer. */
  paywallId?: string;

  /** The offering used to render the custom paywall. */
  offering?: Offering;
}
```

We intentionally omitted the deprecated `offeringId` param that is present in the native SDKs since it is deprecated.

### Testing
- Manually created some custom paywall impression events and viewed them using the RevenueCat MCP.
- Added automated tests

---

> [!NOTE]
> **Low Risk**
> Additive public API and analytics-only path; no purchase or auth changes, with broad test coverage for serialization and attribution edge cases.
> 
> **Overview**
> Adds **`Purchases.trackCustomPaywallImpression()`** so apps can report impressions for **custom** paywalls (parity with native SDKs), using optional `paywallId` and/or an `Offering` for attribution.
> 
> Events are emitted as **`custom_paywall_impression`** (v1) through the existing analytics pipeline: a new event type is queued on `EventsTracker` with a stable **`app_session_id`**, optional placement/targeting from the offering’s `presentedOfferingContext`, and the same flush/keepalive batching as other events.
> 
> When no offering is passed, attribution falls back to a **`cachedCurrentOffering`** updated on `getOfferings()` (including a rule that **keeps the previous current offering** if a filtered fetch drops it). The cache is **cleared on user change**. Calls with no offering still send an unattributed impression.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 96747ab88fd4aa1f1a1452f84ca83e9fdb49a3c9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/main.ts
```diff
@@ -1323,7 +1328,44 @@ export class Purchases {
...
 
-    return await this.getAllOfferings(offeringsResponse, appUserId, params);
+    const offerings = await this.getAllOfferings(
+      offeringsResponse,
+      appUserId,
+      params,
+    );
+    if (appUserId === this._appUserId) {
+      this.cachedCurrentOffering = offerings.current;
+    }
```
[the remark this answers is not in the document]
fire-at-will: Addressed in fd44b1f099e032d0b7ffd6f4f01375fc8a402f5d

src/behavioural-events/events-tracker.ts
```diff
@@ -63,6 +74,8 @@ export default class EventsTracker implements IEventsTracker {
   private readonly workflowContext?: WorkflowContext;
   private isDisposed: boolean = false;
 
+  private static readonly appSessionId = generateUUID();
```
fire-at-will, line 77: This would get cleared out if the developer re-configures the SDK, but I think that's okay!

src/main.ts
```diff
@@ -1323,7 +1328,50 @@ export class Purchases {
...
+  }
+
+  /**
+   * Tracks an impression for a custom paywall.
+   *
+   * Pass the offering used to render the paywall to preserve placement and
+   * targeting attribution. When no offering is passed, the most recently
+   * fetched current offering is used if available.
+   *
+   * @param params Parameters for the custom paywall impression event.
```
rickvdl, line 1359: iOS (and others) also mention that it will be creating an impression per call, and that it should only be called once, for example on iOS
```
/**
 * Each call creates a separate impression event. Call this once per
 * paywall presentation and avoid lifecycle callbacks that may run
 * multiple times for the same display.
 */
```

Maybe we should also mention this point here

rickvdl (approved):
I think this looks good!

src/main.ts
fire-at-will: Good idea! Added in 6cc4d669dd63bb0ba2e28998837e19ecb42a98da

PR #1097, opened 2026-08-27
fix(vega): suppress generic web billing events on vega

fire-at-will:
## Motivation / Description
The JS SDK emits many events that are web-billing specific, and that the backend classifies as RCBilling events. We were incorrectly emitting these events when running on Vega. This caused events to be emitted by Vega apps that got classified as web billing events by the backend.

## Changes introduced
Prevents the SDK from emitting generic web billing events (including sdk_initialized) when the SDK is initialized with an Amazon API key. Paywall events are still emitted, as they aren't classified as web billing events in the backend.

---

> [!NOTE]
> **Low Risk**
> Analytics-only behavior gated on API key prefix; purchase and paywall event paths are explicitly preserved in tests.
> 
> **Overview**
> **Stops generic web billing analytics** (e.g. `sdk_initialized`, external/SDK `web_billing` events) when `EventsTracker` is constructed with an **Amazon API key** (`isAmazonApiKey`), so Vega apps no longer send events the backend misclassifies as RCBilling.
> 
> The guard lives in private `trackEvent` via a new `suppressGenericEvents` flag set in the constructor. **Paywall and custom paywall impression events are unchanged** and still flush to `/v1/events`.
> 
> Tests replace the old Amazon `X-Platform` header expectation with coverage that generic tracking is skipped while paywall paths still post, plus a `Purchases.configure` case asserting no `sdk_initialized` for `amzn_valid_key`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cd13f185e869cad9ccfcbb87549b3e23d6579e68. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/behavioural-events/events-tracker.ts
+++ b/src/behavioural-events/events-tracker.ts
@@ -2,6 +2,7 @@ import { generateUUID } from "../helpers/uuid-helper";
 import { RC_ANALYTICS_ENDPOINT } from "../helpers/constants";
 import { HttpMethods } from "msw";
 import { getHeaders } from "../networking/http-client";
+import { isAmazonApiKey } from "../helpers/api-key-helper";
 import { FlushManager, type FlushOptions } from "./flush-manager";
 import { Logger } from "../helpers/logger";
 import { defaultPurchaseMode, Event, type EventProperties } from "./event";
@@ -70,6 +71,7 @@ export default class EventsTracker implements IEventsTracker {
   private readonly traceId: string;
   private appUserId: string;
   private readonly isSilent: boolean;
+  private readonly suppressGenericEvents: boolean;
   private rcSource: string | null;
   private readonly workflowContext?: WorkflowContext;
   private isDisposed: boolean = false;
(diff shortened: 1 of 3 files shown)


src/behavioural-events/events-tracker.ts
```diff
@@ -165,6 +168,9 @@ export default class EventsTracker implements IEventsTracker {
       Logger.verboseLog("Skipping event tracking, the EventsTracker is silent");
       return;
     }
+    if (this.suppressGenericEvents) {
```
tonidero, line 171: Hmm I would say the proper fix is probably to pass the `store` or something like that in these events instead... Do we plan to do something like that as a follow-up?

src/behavioural-events/events-tracker.ts
```diff
@@ -81,6 +83,7 @@ export default class EventsTracker implements IEventsTracker {
     this.eventsUrl = `${props.httpConfig?.eventsURL ?? RC_ANALYTICS_ENDPOINT}/v1/events`;
     this.appUserId = props.appUserId;
     this.isSilent = props.silent || false;
+    this.suppressGenericEvents = isAmazonApiKey(props.apiKey);
```
tonidero, line 86: FWIW, not sure if we should also have these events in stripe/paddle/test store... But maybe a question for @james-od @nicfix ?

tonidero (commented):
I would love for the web team to take a look at this. On my side just a couple questions 🙏

joshdholtz (approved):

PR #1105, opened 2026-08-31
[WEB-4656] Prepare Apple Pay before RC Billing checkout

nicfix:
Apple Pay must be presented from the original customer click, but the previous `.purchase()` flow waited for RC Billing checkout setup before attempting to show it. Safari can reject the presentation after that asynchronous work.

This adds `prepareForQuickPurchases()` to load and cache package-independent Stripe configuration before the click. When `.purchase({ tryWithApplePay: true })` is called with prepared state, the SDK shows Apple Pay synchronously and starts the normal RC Billing checkout immediately afterwards.

The existing checkout remains the default and is used when preparation is missing, Apple Pay cannot be presented, or the purchase requires the checkout UI. Express Purchase Button behavior is unchanged. The Web Billing demo includes a dedicated Apple Pay-first page.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -1711,6 +1711,7 @@ export class Purchases {
       return null;
     }
 
+    this.preparedQuickPurchaseConfiguration = null;
     Logger.debugLog(
       "Apple Pay first is ready; calling paymentRequest.show() before /start",
     );
@@ -1822,6 +1823,13 @@ export class Purchases {
           }),
         );
         throw PurchasesError.getForPurchasesFlowError(purchaseFlowError);
+      })
+      .finally(() => {
+        void this.prepareForQuickPurchases().catch((error) => {
+          Logger.debugLog(
+            `Could not prepare Apple Pay for another purchase: ${String(error)}`,
+          );
+        });
       });
   }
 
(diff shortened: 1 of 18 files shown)


src/tests/stripe/apple-pay-purchase.test.ts
nicfix, line 1: Probably better to have this test clearly refer to rc_billing instead of stripe in the name.

src/main.ts
```diff
@@ -1537,6 +1557,304 @@ export class Purchases {
...
+      effectiveParams.purchaseOption ?? product.defaultPurchaseOption;
+    const termsAndConditionsUrl = resolveTermsAndConditionsUrl({
+      brandingInfo,
+      termsAndConditionsUrl: effectiveParams.termsAndConditionsUrl,
+    });
+    const skipReasons: string[] = [];
+    if (this.resolveProductChange(effectiveParams)) {
+      skipReasons.push("product change requested");
+    }
+    if (effectiveParams.discountCode) {
```
nicfix, line 1654: We should allow purchasing with discounts through apple pay directly

nicfix:
We will follow a different approach to give priority to Stripe Billing

PR #1121, opened 2026-09-04
Cache getOfferings calls for 10 seconds

lburdock:
#### Customer complaint

The `getCurrentOfferingForPlacement()` function hits `GET /v1/subscribers/{app_user_id}/offerings` on every call, with no cache and no in-flight de-dup. That request carries only the app user id and the SDK resolves the placement client-side, so the responses come back identical. (The `/products` calls after it are fair, those genuinely differ per offering.)

We run 3 placements, so one user session pays for 3 identical `/offerings` fetches at about a second each. Some of those sit on the critical path, with UI held back until they resolve. We plan to put every paywall entry point behind a placement, so that count only goes up from here.

## Changes introduced

The code adds a 10-second in-memory cache for getOfferings/getCurrentOfferingForPlacement responses, plus in-flight request de-duplication so concurrent calls for the same user share one network request. Cache invalidation has been added to all the existing invalidateAllCaches() call sites, plus a new setAttributes invalidation and a public invalidateOfferingsCache() API. 

## Additional comments

---

> [!NOTE]
> **Medium Risk**
> Caching and dedupe touch paywall/placement targeting paths; incorrect invalidation could briefly serve stale offerings, though TTL is short and invalidation hooks are broad.
> 
> **Overview**
> Reduces duplicate **`GET .../offerings`** traffic by caching the raw offerings response in memory for **10 seconds** (per app user) and **deduplicating concurrent** fetches. **`getOfferings`** and **`getCurrentOfferingForPlacement`** now share a private **`getOfferingsResponse`** path instead of calling the backend directly.
> 
> Adds public **`invalidateOfferingsCache()`** and clears offerings cache (and in-flight requests) when targeting-relevant state changes—**`setAttributes`**, **`Purchases.setPlatformInfo`**, user switches, and after purchases / simulated receipts—via **`invalidateRequestDataCaches`**. **`getOfferings`** offering filters apply on a copy so a filtered fetch does not corrupt the shared cached payload (important when mixing filtered **`getOfferings`** with placement lookups).
> 
> **`InMemoryCache`** gains per-type TTLs (offerings vs virtual currencies). Failed offerings responses are not cached; stale in-flight requests cannot repopulate the cache after invalidation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2c9b520898730df05cb4f9840d037ff68ae1ac74. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


api-report/purchases-js.api.md
```diff
@@ -563,6 +492,7 @@ export class Purchases {
     static getSharedInstance(): Purchases;
     getVirtualCurrencies(): Promise<VirtualCurrencies>;
     identifyUser(appUserId: string): Promise<IdentifyResult>;
+    invalidateOfferingsCache(): void;
```
james-od, line 495: I think @internal is probably fine/preferable?

src/main.ts
```diff
@@ -1479,6 +1492,58 @@ export class Purchases {
...
+    if (appUserId === undefined) {
+      this.offeringsRequests.clear();
+    } else {
+      this.offeringsRequests.delete(appUserId);
+    }
+  }
+
+  private invalidateAllCaches(): void {
+    this.offeringsRequests.clear();
+    this.inMemoryCache.invalidateAllCaches();
```
james-od, line 1544: Should we also set `this.cachedCurrentOffering = null` in `invalidateAllCaches` and `clearOfferingsCache`

api-report/purchases-js.api.md
lburdock: I made it public so it would be consistent with the existing invalidateVirtualCurrenciesCache() method for the virtual currencies cache. Since offerings can also change externally through dashboard configuration or targeting, I thought it would be safer to allow developers to force an immediate refresh if needed. Would it be better to start with it as internal and make it public later if needed?

src/main.ts
lburdock: `cachedCurrentOffering` is mainly there for impression attribution, not to serve cached requests. If we cleared it during normal cache invalidation, impressions could lose their offering and targeting context, so we only clear it when the app user changes.

api-report/purchases-js.api.md
james-od: Ah ok - yeah I think that makes sense

james-od (approved):

PR #1123, opened 2026-09-07
Make express checkout optional

lburdock:
## Motivation / Description

We've had a few instances where the Stripe Payment Element is slow to load. This decouples the payment form's "ready to display" state from the Express Checkout element's readiness/error state, so a slow or failing Express Checkout no longer blocks or breaks the main payment form.

## Additional comments

---

> [!NOTE]
> **Medium Risk**
> Changes checkout loading and error-handling paths for Stripe Express Checkout; incorrect gating could show the form too early or hide wallets incorrectly, but core payment/link/address fatal errors are unchanged.
> 
> **Overview**
> **Express Checkout no longer blocks the main payment form.** The loading spinner and form visibility now wait only on Link Authentication, Payment, and Address (when required)—not on Express Checkout readiness.
> 
> When Express Checkout fails to load or initialize, the UI **hides the wallet section** and logs a debug message instead of surfacing a gateway error or calling the parent `onError`. Card checkout and analytics (e.g. `CheckoutPaymentFormImpression`) still proceed.
> 
> The **“OR PAY BY CARD”** divider appears only after Express Checkout fires `ready` with available payment methods, so users don’t see a separator before wallets are actually offered.
> 
> Tests were expanded to cover non-blocking form display, separator timing, nonfatal Express Checkout errors, and updated Stripe Elements debug log ordering.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dee558060b636e99eade520b03ec981b204ed4c1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/ui/molecules/stripe-elements.svelte
+++ b/src/ui/molecules/stripe-elements.svelte
@@ -40,7 +40,6 @@
     expressCheckoutOptions?: StripeExpressCheckoutConfiguration;
     brandingInfo: BrandingInfoResponse | null;
     forceEnableWalletMethods: boolean;
-    disableExpressCheckout?: boolean;
     skipEmail: boolean;
     onLoadingComplete: () => void;
     onError: (error: StripeServiceError) => void;
@@ -58,7 +57,7 @@
       paymentMethod: string,
       emailValue: string,
     ) => void;
-    expressCheckoutPermittedByConsent?: boolean;
+    allowExpressCheckout?: boolean;
   }
 
   let {
@@ -70,7 +69,6 @@
     expressCheckoutOptions,
     brandingInfo,
     forceEnableWalletMethods,
-    disableExpressCheckout = false,
     skipEmail,
     onLoadingComplete,
     onError,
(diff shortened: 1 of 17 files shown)


src/ui/molecules/stripe-elements.svelte
```diff
@@ -159,7 +161,7 @@
     if (
       emailElementReadyForSubmission &&
       paymentElementReadyForSubmission &&
-      expressCheckoutElementReadyForSubmission &&
+      (disableExpressCheckout || expressCheckoutElementReadyForSubmission) &&
```
james-od: I'm not sure about this, adding a permanent extra prop to the API for some niche debugging feels awkward, and I don't think we'd learn much? I'd maybe be more in favour of first testing whether dropping `expressCheckoutElementReadyForSubmission` entirely here is worth looking at - which I think is our (not Stripe's) block on displaying the checkout form?

lburdock: Thanks for this suggestion! I created a PR to add logging for Stripe Element loading (https://github.com/RevenueCat/purchases-js/pull/1132), and then updated this PR so that we display the form even if Express Checkout isn't ready yet. The Express Checkout buttons will appear when the element is ready.

james-od (approved):

PR #1138, opened 2026-09-10
other(vega): move vega code to separate package

fire-at-will:
## Motivation / Description
Currently, the Vega code is contained in the main `purchases-js` package, and all vega dependencies are declared as optional dependencies on the main package. The Vega code is only used if `purchases-js` is imported through a unique vega entrypoint, which wires up all of the Vega code and imports the dependencies at runtime:

```typescript
import {Purchases} from '@revenuecat/purchases-js/vega'
```

This works, but has a few downsides:
- Since the Vega dependencies are declared as peer dependencies on the main module, it's possible that a web developer could accidentally pull in the vega dependencies at runtime, which would lead to a crash on the web since the vega dependencies rely on Flow, which is only available with React Native.
- It would be possible for future SDK contributors to accidentally import one of the Vega dependencies into the main part of the SDK without realizing it, causing web developers to bring the Vega dependencies into their builds.
- Some developers could be turned away by the core `purchases-js` module depending on react native/other native dependencies
- The entry-point method doesn't feel super scalable if we were to add more stores to `purchases-js` in the future
- It required us to import and somewhat obfuscate imports of Vega/RN dependencies so that they wouldn't be detected and imported when the package bundles were created. This works, but is difficult to read, maintain, and may be prone to error in the future.

Instead, this PR migrates the Vega implementation to a pattern used in `purchases-android`, `purchases-hybrid-common`, and `react-native-purchases`, where the implementation & dependencies for additional stores are contained in a supplemental module, which is consumed by the primary module at runtime.

The new repo structure looks like this:
```
purchases-js/
├── package.json
├── src/                # Existing SDK and shared implementation
├── dist/               # Existing SDK build output
└── vega/
    ├── package.json
    ├── src/            # Vega-specific implementation
    └── dist/           # Vega build output
```

This is a breaking change for developers using the Vega beta; we will coordinate with them to help them with the migration process.

## Changes introduced

1. Created a new `purchases-js-vega` package under `/vega`. It has its own manifest, entry point, TypeScript configuration, build configuration, and `dist/` folder.
2. Moved the Vega implementation and tests to `/vega/src`. Amazon billing, receipt caching, and the IAP, filesystem, compatibility, and React Native loaders now live there. Most of this was relocation and import-path changes.
3. Added billing-provider injection to the main module. `main.ts` no longer imports or constructs the `AmazonBillingWrapper` directly. The Vega entry point in the new module registers a factory that receives the backend, API key, and current-user callbacks. Registration happens automatically when the app imports the new Vega package.
4. Moved dependency declarations. The Amazon and React Native dependencies are removed from the primary package and now placed in the Vega module’s manifest. They are declared as required peer dependencies since they are bundled with the Vega operating system.
5. Separated build and packaging boundaries. Removed the old `/vega` export and Metro shim. The web package includes only its own output, while the Vega module generates its own separate JavaScript and declarations. Added workspace configuration, declaration exclusions, generated-file lint exclusions, and package-boundary checks.
6. Updated the Vega PurchaseTester app. Its imports now use `@revenuecat/purchases-js-vega`, with a local `file:../../../vega` dependency. Both example apps retain independent dependency installations and lockfiles.

## TODO
This PR intentionally omits a few things that will be addressed in follow-up PRs:
- The Vega implementation previously imported RN/Vega dependencies at runtime using a loader pattern to avoid having these dependencies get bundled in when the package was built. This is no longer needed since we have a dedicated vega module. We will remove the complicated loader code in a follow up PR. (implemented here: https://github.com/RevenueCat/purchases-js/pull/1141)
- The vega module isn't currently published by CI. We'll add support for this in CI in a future PR.
- We will need to create a new API extractor for this package so we can monitor API changes over time.

---

> [!NOTE]
> **Medium Risk**
> Breaking change for Vega consumers and changes how Amazon API keys initialize billing; web-only flows are tightened but Amazon configure/purchase paths need careful regression testing.
> 
> **Overview**
> **Splits Amazon Vega / Appstore support out of `@revenuecat/purchases-js` into a dedicated workspace package `@revenuecat/purchases-js-vega`.** Vega apps must switch from `@revenuecat/purchases-js/vega` to `@revenuecat/purchases-js-vega` (breaking for the Vega beta).
> 
> The main SDK drops the `./vega` export, Metro shim, and optional Amazon/React Native peer dependencies. Amazon billing code and its build output live under `vega/` with its own `package.json`, Vite build, and `dist/`. Root `build` delegates to `pnpm --filter @revenuecat/purchases-js-vega build`; `verify-entry-point-boundaries` now checks separate package manifests and artifact paths.
> 
> **Runtime wiring uses a billing-provider registry** instead of `isVegaEntryPoint()`. Importing the Vega package registers loaders and an `AmazonBillingWrapper` factory via `registerBillingProvider`. Configuring with an Amazon API key on the web package alone fails with a clear error directing developers to the Vega package.
> 
> PurchaseTester and lockfiles are updated to depend on `file:../../../vega`. Example apps get empty `pnpm-workspace.yaml` files so they do not join the SDK workspace.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a4734b8ba52b46cfc841f58b320592c7b4296116. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/helpers/billing-provider.ts
```diff
@@ -0,0 +1,42 @@
+import type { Backend } from "../networking/backend";
+import type { BillingWrapper } from "./billing-wrapper";
+
+/** Context supplied when a platform package creates its billing implementation.
+ *
+ * @internal
+ */
+export interface BillingProviderContext {
+  backend: Backend;
```
tonidero, line 5: Hmm the `Backend` is not exported right? Not sure if this would work then...

src/main.ts
```diff
@@ -641,12 +641,12 @@ export class Purchases {
       eventName: SDKEventName.SDKInitialized,
     });
     if (isAmazonApiKey(this._API_KEY)) {
-      this.amazonBillingWrapper = new AmazonBillingWrapper(
-        this.backend,
-        this._API_KEY,
-        () => this._appUserId,
-        () => this.isAnonymous(),
-      );
+      this.amazonBillingWrapper = getBillingProvider()!.createBillingWrapper({
```
tonidero: Hmm so if someone uses an amazon API key but the wrong library, it would fail here? Maybe we should provide a more explicit error message?

eslint.config.js
```diff
@@ -4,6 +4,7 @@ import tseslint from "typescript-eslint";
 
 /** @type {import('eslint').Linter.Config[]} */
 export default [
+  { ignores: ["src/vega/dist/**"] },
```
tonidero, line 7: Hmm so, I guess this ignores the `dist` folder for the linter, which kinda makes sense... But I wonder if we should also add the `dist` of the "main" library? Or if not, why do we need to add it only for one?

scripts/verify-entry-point-boundaries.mjs
```diff
@@ -12,8 +11,8 @@ const amazonModule = "@amazon-devices/keplerscript-appstore-iap-lib";
 const fileSystemModule = "@amazon-devices/kepler-file-system";
 const defaultArtifacts = ["dist/Purchases.es.js", "dist/Purchases.umd.js"];
 const vegaArtifacts = [
-  "dist/Purchases.vega.es.js",
-  "dist/Purchases.vega.umd.js",
+  "src/vega/dist/Purchases.vega.es.js",
+  "src/vega/dist/Purchases.vega.umd.js",
```
tonidero, line 15: Hmm so... something that I find weird is that now the `dist` folder lives alongside the `src`... IMO, the `dist` should ideally be on a folder outside `src`... Not sure if that's easy to do though, since we already have a root `dist`...

src/vega/package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "@revenuecat/purchases-js-vega",
```
tonidero, line 2: [Not related to this line]
In general, it feels weird to have a package like this within a folder inside `src`... In my mind, I would expect these to be separate from `src`, so feels it's out of place... Maybe we can put all these inside a separate folder at the root?

tsconfig.json
```diff
@@ -30,7 +30,7 @@
     "./src/**/*.svelte",
     "src/main.test.tsx",
   ],
-  "exclude": ["node_modules", "./src/tests", "./**/*.test.ts"],
+  "exclude": ["node_modules", "./src/tests", "./src/vega", "./**/*.test.ts"],
```
tonidero, line 33: Since we're excluding vega here, do we need separate CI jobs to build, pass all the linters and ideally some tests and E2E tests on the vega library as a standalone?

tonidero (commented):
Mostly some questions about whether the current location is the best, since it feels confusing to me to have a nested library within the `src` folder... Also, would love if @james-od or someone from the web team can also take a look 🙇

src/helpers/billing-provider.ts
tonidero: Ahh nvm... I guess we're not actually exporting this outside, it's just have the class internally... That `@internal` annotation confused me, maybe we can remove those since they shouldn't be part of the public API?

tsconfig.json
tonidero: Another thing we should do is add separate api extractor files for the Vega library, to make sure we keep an eye out on the public API of that library

src/helpers/billing-provider.ts
fire-at-will: Makes sense, removed in the comments in 0250b0ce5af666e4d9a76d9a005167a12b2d90db

src/main.ts
fire-at-will: I agree with the concern. Luckily, this should already been handled - the `validateApiKey()` function throws if you use an Amazon API key and the billing provider isn't configured. The error message it uses is: `Using the Amazon Appstore requires usage of the @revenuecat/purchases-js-vega package.`. We also have a unit test to verify this: `requires the Vega package for an Amazon API key` in `main.test.ts`.

That said, force unwrapping here isn't ideal, it's possible that the code gets refactored in the future in a way that could expose this callsite. I've removed the force unwrapping and added in a more descriptive error message here: 4bfe3b0f37afdf36a4cdbe11f102285cecca9ef7

vega/package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "@revenuecat/purchases-js-vega",
```
fire-at-will: Moved the package to `/vega`! The repo structure now looks like this:

```
purchases-js/
├── package.json
├── src/                # Existing SDK and shared implementation
├── dist/               # Existing SDK build output
└── vega/
    ├── package.json
    ├── src/            # Vega-specific implementation
    └── dist/           # Vega build output
```

eslint.config.js
fire-at-will: we needed this at the time since vega's `dist` was under the main `src` directory, which was being linted. However, now that this will go under `vega/dist`, this isn't needed. removed in a4734b8ba52b46cfc841f58b320592c7b4296116

tsconfig.json
fire-at-will: Luckily, the linters/unit test runners also already cover the vega files in `/vega`. We do need to add a separate API extractor - I'll do that in a follow-up PR.

src/main.ts
```diff
@@ -641,12 +641,19 @@ export class Purchases {
       eventName: SDKEventName.SDKInitialized,
     });
     if (isAmazonApiKey(this._API_KEY)) {
```
lburdock, line 643: Should we split the current Purchases implementation into shared and platform-specific pieces? Right now Vega re-exports the entire Web implementation (including Web Billing, Paddle, Stripe, and the Svelte UI) into the Vega package, even if they aren't used.

A PurchasesCore class could contain the shared functionality (offerings, customer info, identity, etc). A Web Purchases class could add Web Billing, Paddle, and Stripe, while a Vega Purchases class could add the Amazon-specific behavior.

fire-at-will: That's a _great_ idea! None of those dependencies will be used in a Vega app, so it'd be nice to not bundle the dependencies in it. That said, I'm assuming that would be a pretty good sized refactor since `main.ts` has a good bit of the SDK's logic in it, right? I'll take a look at it and see what it might look like..

lburdock: Yeah, it would be pretty large. Extracting the reusable core logic from the web-specific logic could be its own PR, separate from the Vega work entirely. What's the urgency level for the Vega work? This could potentially be a quick follow-up once some of this work is merged.

fire-at-will: I think that ideally, we'd be able to merge the vega work into `main` sometime late this week or mid next week. We have a few customers that we'd like to onboard, and we'd rather have them adopt from an actual release instead of a beta release.

I could see us splitting it up like this:
- `purchases-js`: Web implementation. Not named `purchases-js-web` to avoid a breaking change for existing users
- `purchases-js-vega`: Vega implementation
- `purchases-js-core`: Shared core implementation, with the backend, etc.

While not ideal, I don't think that the extra ~1mb from those dependencies is a blocker today, since we already had that in the entrypoint solution, and the 1mb download cost is incurred when the app is installed from the Amazon Appstore, not on each website load.

One awkward bit of shipping as is and then removing the dependencies later is that there are a two web only functions that we'd need to remove, deprecate, or something. Paywalls aren't implemented for Vega, so these functions don't work on Vega today, so maybe this isn't a huge deal? Anyways, we should probably figure out something to do with them:
- presentPaywall()
- presentExpressPurchaseButton()

fire-at-will: I had a discussion with @MarkVillacampa about this via Slack DMs, and we're okay with the extra dependencies being included in the Vega builds for now, and can follow up on removing them in the future if needed :)

lburdock: Yeah, I think we could get away with these for now:
* Add a guard to presentPaywall, similar to the one in presentExpressPurchaseButton, that throws for Amazon API keys
* Document both methods as unsupported for Vega

lburdock (approved):

PR #1141, opened 2026-09-11
other(vega): remove dependency loaders

fire-at-will:
## Motivation / Description
When the Vega implementation was wired up using an entry point in the main `purchases-js` package, we couldn't directly import Vega dependencies to avoid having them get bundled into the SDK's package and include them in web/hybrid builds. We got around this issue by dynamically loading the dependencies and injecting them in where needed. This worked, but was verbose and difficult to reason about.

Now that the Vega implementation is housed in its own separate `purchases-js-vega` package, we no longer need to do this, and can import the dependencies normally.

## Changes introduced

This PR removes the Vega dependency loaders and imports the dependencies directly. This not only removes the Loader classes, but also simplifies the `AmazonBillingWrapper` and `VegaDeviceCache` classes quite a bit.

## Additional comments

---

> [!NOTE]
> **Medium Risk**
> Touches Amazon purchase/sync/restore and receipt caching paths and package boundary guarantees, though behavior is intended to be equivalent with simpler wiring.
> 
> **Overview**
> Moves the **Vega / Amazon Appstore** implementation from `src/vega` to a top-level **`vega/`** workspace package and stops using runtime **dependency loaders** now that `@revenuecat/purchases-js-vega` is separate from the web SDK.
> 
> **Vega package:** Deletes the IAP, Kepler File System, compatibility, and React Native `AppState` loader modules. `AmazonBillingWrapper` and `VegaDeviceCache` import `@amazon-devices/*` and `react-native` directly; the Vega entry only **registers** the billing provider and re-exports the main SDK. Build, lint, workspace, demo `file:` paths, and TypeScript excludes are updated for `vega/`.
> 
> **Core SDK:** Amazon initialization no longer uses `getBillingProvider()!`; it throws the same configuration error when the Vega package is missing. User-facing Amazon messages are slightly reworded. `billing-provider` JSDoc is trimmed.
> 
> **Tooling & tests:** `verify-entry-point-boundaries.mjs` checks more native deps (including `react-native`), Vega/web artifact paths under `vega/dist`, quoted module references, and `.d.ts` bundles. Main `main.test.ts` uses a **`BillingWrapper` mock** instead of `AmazonBillingWrapper`; Vega tests mock native modules and add AppState edge-case coverage.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c4a1e97fe2b3052bbeb671fb8d856c234ff0fa9e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/vega/amazon/vega-device-cache.ts
```diff
@@ -87,11 +76,11 @@ export class VegaDeviceCache {
   }
 
   private async doesCacheFileExist(
-    fileSystem: KeplerFileSystem,
+    fileSystem: typeof KeplerFileSystem,
   ): Promise<boolean> {
     // `exists` arrived in Kepler File System 0.0.7. On older Vega OS versions,
     // use a read attempt to determine whether the cache file is present.
-    if (this.isKeplerFileSystemExistsSupported()) {
+    if (isPresentOnOS("@amazon-devices/kepler-file-system", "0.0.7")) {
```
tonidero, line 83: I think this version number should be extracted?

tonidero (approved):
Indeed much cleaner! I think this makes sense!

vega/src/amazon/vega-device-cache.ts
```diff
@@ -87,11 +76,11 @@ export class VegaDeviceCache {
   }
 
   private async doesCacheFileExist(
-    fileSystem: KeplerFileSystem,
+    fileSystem: typeof KeplerFileSystem,
   ): Promise<boolean> {
     // `exists` arrived in Kepler File System 0.0.7. On older Vega OS versions,
     // use a read attempt to determine whether the cache file is present.
-    if (this.isKeplerFileSystemExistsSupported()) {
+    if (isPresentOnOS("@amazon-devices/kepler-file-system", "0.0.7")) {
```
fire-at-will: Good idea! Extracted the version number + dependency name to constants in c4a1e97fe2b3052bbeb671fb8d856c234ff0fa9e

PR #1143, opened 2026-09-11
FUN-2346 Paddle package discounts: PricePreview and discountId at checkout

francocorreasosa:
FUN-2346

## Motivation / Description

Funnels can now configure a Paddle discount per package component (`paddle_discount.discount_id` in the paywall JSON). The host needs two things from the SDK: the discounted price before checkout, so the paywall can render `product.offer_price` and the `promo_offer` condition, and a way to hand that discount to Paddle checkout for the selected package. Existing discount codes go through the RC backend, which knows nothing about Paddle discounts, so the preview has to come from Paddle.js.

## Changes introduced

- `PaddleService.initializeForPreview(productId, purchaseOption)` initializes Paddle.js from the `/checkout/prepare` client token, without creating a transaction.
- `PaddleService.previewDiscount({ priceId, discountId, currencyCode })` runs `Paddle.PricePreview` and maps the first line item to a `DiscountPhase` (`src/paddle/paddle-discount-preview.ts`). Minor-unit amounts convert to micros using the currency's fraction digits; `recur` / `maximumRecurringIntervals` map to `forever` / `time_window`. Returns `null` when Paddle applies no discount.
- `Purchases.applyPaddleDiscountsToOffering(offering, discountIdsByPackage)` (`@internal`, for Workflows): returns a copy of the offering where matching packages carry the previewed discount on their default option, so `buildVariablesPerPackage` / `buildInfoPerPackage` surface it unchanged. No-op for non-Paddle keys; per-package failures are logged and leave that package as fetched.
- `PurchaseParams.discountId` (`@experimental`, Paddle only) threaded through `performPaddlePurchase` -> `PaddlePurchasesUi` -> `Checkout.open`. When set it replaces `discountCode`, since Paddle accepts one or the other.
- `getCurrencyFractionDigits` and `getPriceForCurrency` exported for reuse.

Tests: `paddle-discount-preview.test.ts` (amount conversion, phase mapping, offering cloning feeding the variable helpers), `paddle-service.test.ts` (checkout options, `initializeForPreview`, `previewDiscount`), `purchase.paddle-discounts.test.ts` (`applyPaddleDiscountsToOffering` end to end with msw + mocked Paddle.js). Full suite: 943 passed. `api-report` regenerated (adds `PurchaseParams.discountId`).

## Linear ticket (if any)

FUN-2346

## Additional comments

Companion PRs: Backend API (schema), purchases-ui-js (type), dashboard (editor), rc-workflows (runtime). rc-workflows needs a release of this package to bump to.

---

> [!NOTE]
> **Medium Risk**
> Changes Paddle checkout discount wiring and customer-facing offer prices; failures degrade to undiscounted packages but incorrect previews could mislead before purchase.
> 
> **Overview**
> Adds **Paddle-configured package discounts** so paywalls can show promo pricing before checkout and pass the same discount into Paddle Checkout.
> 
> A new internal **`Purchases.applyPaddleDiscountsToOffering`** uses **`Paddle.PricePreview`** (via **`initializeForPreview`** / **`previewDiscount`** on **`PaddleService`**) to build a **`DiscountPhase`** per package and clone the offering so existing paywall helpers expose **`product.offer_price`** and **`promo_offer`**. Preview math scales off the displayed catalog price so tax-inclusive totals stay aligned with checkout.
> 
> Checkout now accepts an internal **`PurchaseParams.discountId`** (`dsc_...`), threaded through the Paddle UI; **`buildPaddleCheckoutOptions`** prefers **`discountId`** over **`discountCode`** because Paddle allows only one. **`getPriceForCurrency`** and **`getCurrencyFractionDigits`** are exported for the preview mapper. Tests cover preview mapping, offering patching, service APIs, and the **`applyPaddleDiscountsToOffering`** integration path.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3ee5ad345cb5022895bdd7b569e614b7d8cec524. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/entities/purchase-params.ts
+++ b/src/entities/purchase-params.ts
@@ -160,6 +160,14 @@ export interface PurchaseParams {
 
   /**
    * @experimental
+   * Paddle discount id (`dsc_...`) to apply at checkout. Paddle only.
+   * Takes precedence over {@link PurchaseParams.discountCode}, since Paddle
+   * Checkout accepts one or the other but not both.
+   */
+  discountId?: string;
+
+  /**
+   * @experimental
    * Called when the applied discount code changes in the Web Billing checkout.
    * This can be used by host applications to keep external state, such as the URL,
    * in sync with the checkout.
--- a/src/paddle/paddle-discount-preview.ts
+++ b/src/paddle/paddle-discount-preview.ts
@@ -1,0 +1,222 @@
+import type { PricePreviewResponse } from "@paddle/paddle-js";
+import {
+  type DiscountPhase,
+  getPriceForCurrency,
+  type NonSubscriptionOption,
+  type Offering,
+  type Package,
+  type Product,
+  ProductType,
+  type SubscriptionOption,
+} from "../entities/offerings";
+import { parseISODuration, type Period } from "../helpers/duration-helper";
+import { getCurrencyFractionDigits } from "../helpers/price-labels";
+
+type PricePreviewLineItem =
+  PricePreviewResponse["data"]["details"]["lineItems"][number];
+
+/**
+ * Paddle returns amounts as strings in the currency's minor unit (e.g. "1999"
+ * for $19.99, "1500" for ¥1500). Convert to RevenueCat micros.
+ */
+export function paddleAmountToMicros(amount: string, currency: string): number {
+  const minorUnits = Number(amount);
+  if (!Number.isFinite(minorUnits)) {
+    return 0;
+  }
+  const fractionDigits = getCurrencyFractionDigits(currency);
+  return Math.round(minorUnits * 10 ** (6 - fractionDigits));
+}
+
+function toPeriod(
+  billingCycle: PricePreviewLineItem["price"]["billingCycle"],
+  fallbackPeriodDuration: string | null,
+): { period: Period | null; periodDuration: string | null } {
+  if (billingCycle) {
+    const unitLetter = { day: "D", week: "W", month: "M", year: "Y" }[
+      billingCycle.interval
+    ];
+    const periodDuration = `P${billingCycle.frequency}${unitLetter}`;
+    return { period: parseISODuration(periodDuration), periodDuration };
+  }
+  if (fallbackPeriodDuration) {
+    return {
+      period: parseISODuration(fallbackPeriodDuration),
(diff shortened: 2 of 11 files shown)


src/entities/purchase-params.ts
```diff
@@ -158,6 +158,14 @@ export interface PurchaseParams {
    */
   discountCode?: string;
 
+  /**
+   * @experimental
```
jholwill: what is the difference here between @experimental and @internal?

I have generally only see @internal be used when we are 100% sure this is the final API shape / rolling out behind a FF?

jholwill (approved):

src/entities/purchase-params.ts
francocorreasosa: ahh good catch, will change

francocorreasosa: Changed to `@internal` in 5fd4c51e. It only makes sense together with `Purchases.applyPaddleDiscountsToOffering`, which is already `@internal` for the rc-workflows host, so the two now match.

src/paddle/paddle-discount-preview.ts
```diff
@@ -0,0 +1,243 @@
...
+    if (subscriptionOptions[defaultOption.id]) {
+      subscriptionOptions[defaultOption.id] = defaultOption;
+    }
+    return {
+      ...product,
+      defaultPurchaseOption: defaultOption,
+      defaultSubscriptionOption: defaultOption,
+      subscriptionOptions,
+      discountPhase: discount,
+    };
```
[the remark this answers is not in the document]
francocorreasosa: Keeping trial and intro on purpose. Paddle does not discount the trial charge; the discount is applied to the first billing period after the trial (https://developer.paddle.com/build/products/offer-discounts-promotions-coupons#trials), so both offers are really present at checkout. Dropping the trial from the product would hide it on the paywall while Paddle still grants it. `setOfferVariables` already picks the discount as the primary offer when both exist, so the copy does not stack. Documented in `withPaddleDiscountOnProduct` (5fd4c51e).

PR #1145, opened 2026-09-14
Ignore bundler artifacts so the dependency-update lane can run

facumenzella:
### Motivation

The `dependency-update` job fails before it does anything:

```
$ git status --porcelain
?? .bundle/
?? vendor/
[!] Git repository is dirty!
```

The job config already avoids `install-dependencies` for the same reason ("it dirties the tree"), it just didn't account for the gem step doing it too.

Noticed this while testing the automatic bump from purchases-ui-js

So `dependency-update` has exactly one execution in its history, and this is what it hit. Nothing here was a regression, both halves of the automation shipped untested.

---

> [!NOTE]
> **Low Risk**
> Only updates `.gitignore`; no runtime or application logic changes.
> 
> **Overview**
> Adds **`.bundle/`** and **`vendor/`** to `.gitignore` so Bundler output from CI’s `install-gem-unix-dependencies` step is not treated as untracked files.
> 
> This unblocks the **`dependency-update`** pipeline: the `open_pr_upgrading_dependencies` fastlane lane calls `ensure_git_status_clean`, which was failing when those vendored gem directories appeared after the gem install step (similar to why `install-dependencies` is already skipped for that job).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 564c20e4ec694023299a7ca9940dfc25a5f669cd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


AlvaroBrey (approved):
Thank you!

PR #1163, opened 2026-09-16
Show discount code prices in Web SDK paywalls

lburdock:
## Motivation / Description

`presentPaywall({ offering, discountCode })` only applies the discount code at checkout, leaving the paywall with undiscounted product prices and `hasPromoOffer` set to `false`. 

## Changes introduced

This code refetches the products when a discount code is present so that the discounted price is displayed on the paywall.

| Paywall Before Changes  | Paywall After Changes |
| --- | --- |
| Screenshot 2026-09-16 at 9 49 10 AM | Screenshot 2026-09-16 at 9 51 17 AM |

## Additional comments

---

> [!NOTE]
> **Medium Risk**
> Changes core paywall offering resolution and displayed prices tied to checkout purchase options; failures degrade gracefully but incorrect discount merging could misprice purchases.
> 
> **Overview**
> **`presentPaywall`** now loads **discounted product pricing** when a **`discountCode`** is set, so the paywall shows promo prices and **`hasPromoOffer`** instead of only applying the code at checkout.
> 
> If the caller passes an **`offering`**, the SDK calls **`getProducts`** with that code (and currency when uniform) and merges results via **`replaceOfferingProducts`**; failures or missing products keep the original offering. Without an offering, **`getOfferings`** is invoked with **`discountCode`** on the current offering.
> 
> **`offerings-parser`** adds shared **`mapOfferingPackages`** for placement enrichment and product replacement. The webbilling demo accepts **`paywallDiscountCode`** as a query param. Tests cover discounted paywall rendering, fallbacks, and **`offer_id`** on checkout start for discounted purchase options.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 87151a1a9b20679fec568e126b0865fcc1da6718. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


dalshamaa (approved):

src/main.ts
```diff
@@ -1631,6 +1645,44 @@ export class Purchases {
...
+        this._appUserId,
+        productIds,
+        currency,
+        discountCode,
+      );
+    } catch (error) {
+      Logger.warnLog(
+        `Failed to refresh paywall products with discount pricing: ${String(error)}`,
+      );
+      return offering;
```
dalshamaa, line 1760: I signed off then realized I had a question. should we be explicit here if we fail to get the discounted price (vs. just showing the user the full price without any errors)?

lburdock: I didn't throw an error here because doing so would prevent the paywall from opening. We still pass the discount code to checkout. If the code is successfully applied, checkout shows the discounted price. If not, checkout displays the discount error.

Falling back to the original price on the paywall isn't ideal, but it avoids preventing the user from making a purchase.

PR #1164, opened 2026-09-16
other: move sync/restore purchases to vega module

fire-at-will:
## Motivation / Description
`restorePurchases()` and `syncPurchases()` are currently Vega only APIs, but they were exposed by the main `purchases-js` package. Now that we have a separate `purchases-js-vega` package, we no longer need to expose these functions to web users.

This PR moves these functions to the Vega package so the public APIs better reflect where these methods are supported.

This is not a breaking change since these functions are not yet merged into `main`

## Changes introduced
- Added a Vega-specific Purchases subclass with `restorePurchases()` and `syncPurchases()` functions
- Removed both methods and their result types from the main `purchases-js` public API
- Added an internal factory for creating the appropriate Purchases implementation and billing wrapper. This allows us to inject the Vega Purchases instance at runtime when the Vega package is used
- Removed the previous billing provider registration mechanism, which is no longer needed
- Moved the relevant tests into the Vega test suite
- Kept regular web tests using the default Purchases implementation without a factory
- Updated the generated API report

---

> [!NOTE]
> **Medium Risk**
> Changes SDK initialization and public API boundaries for Amazon/Vega; misconfigured imports could leave Amazon apps without restore/sync or wrong `Purchases` typing, though core purchase/offering paths are largely unchanged.
> 
> **Overview**
> **Amazon-only** `restorePurchases()` and `syncPurchases()` (and their result types) are removed from the main `@revenuecat/purchases-js` public API and live on a **Vega `Purchases` subclass** that delegates to the billing wrapper.
> 
> Core SDK wiring is refactored: the old **billing provider** registry is replaced by an internal **`PurchasesFactory`** that can validate API keys, construct the `Purchases` instance, and create the store **`BillingWrapper`**. Vega registers this factory on import so `configure()` returns the Vega class with restore/sync; plain web usage still gets the default core `Purchases`, and Amazon keys without Vega still fail with the existing “use purchases-js-vega” error.
> 
> Supporting renames/generalization include `amazonBillingWrapper` → `billingWrapper`, a **protected** constructor and `unwrappedBillingWrapper()` for subclassing, and tests moved from core to Vega for sync/restore behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bbae3db96ebbc02083bcbaf2b2f273e8c0675926. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


fire-at-will:
@claude please review

src/main.ts
```diff
@@ -364,7 +360,7 @@ export class Purchases {
   private readonly inMemoryCache: InMemoryCache;
 
   /** @internal */
-  private readonly amazonBillingWrapper: BillingWrapper | null = null;
+  private readonly billingWrapper: BillingWrapper | null = null;
```
fire-at-will, line 363: Renamed to make things more straightforward in the future in case we introduce BillingWrappers for additional stores, or port existing web stores to this format

src/main.ts
```diff
@@ -644,17 +649,19 @@ export class Purchases {
...
-      finalFlags,
-      subscriberToken,
-      brandingAppearanceOverride,
-      context,
-      trace_id,
-    );
+    Purchases.instance?.billingWrapper?.close();
+    Purchases.instance = purchasesFactory
+      ? purchasesFactory.createPurchases(config)
+      : new Purchases(
```
tonidero, line 655: Should we have a "defaultPurchasesFactory" that does this, so we don't need to be splitting the logic like this?

vega/src/index.ts
```diff
@@ -12,8 +15,11 @@ registerBillingProvider({
       );
     }
   },
+  createPurchases: (config) => new Purchases(config),
   createBillingWrapper: ({ backend, apiKey, getAppUserId, getIsAnonymous }) =>
     new AmazonBillingWrapper(backend, apiKey, getAppUserId, getIsAnonymous),
 });
 
 export * from "../../src/main";
+export { Purchases } from "./purchases";
```
tonidero, line 24: Hmm so I'm thinking... the line above `export * from "../../src/main";` already exports a version of `Purchases`. And this line exports a different one... I'm not sure how the users  of the SDK respond to that TBH... but maybe we should confirm they get the expected APIs? (probably we should have the api extractor job for the vega library setup 🙏

tonidero (approved):
Mostly a couple of questions... But not blockers!

PR #1170, opened 2026-09-16
feat: vega API checks + CI validation

fire-at-will:
## Changes introduced
Introduces an API extractor, baseline, and CI check for the new Vega module.
 
Manually tested that the CI checks work by temporarily introducing a new function to the Vega-only Purchases class, and it failed as expected :)

Screenshot 2026-09-16 at 2 25 49 PM

### API Extractor Warnings
The API extractor generates warnings for links to functions contained in the base Purchases class, like this:
```
Warning: dist/Purchases.vega.es.d.ts:390:1 - (ae-unresolved-link) The @link reference could not be resolved: No member was found with name "getOfferings"
```

However, this seems to just be an issue with the API extractor, and has been discussed by other users here: https://github.com/microsoft/rushstack/issues/3430

---

> [!NOTE]
> **Low Risk**
> Tooling and committed API snapshots only; no runtime SDK behavior changes in this diff.
> 
> **Overview**
> Adds **API Extractor coverage for `@revenuecat/purchases-js-vega`**, parallel to the main web package. A new `vega/api-extractor.json` targets `Purchases.vega.es.d.ts`, and the initial baseline lives in `vega/api-report/purchases-js-vega.api.md`.
> 
> `pnpm run extract-api` now runs API Extractor for **both** packages. CI’s “Check if API has changed” step and the `update_error_codes` automation were extended so uncommitted drift in **either** report fails the build or gets committed. README notes that both report folders must stay in sync when updating public APIs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ba8432bc24dd5a0a9e621f6bdbd9dcd8f9732440. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


vega/api-report/purchases-js-vega.api.md
```diff
@@ -0,0 +1,787 @@
...
+    readonly redemptionInfo: RedemptionInfo | null;
+    readonly storeTransaction: StoreTransaction;
+    /* Excluded from this release type: attributionMetadata */
+    /* Excluded from this release type: productChange */
+}
+
+// Warning: (ae-forgotten-export) The symbol "Purchases_2" needs to be exported by the entry point Purchases.vega.es.d.ts
+//
+// @public (undocumented)
+export class Purchases extends Purchases_2 {
```
fire-at-will, line 539: `Purchases_2` refers to `Purchases` in the main module. I think this is fine (we refer to it as `CorePurchases` in the code), but it is a little weird to see at first

tonidero: Ah ok I guess this answers one of my questions in the previous PR 😅... This does mean we're exposing a type that we're not actually exporting (as the warning above indicates)... So not sure if it could cause any issues TBH... Maybe @james-od could also confirm whether this is ok?

tonidero (approved):
I think this is ok as well... But would be good to confirm with someone else 🙏

vega/api-report/purchases-js-vega.api.md
fire-at-will: @lburdock @james-od Do y'all have any thoughts here? I'm also open to other approaches if there's a better one 👀

lburdock: This should be okay! The Vega Purchases package includes the core Purchases type and class internally to provide the inherited API, but it doesn’t export them, so users can’t import them directly.

lburdock (approved):

james-od (approved):

PR #1177, opened 2026-09-17
feat(checkout): send submitted billing country on checkoutComplete

dalshamaa:
## Description

When tax collection is disabled, checkout never calls the pricing/tax endpoint, so no country is ever captured for the purchase — Web Billing receipts for tax-disabled apps end up with a null country. The Payment Element already knows the customer's billing country (`selectedCountry`); this passes it as `billingAddress` on `checkoutComplete` so the backend can persist it as the receipt's country.

Backend counterpart (persists it + gates behind a per-app flag): RevenueCat/khepri#25329.

Screenshot from test against canary showing country code as France (I set it that way when i entered the card info)
image

## Details

- `payment-entry-page.svelte` `completeCheckout`: include `billingAddress: { countryCode: selectedCountry }` when the payment element reported a country; otherwise unchanged (`undefined`).
- No behavior change when the element didn't report a country, and the backend only uses this when a tax calculation did **not** already capture a country — so tax-enabled flows are unaffected.

## Testing

- New test in `payment-entry-page.test.ts`: with tax collection disabled and the payment element reporting `US`, submitting the form calls `checkoutComplete` with `billingAddress: { countryCode: "US" }`.
- `vitest` (payment-entry-page: 33 passed), `prettier`, `eslint`, `tsc --noEmit`, `svelte-check` (0 errors) all pass.


dalshamaa:
closing this as we went with an approach that does not require an sdk change

